PR #8, opened 2015-02-12
Remove usage of `Function.prototype.bind`.

rwjblue:
Using `Function.prototype.bind` makes this fail on Phantom 1.x (unless you polyfill).  This was the only instance, and seems to be fairly easy to remove the `bind`, so I think it makes sense to make this easier to consume.


rwjblue:
I don't see this under test, so you should probably manually double check that things function properly...

PR #9, opened 2015-02-12
Enable usage without Ember Data.

rwjblue:
`ember-inflector` is used by Ember Data, but we should pull it in
manually so that it does work without Ember Data.


rwjblue:
This doesn't work quite right due to the inflector not being a proper ember-cli addon.

rwjblue:
OK, updated (added a simple test), and ready for review.

samselikoff:
awesome, now this will spur me to write tests for all the different shorthands.

thanks!

PR #21, opened 2015-02-18
Don't include files in production build

### LINKED ISSUE #17: Exclude data from production build
Right now, the server doesn't run in production (unless force is specified), but all the data + route definitions end up in `app.js`. Is it possible to exclude these files?

marcioj:
In order to test this behaviour I added some files ending with `*test-node.js` which will be tested in the node context using mocha. I tried to not touch so much the ember-cli internals, since testing these file inclusions isn't so clear, but if something changes in the next ember-cli versions, the tests will point out.

//cc @samselikoff 

Fixes #17

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,17 +7,32 @@ module.exports = {
   included: function included(app) {
     this.app = app;
 
-    app.import(app.bowerDirectory + '/FakeXMLHttpRequest/fake_xml_http_request.js');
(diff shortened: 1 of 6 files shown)


samselikoff:
awesome, thanks so much for this! would have taken me forever :smile:

marcioj:
Glad to help @samselikoff !

PR #22, opened 2015-02-20
Remove app/pretender dir in production builds

### LINKED ISSUE #17: Exclude data from production build

marcioj:
By default ember-cli includes anything found in [`app/*`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L206) dir like `app/pretender`. To get around this we can remove the `app/pretender` from the app tree using broccoli-funnel when it's not needed.

Fixes #17


samselikoff:
<3

PR #40, opened 2015-02-28
Fix bug with dasherized names

### LINKED ISSUE #39: Discordance when POST url is undesrcored and resource is hypheneated
I have a model named `StaffMember`, whose normalize name following ember-cli convections would be `staff-member`. I use rails as a backend, so my urls have to use snake case (`staff_member`)

My `app/pretender/config.js` contains this:
```js
  this.get('/staff_members', 'staff-members');
  this.post('/staff_members', 'staff-member');
  this.get('/staff_members/:id', 'staff-member');
  this.del('/staff_members/:id', 'staff-member');
  this.put('/staff_members/:id', 'staff-member');
```

The problem comes when I save a record, performing a POST request to `/api/staff_members` with a type string of `staff-members`. The data doesn't match in [this line](https://github.com/samselikoff/ember-pretenderify/blob/3211cad6305064c2c22c79d586386d8dd8de66af/addon/controllers/post.js#L18) and therefore breaks in [this other line](https://github.com/samselikoff/ember-pretenderify/blob/3211cad6305064c2c22c79d586386d8dd8de66af/addon/store.js#L52).

To make it easier, I created this little project to demonstrate the problem: https://github.com/cibernox/bug_demo

Install dependencies, run `ember serve` and go to http://localhost:4200/tests to see only one test failing.

cibernox:
Long story short:
no-idea

This is an attempt to fix #39, because `POST` requests don't work with models whose name contains many words, p.e `StaffMember`.

Since the JSON in the body of post requests when using ActiveModel Serializer/Adapter uses underscored names, p.e `staff_members`, but the normalized names of the models in ember-cli use hyphens, ember-pretenderify never crashes then trying to add those record to the store.

The solution in this PR is to always undercore the string containing the normalized name of the model to match the root key of the json in the post request.
However, not everybody will AMS-like apis and I don't think this convention will be good for everybody, but I couldn't come up with a better solution.

### DIFF
--- a/addon/controllers/post.js
+++ b/addon/controllers/post.js
@@ -1,4 +1,4 @@
-import { singularize, underscore } from '../utils/inflector';
+import { singularize } from '../utils/inflector';
 import BaseController from './base';
 
 /*
@@ -15,7 +15,6 @@ export default BaseController.extend({
   stringHandler: function(string, store, request) {
     var type = string;
     var postData = this._getJsonBodyForRequest(request);
-    // var attrs = postData[underscore(type)];
     var attrs = postData[type];
     var model = store.push(type, attrs);
 
--- a/addon/controllers/post.js
+++ b/addon/controllers/post.js
@@ -1,4 +1,4 @@
-import { singularize } from '../utils/inflector';
+import { singularize, underscore } from '../utils/inflector';
 import BaseController from './base';
 
 /*
@@ -15,7 +15,7 @@ export default BaseController.extend({
   stringHandler: function(string, store, request) {
     var type = string;
     var postData = this._getJsonBodyForRequest(request);
-    var attrs = postData[type];
+    var attrs = postData[underscore(type)];
     var model = store.push(type, attrs);
 
     var response = {};
--- a/addon/controllers/post.js
+++ b/addon/controllers/post.js
@@ -1,4 +1,4 @@
-import { singularize } from '../utils/inflector';
+import { singularize, underscore } from '../utils/inflector';
 import BaseController from './base';
 
 /*
@@ -15,6 +15,7 @@ export default BaseController.extend({
   stringHandler: function(string, store, request) {
     var type = string;
     var postData = this._getJsonBodyForRequest(request);
+    // var attrs = postData[underscore(type)];
     var attrs = postData[type];
     var model = store.push(type, attrs);
 
--- a/addon/store.js
+++ b/addon/store.js
@@ -1,4 +1,4 @@
-import { pluralize, underscore } from './utils/inflector';
+import { pluralize } from './utils/inflector';
 
 /*
   An identity map.
@@ -116,7 +116,7 @@ export default function() {
   };
 
   this._keyForType = function(type) {
-    return underscore(pluralize(type));
+    return pluralize(type);
   };
 
   this._findDataForType = function(type) {
--- a/addon/store.js
+++ b/addon/store.js
@@ -1,4 +1,4 @@
-import { pluralize } from './utils/inflector';
+import { pluralize, underscore } from './utils/inflector';
 
 /*
   An identity map.
@@ -116,7 +116,7 @@ export default function() {
   };
 
   this._keyForType = function(type) {
-    return pluralize(type);
+    return underscore(pluralize(type));
   };
 
   this._findDataForType = function(type) {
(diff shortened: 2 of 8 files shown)


samselikoff:
sorry btw, I've looked at this but want to see it locally. I've been busy with emberconf but now that my talk is over I'll have time to look at this oon

cibernox:
Too bad I wasn't there :(
At least my timeline was flooded of tweets of people excited about all talks.

samselikoff:
So in terms of a test that proves the behavior is broken currently, I looked at one of your acceptance tests:

```js
test("works with multi-word models using a underscored URL and a hypheneated model name", function() {
  var body = '{"staff_member":{"name":"Ganon"}}';
  var result = controller.handle('post', 'staff-member', store, {requestBody: body, url: '/staff_members'});
  var staffMembersInStore = store.findAll('staff-member');
  equal(staffMembersInStore.length, 1);
  deepEqual(result[2], {'staff-member': {id: 1, name: 'Ganon'}});
});
```

This is saying that your POST message body contains a `staff_member` key. In this case, you should use `this.post('/staff_members', 'staff_member')` and everything should work. (You could even drop the second arg, since it matches the singularized version of the first).

Remember, the models in your Mirage store are representative of your back-end models, not your Ember models. You should format them however your backend would expect.

If you change the test above to

```diff
- var body = '{"staff_member":{"name":"Ganon"}}';
+ var body = '{"staff-member":{"name":"Ganon"}}';
```

it works. Is that clear?

cibernox:
Sorry for the delay, I'll try today.

cibernox:
Ok, I managed to make it work.

The steps to make it work were:

* Rename `factories/staff-member.js` to `factories/staff_member.js`.
* Rename `mirage/data/staff-member.js` to `mirage/data/staff_member.js` (I haven't updated to 0.0.22 yet)
* Use underscored names when defining endpoints in the config:
```js
  this.post('/staff_members');
  this.get('/staff_members/:id');
  this.del('/staff_members/:id');
  this.put('/staff_members/:id');
```
* Also, use the underscored name in factories `server.createList('staff_member', 20)`

The part that was missing the first try was to rename the files. It's somehow unexpected to use underscored names when ember-cli convention is hyphens, and when name of the models in the resolver also use hyphens, but I can get over it.

cibernox:
I'm closing this. 

I find particularly odd to call factories with underscored names because the resolver uses hyphens in model names. The name of the fixtures matching the underscored API makes a bit more sense though.

samselikoff:
@cibernox thanks for working through this. I definitely understand the confusion, given the dasherized convention within ember cli. The reason I'm still swaying towards keeping it how it is, is because Mirage actually has very little to do with Ember & Ember CLI at all. It's really more about reproducing your backend, than it is about integrating with Ember CLI. And since backends serve all sorts of stuff, I want to keep away from imposing any conventions where I don't have to.

You could imagine Mirage being a standalone library (that's the plan), and ember-cli-mirage simply being a wrapper/setup lib that helps you get going. In that case, I still think it'd make sense to keep it how it is (respect underscores, dashes, whatever) to keep it consistent. 

I think the way to think about it is, if your backend is sending up a `staff_member`, then you want a `staff_member.js` in either factories or fixtures. That's all. There's no conversion to what ember expects or anything like that, it's just, conform to whatever your backend is serving up.

Perhaps there's room in the docs for this clarification. Again, I know this must have been frustrating and I appreciate you spending time on this.

cibernox:
Don't get me wrong, I understand the motivation. Naming things after the payload you expect makes sense, and probably also makes the implementation simpler. 

And I can totally live with the configuration using `this.get('/staff_members/:id');` because it matches the API, and fixtures files being underscored also makes sense. The library happens to be born from the ember community but it's a general purpose utility.

The bit I found more confusing was that factories also relate to the server, not to your models. I am used to factory_girl in the ruby world where factories create AR objects, and follow rails' models conventions. In mirage factories don't instantiate your models, they just populate the store of fake server with payloads, so they match your APIs conventions, not ember ones.

Now everything makes sense, but it's wasn't obvious, probably because of my factory_girl background. That is the bit that should be clarified to avoid mislead rubysts.

Nice work!

samselikoff:
Yes exactly, I understand now. I actually also think there's a need for a lib that uses factories to create Ember Data objects - but as you said, the factories here are to populate Mirage's database.

thx again!

PR #42, opened 2015-03-04
Allow to modify the server timing in runtime.

cibernox:
Motivation:

I have some acceptance tests where I want to tests that if the server
take too long to respond, a message of "Loading more data" is displayed.
Making the interceptor respond without delay in tests is a reasonable
default but for my use case made some things untestable.

What I wanted was to allow in specific tests to set `server.timing = 1000`
and then restore the old value. Example:

```js
module('Acceptance: Slow Filtering', {
  beforeEach: function() {
   server.timing = 1000;
    application = startApp();
    server.createList('posts', 21);
  },

  afterEach: function() {
    server.timing = 0;
    Ember.run(application, 'destroy');
  }
});

test('when the server takes too much to respond a loading message is displayed', function(assert) {
   // Some test
});
```

That was not possible because the route handlers were
registered with the timing at the moment when the configuration is
loaded, by passint that number directly:

```js
  this.stub = function(verb, path, handler, code) {
    var timing = environment === 'test' ? 0 : this.timing;
    /* some code */
    interceptor[verb].call(interceptor, this.namespace + '/' + path, function(request) {
     /* route handling */
    }, timing});
  };
```

It turns out that Pretender is impressively good designed and the
`async` parameter can also be a function that returns the number of
milliseconds.

Thanks to that, the changes are minimal. The only tricky part was how
to test that. I created an acceptance test for that because I cound't
figure out how to test that in isolation.

Check it and tell me what you think and if you want me to document this a little bit in the readme or the wiki.

### DIFF
--- a/tests/acceptance/timing-test.js
+++ b/tests/acceptance/timing-test.js
@@ -1,0 +1,41 @@
+import Ember from 'ember';
+import startApp from '../helpers/start-app';
+
+var App;
+var contact;
+var appStore;
+
+module('Acceptance: Timing', {
+  setup: function() {
+    App = startApp();
+    appStore = App.__container__.lookup('store:main');
+    contact = server.create('contact');
+  },
+  teardown: function() {
+    Ember.run(App, 'destroy');
+  }
+});
+
+
+test('delays the response the configured time', function(assert) {
+  var done = assert.async();
+  var startTime = +new Date();
+  server.timing = 600;
+  appStore.findAll('contact').then(function(contacts) {
+    assert.equal(contacts.get('length'), 1);
+    var endTime = +new Date();
+    assert.ok(endTime - startTime >= 600, 'response has beed delayed 600ms');
+    done();
+  });
+});
+
+test('by default the response is not delayed', function(assert) {
+  var done = assert.async();
+  var startTime = +new Date();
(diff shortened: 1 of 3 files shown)


tests/acceptance/timing-test.js
```diff
@@ -0,0 +1,42 @@
...
+var appStore;
+
+module('Acceptance: Timing', {
+  setup: function() {
+    App = startApp();
+    appStore = App.__container__.lookup('store:main');
+    contact = server.create('contact');
+  },
+  teardown: function() {
+    server.timing = 0;
```
samselikoff: do you need this?

samselikoff:
this is awesome! just had one comment above

cibernox:
Comment corrected. I wasn't sure the the server was a singleton and I was concerned by leaking state, but apparently it's ok.

samselikoff:
Yep, server is instantiated in initializer on app startup, which happens for each test.

Thanks!!

cibernox:
I'll add some docs to the readme now.

cibernox:
@samselikoff The wiki is not public editable. Can you open it? There is no such thing as wiki PRs

samselikoff:
@cibernox ah thx for reminder. I'll take care of it !

PR #45, opened 2015-03-06
Enable only when visiting tests page.

abuiles:
We can't rely always on `--proxy` being set since we can also specify
multiple proxies doing `ember g proxy`.

With this we can use both proxy and mirage without having any
problems, additionally if we want all the request to be handle by
mirage we can just use `config.force`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,10 @@ Are you tired of
 - Wiring up tests for each of your apps manually, from scratch?
 - Changing lots of files/tests when your API changes?
 
-Ember CLI Mirage may be for you! It lets you create a client-side server using [Pretender](https://github.com/trek/pretender) to help you develop and test your app. By default, it only runs if you're not in production and if you're not proxying to an explicit API server via `ember serve --proxy`.
+Ember CLI Mirage may be for you! It lets you create a client-side
+server using [Pretender](https://github.com/trek/pretender) to help
+you develop and test your app. By default, it only runs if you're not
+in production and if you're in the tests page: `/tests`.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


samselikoff:
Ah, interesting, good point.

Question, when you use `ember g proxy` do you use proxies for some parts of your api and mocks for others?

samselikoff:
So it seems like, if you have proxies, that means you're developing against your backend that already exists. So you wouldn't be using Mirage for development, you'd just use it for testing.

So, we could make the assumption, if we find any files in `/server/proxies/` that Mirage shouldn't run during development. Thoughts?

abuiles:
@samselikoff 

>So it seems like, if you have proxies, that means you're developing against your backend that already exists. So you wouldn't be using Mirage for development, you'd just use it for testing.

Correct

>So, we could make the assumption, if we find any files in /server/proxies/ that Mirage shouldn't run during development. Thoughts?

It sounds right, but I'm not sure if this will disable Mirage when visiting /tests

samselikoff:
It wouldn't, I just need to update the [usingProxy](https://github.com/samselikoff/ember-cli-mirage/blob/master/config/environment.js#L3) function to check for the presence of any files in `/server/proxies`. Then, Mirage would only run in tests.

Your current PR would disable during development, which I don't want (one of the uses of Mirage is for development when your backend doesn't exist).

abuiles:
Makes sense, let's go with that.

samselikoff:
@abuiles should be all set with 8665000a931da7b909aa29f539e65f594aac28f4, I'll cut a release soon

Issue #51, opened 2015-03-11
Generators for factories and fixtures


samselikoff:
Factory would be

```
ember g mirage-factory contact
```

which creates

```js
//app/mirage/factories/contact.js
import Mirage from 'ember-cli-mirage';

export default Mirage.Factory.extend({
});
```

and fixture would be

```
ember g mirage-fixture contact
```

which creates

```js
//app/mirage/fixtures/contact.js
export default [
];
```

blimmer:
This was implemented.

PR #57, opened 2015-03-13
[FEATURE] Add mirage-factory and mirage-fixture blueprints

trabus:
This PR adds blueprints for mirage-factory and mirage-fixture as described in #51

I don't have any tests for them yet, would you like me to add the blueprint testing helpers so they can be tested?

### DIFF
--- a/blueprints/.jshintrc
+++ b/blueprints/.jshintrc
@@ -1,0 +1,6 @@
+{
+  "predef": [
+    "console"
+  ],
+  "strict": false
+}
--- a/blueprints/factory/files/app/mirage/factories/__name__.js
+++ b/blueprints/factory/files/app/mirage/factories/__name__.js
@@ -1,0 +1,4 @@
+import Mirage from 'ember-cli-mirage';
+
+export default Mirage.Factory.extend({
+});
--- a/blueprints/factory/index.js
+++ b/blueprints/factory/index.js
@@ -1,0 +1,14 @@
+module.exports = {
+  description: 'Generates a mirage factory.'
+
+  // locals: function(options) {
+  //   // Return custom template variables here.
+  //   return {
+  //     foo: options.entity.options.foo
+  //   };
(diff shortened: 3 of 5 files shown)


trabus:
I have a few things I'd like to add to this before it's ready. I'm going to add a description as well as the ability to pass params to generate factory properties. 

It would also be possible to add to the config when adding a fixture, similar to how the route blueprint generation adds your route to the router.js (thanks to @abuiles).

samselikoff:
That sounds awesome! Definitely give it a shot if you're up for it

samselikoff:
What about `m-fixture` and `m-factory` as aliases? (other ideas?) I know programmers hate to type...

trabus:
I'd be fine with fixture and factory personally. How about tests? I don't want to drag all that setup code in for testing without your okay first.

samselikoff:
I actually don't know how tests work for blueprints, you're saying to make sure they actually generate the files? I'm ok without them for now.

trabus:
Yeah, they run in mocha, generating an ember app, then running the generate command to verify the contents. Not very important for simple blueprints, but when the blueprint modifies exisiting content (like the route blueprint), or has complex rules to generate content (like the model blueprint), it's probably good to have.

I've actually got some more pressing things to work on right now, so I'm thinking I'll take the additional functionality I've listed above and break it into another issue to implement later. Then we can take the time to add the additional support for the tests.

samselikoff:
gotcha - so I can go ahead and merge this, and we'll track the other features in a different issue?

trabus:
Actually let me change the blueprint names to fixture and factory first.

trabus:
I went with factory and fixture. I looked through and there was only one other project using a fixture blueprint and it's not likely the two projects would be used together.

This is ready to go :+1:

samselikoff:
thanks mate!

PR #58, opened 2015-03-13
Disable if proxies folder present

samselikoff:


index.js
```diff
@@ -33,7 +33,8 @@ module.exports = {
 
   shouldIncludeFiles: function() {
     var config = this.app.project.config()['ember-cli-mirage'];
-    return config.force || (!config.usingProxy && this.app.env !== 'production');
```
samselikoff: this was wrong, because we should have included the files even if user is proxying, since they'll want their Mirage server running for tests

PR #62, opened 2015-03-15
Update config.js examples with new db API

coleww:
I noticed the config.js was using an old version of the db API, updated the syntax/methods to reflect the new one.

I'm not sure if {collection} is the best way to express what's happening, but I think between the comment on line 55 and the example function the intent should be clear.


samselikoff:
thanks !

PR #65, opened 2015-03-22
[WIP] Added isEnabledEnvironment to config

### LINKED ISSUE #52: Add option to run Mirage in test environment only
Even if someone isn't using `--proxy` they may want to disable Mirage in development. Let's add an option that lets them do this.

```
// config/environment.js
ENV['ember-pretenderify'] = {
  enableInTestingOnly: true
}
```

We'll need to modify `index.js` [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/index.js#L36) as well as the initializer [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/app/initializers/ember-cli-mirage.js#L12).

As a side note, is there a way to ignore that initializer based on something in the index.js? Seems like we're duplicating some logic.

trabus:
Fixes #52 by adding `isEnabledEnvironment` to the config, and only builds if `usingProxy` or `isEnabledEnvironment` are true. Also fixed the config so it isn't called multiple times in the `shouldIncludeFiles` method, as it was being called without the environment being passed, and was coming back with the environment as undefined most of the times it was called.

I still need to add tests, will do so tomorrow.


trabus:
@samselikoff if you don't mind, I'd like to upgrade the addon to use ember-cli v0.2.0. It's currently on [v0.1.9](https://github.com/samselikoff/ember-cli-mirage/blob/master/package.json#L33).

index.js
```diff
@@ -32,9 +33,7 @@ module.exports = {
   },
 
   shouldIncludeFiles: function() {
-    var config = this.app.project.config()['ember-cli-mirage'];
-
-    return config.force || (this.app.env !== 'production')
+    return this.addonConfig.force || !this.addonConfig.usingProxy ? this.addonConfig.isEnabledEnvironment : true;
```
samselikoff: during `ember serve` is there two builds for each environment (dev, test)? if there isn't this may not work

trabus: Ah, I see. I can check and make sure it's working for both. The main reason I changed it was because it was being called multiple times without the environment being passed in, so it had environment as `undefined` for some of the times it was called.

samselikoff: ah gotcha. Yah the reason I originally changed it here was bc I thought I needed to include the files into the build, since even tho `--proxy` may be passed the user would still want it for testing. But if there's a separate build with `env == test` during `ember s`, this should work

samselikoff:
can we do the 0.2 update in a separate PR, once we get this logic working + tests passing?

trabus:
I can do that, though part of the reason I upgraded was because I was having rimraf issues when trying to test.

samselikoff:
ugh :( ok we can do it all in one PR if we get it working

trabus:
I think I've got the config bit handled, I just need to finish writing some tests for `isEnabledEnvironment`.

samselikoff:
ok great, let me know when its all set and ill test out locally

samselikoff:
@trabus what do you think about this? I think I want to upgrade to 0.2.x because rimraf is causing headaches

trabus:
I feel pretty good about the current state, I just need to write the tests. Sorry it's taking so long, I got a little bogged down towards the end of the week.

samselikoff:
I can merge this, but I have an idea I want to run by you.

To simplify all these enabled/disabled/forced options, and to best align with Ember CLI conventions, why don't we just have a single `enabled` property?

By default, `enabled` will be true in test and in dev (if no proxy is present), and false in prod. So, current default behavior is unchanged. But, if a user wants to override, they just specify enabled as true or false in the particular environment:

```diff
// config/environment.js

module.exports = function(environment) {
  var ENV = {
    modulePrefix: 'app',
    ...
    'ember-cli-mirage':  {
    },
    APP: {
      // Here you can pass flags/options to your application instance
      // when it is created
    }
  };

  if (environment === 'development') {
+   ENV['ember-cli-mirage'].enabled = false;
  }

  if (environment === 'production') {
+   ENV['ember-cli-mirage'].enabled = true;
  }

  return ENV;
};

```

If a user specifies enabled as either true or false, this always wins; otherwise, fall back to default. the `enabled` keyword mirrors things like fingerprinting in Ember CLI.

What do you think? (I can do this in a separate commit after merging this, if you think it's a good move.)

trabus:
That would be much simpler, I'm totally fine changing my PR to align with this.

samselikoff:
cool, ty for your help. again its up to you, if you want i can merge + work on. but I do want to get your 2.x changes merged (some rimraf bug is killing me locally).

i'll be on later tonight

trabus:
So we'd go with `enabled = false` by default (if there is nothing to override in the config per environment), that way you'd have to opt into using it?

samselikoff:
Start with enabled as undefined and just use the rules we have, then if
it's defined it trumps all (whether it's true or false)

trabus:
I actually just had something come up and can't work on this tonight (in the process of buying a house). Feel free to take this over if you'd like to take care of this tonight.

samselikoff:
No worries at all, I'll take care of this in the morning

samselikoff:
I'll wrap up the rest here #76 and close this when I'm done

PR #70, opened 2015-03-29
Allow function handlers to dynamically overwrite the status code

jens-a-e:
I was trying to implement a `post` handler that validated the record and returned errors. Unfortunatly the status code was only overwritable statically by providing it as a third argument.
This way a function handler can look like this:

```
this.post('users', function(db, request) {
  var postedUser = JSON.parse(request.requestBody);
  if (postedUser.name === "") {
    request.status = 422;
    return { name: "can't be blank" };
  }
  // ... otherwise normal handling
});
```

Not sure if this is too dirty.

### DIFF
--- a/addon/controllers/front.js
+++ b/addon/controllers/front.js
@@ -26,6 +26,7 @@ export default {
     var data = this[controller][handlerMethod](handler, db, request, code);
 
     if (data) {
+      code = request.status !== 0 && request.status !== code ? request.status : code;
       return [code, {"Content-Type": "application/json"}, data];
     } else {
       return [code, {}, undefined];
--- a/tests/acceptance/controllers/get-controller-test.js
+++ b/tests/acceptance/controllers/get-controller-test.js
@@ -24,7 +24,7 @@ module('mirage:frontController GET', {
 // TODO: AMS-type response dependency, includes key
 // TODO: Extract to single class for data, handle response code/app elsewhere
 test("string shorthand works", function() {
-  var result = controller.handle('get', 'contacts', db);
+  var result = controller.handle('get', 'contacts', db, {});
 
   deepEqual(result[2], {contacts: contacts});
 });
@@ -39,7 +39,7 @@ test("string shorthand with id works", function() {
 
 // e.g. this.stub('get', '/', ['contacts', 'addresses']);
 test("array shorthand works", function() {
-  var result = controller.handle('get', ['contacts', 'addresses'], db);
+  var result = controller.handle('get', ['contacts', 'addresses'], db, {});
 
   deepEqual(result[2], {contacts: contacts, addresses: addresses});
 });


jens-a-e:
Hm. Ok. Travis failed (the GET string shorthand), i will investigate as soon as possible...

jens-a-e:
The issue was that the GET controller tests did not pass a request object to the `handle`.

Maybe it is a better option to actually check the request object in the controller `handle` itself?

samselikoff:
This is interesting, I think having the ability to change status response dynamically should be supported. I don't think we want to add the response code to the request object, we should figure out some other way  to pass it to the response.

Just thinking out loud, are we moving towards too much complexity? On the one hand, Mirage was originally intended to be a lightweight representation of your backend. On the other, I agree it's really nice to have it functional for prototyping and such.

Anyway, if you don't want to get bogged down in specifics, and you're just doing this for testing, what I've done is redefine the route to always return an error, and then assert against your UI. Something like this:

```js
test("If the server errors on /contacts, the first error message should show", function() {
  server.get('/contacts', {
    errors: ['improper auth']
  }, 404);

  visit('/');

  andThen(function() {
    equal( find('.error span').text(), 'improper auth' );
  });
});
```

jens-a-e:
I totally agree on the thin-ness of Mirage. In testing scenarios this can already be handled in a good way. I would not change it.
The issue did arise in the prototyping phase (development mode). Mirage has saved me tons of time figuring out how my API should be structured to work well with ember and sketch ideas. :smiley: Just a dynamic status code was missing to return errors, etc.
I also think, there is no real need to figure out a proper response code to the request object. The `FakeXMLHTTPRequest` object Pretender uses already has the common `status` property ([see here](https://github.com/trek/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js#L142)).
Maybe it should be more consitently passed into the handlers/controllers?

samselikoff:
thanks for the info :) I definitely think dynamic status codes should be supported, along with a suggestion that you shouldn't spend a ton of time recreating every validation in mirage.

Something still feels wrong to me about adding the status code to the request object, but I'll take another look when i have some time. Thanks for working on this :)

jens-a-e:
Yes, it feels a bit odd the way FakeXMLHTTPRequest leverages the request object to be a combinde request AND response object. But i don't think it is a good approach to abstract them into two seperate object like a common approach would be either.

tchak:
It would be nice to be able to return custom headers also... The api I am planing now, use headers for pagination (Range) for example.
So, why not return from function an array with [status, headers, body]? This is what pretender is doing internally. And it is something pretty common (Rack).
I know you want preserve simplicity but if one is up to override the function call, it already means he is doing something custom, so a little more options could not hurt.

samselikoff:
@tchak great point. I'm planning on introducing a serializer layer that takes the data from the response handler and runs it through a serializer, so that was another reason I wanted to keep the response just about the data.

Perhaps I could just pass in a `response` object, and you could set status code with `response.status` and headers with `response.headers`? If 95% of the time people want to stick with the defaults, I'd really prefer them not to have to do `return [null, null, {some: 'data'}]` or whatever.

tchak:
I think there is two ways of seeing it:
* As I mentioned earlier, my point is if someone is down to use callback he is out if 95% case.
* The other way of seeing it, is to say that there should be a middle ground. In this case I would add another layer of thin abstraction for cases when you want to just return the body.

```
this.post('users', function(db, request) {
    return { name: "can't be blank" };
});
```
vs with `raw` option
```
this.post('users', function(db, request) {
    return [422, null, { name: "can't be blank" }];
}, true);
```

tchak:
I actually kinda love the second solution with a boolean option. I would be happy to provide a PR for this, if there is an agreement.

tchak:
My grief with passing in response object is that we loose a nice pure and functional nature of the callback. But I understand if you would not consider this a very strong argument :)

samselikoff:
thanks for your input! My worry about `true` as fourth param, is, what if we haven't thought of everything? Whereas 

```js
this.post('users', function(db, request, response) {
  response.code = 422;

  return { name: "can't be blank" };
});
```

gives us more flexibility if there's more stuff we want/need to add to response. Also I think just by reading that, it's clearer what's going on - whereas just passing `true` as third param, it isn't clear that now the response is `[code, header, data]`. You'd have to reference the docs to understand that.

Thoughts?

tchak:
I can hardly argue with this, and even if I prefer a strictly functional triplet form, it would be easy to achieve with a wrapper in my code. So :+1: for passing in a response object.

samselikoff:
k. curious, what do you mean by strictly functional form? you mean just working with hashes/arrays instead of objects?

tchak:
I mean, callback function being a pure function. No assignments to passed in objects. But, as I mentioned earlier, this is largely a matter of personal preference :)

samselikoff:
ahh I see, no side effects. I like this too, I'll think more about it before making a decision.

samselikoff:
I'm leaning back towards returning the triplet. It's not consistent to set `statusCode` on a `response` object but not the `body`.

cibernox:
Today I just face this very problem, I wanted to return a 422 from a post action dynamically. And in fact, Y foreseen that I'll need to do a similar thing to customize the request headers next week, so I think that mirage should allow completely customized responses.

That said, I really love it's API being so consistent and easy to learn. In 85% of the cases the default behaviour of the shorthand methods is enough for me. 
I think that a good way of allowing more taylor made responses and at the same time allow uses to return triplets could be accomplished by just exposing the underlying primitive, call it `Response` or something different.

Example:

```js
  this.post('/homeworks', function() {
    let data = {
      errors: { description: ["can't be blank"] }
    };
    let headers = { 'X-User-Id': 123, 'X-School-Id': 234 };
    return new FakeResponse(422, headers, data);
  });
```

Like in ember, the default behaviour/config will be good for you, but when you want to go closer to the metal you can use the basic primitives to accomplish the task.

samselikoff:
This is exactly where I was headed too:

```js
 this.post('/homeworks', function() {
    let data = {
      errors: { description: ["can't be blank"] }
    };
    let headers = { 'X-User-Id': 123, 'X-School-Id': 234 };
    return Mirage.response(422, headers, data);
  });
```

that way it's still purely functional, and if the return value isn't an instance of `Mirage.response`, it stays with the default behavior (i.e. it's just returning data).

samselikoff:
f70066fe0ab31eee82b1c5f41f10dc7bc1a32881 closes, let's kick the tires a bit and see if we need anything else. The API is

```js
// mirage/config.js
import Mirage from 'ember-cli-mirage';

export default function() {
  this.get('/blah', function() {
    return new Mirage.Response(202, {some: 'header'}, {some: 'data'});
  });
}
```

cibernox:
@samselikoff Great. 
I've been working in another feature in pretender to allow to simulate file uploads and progress event. Once it's merged I'll bring that feature to mirage. First only as an entry in the docs, maybe later with some convenience syntax sugar over the low level FakeXMLHttpRequest.

PR #83, opened 2015-04-07
Updates readme to use the latest ember-cli command.

willrax:
The ember cli addon command was changed in the latest version. I'm making the assumption that the readme always assumes the user is using the latest ember-cli version. :smile:


samselikoff:
ty senor :)

PR #84, opened 2015-04-08
Update base controller id coercion

mikehollis:
I have a model that has an id with dashes in it. When trying to model this in mirage, I noticed that it grabbed the id with just the first number.

I added a couple of tests for my edge case (where `2-1` was the id which resolved to `2`)

I'd love to see this merged, but if there's something that I'm overlooking, just let me know!


samselikoff:
this looks great, and all the tests pass! thank you!

Generally I like to stay away from testing private methods but I think this makes sense. :+1:

PR #85, opened 2015-04-08
Updates to use broccoli-merge-trees.

willrax:
This switches to using broccoli-merge-trees directly instead of the deprecated `this.mergeTrees`. I also added broccoli-funnel.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 'use strict';
 var path = require('path');
+var mergeTrees = require('broccoli-merge-trees');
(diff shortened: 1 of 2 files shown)


samselikoff:
Thanks Will! Could you add

    var mergeTrees = require('mergeTrees');
    var Funnel = require('broccoli-funnel');

to the top of index.js? I think that will get the tests passing.

willrax:
Yep, just realised this issue. I'll fix it up now.

PR #97, opened 2015-04-15
Put broccoli-merge-trees and broccoli-funnel as dependencies

### LINKED ISSUE #93: Cannot find module 'broccoli-merge-trees'
After 1246cdc46161ca01f54b476c58f8e95ac36b13a6,
ember serve fails with:
Cannot find module 'broccoli-merge-trees'.
(using ember-cli 0.2.3)

Workaround:
```
npm install --save-dev broccoli-merge-trees
npm install --save-dev broccoli-funnel
```

cibernox:
This should fix #93

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,6 @@
   "author": "Sam Selikoff",
   "devDependencies": {
     "broccoli-asset-rev": "^2.0.0",
-    "broccoli-funnel": "^0.2.3",
-    "broccoli-merge-trees": "^0.2.1",
     "chai": "2.0.0",
     "ember-cli": "0.2.0",
     "ember-cli-app-version": "0.3.2",
@@ -54,5 +52,8 @@
   "ember-addon": {
     "configPath": "tests/dummy/config"
   },
-  "dependencies": {}
+  "dependencies": {
+    "broccoli-funnel": "^0.2.3",
+    "broccoli-merge-trees": "^0.2.1"
+  }
 }


samselikoff:
Did you try it locally? I'm able to reproduce the bug + moving the libs to `dependencies` doesn't seem to fix it. `ember g ember-cli-mirage` generates the same error:

> Cannot find module 'broccoli-merge-trees'

Also the docs seem to suggest adding dependencies to `devDependencies` [here](http://www.ember-cli.com/#packagejson). Could this be a 0.2.3 bug?

cibernox:
I tried. In fact in one app I am using my fork, and the error disappeared for me.

I'm a rubyist but it the differente between `devDependencies` and `dependencies` I think is analogous to the difference between `add_development_dependency` and `add_runtime_dependency` in `.gemspec` files.

Things that are needed to develop the addon go inside `devDependencies`, like by example `mocha` or `qunit` go there, but other things that are needed while this code is being used by other project and are needed in order to make it work should go inside `dependencies`.

Since [this commit](https://github.com/samselikoff/ember-cli-mirage/blob/68eca7a7ec6034421eb01205df6fcc464e34207a/index.js#L3-L4) added `require` statements in the `index.js` for those dependecies and the `index.js` is the entry point of the addon for ember-cli to *consume* this code, those dependencies have to be runtime dependencies, because otherwise **npm** won't install them.

It *might* work if it turns out that you are already installing those addons because other dependencie happens to require them as well, but it's just a lucky strike.

samselikoff:
right I do understand that, and I buy your reasoning, but on a new app I'm getting the same errors. Could you try a new app and see if this fork works for you?

We may need to go back to `this.mergeTrees` and `this.Funnel` for now, since they're deprecated but still supported.

cibernox:
I'll try to create a brand new app now

cibernox:
Brand new app created and it works using my fork. The `package.json`.

```json
{
  "name": "sample-app",
  "version": "0.0.0",
  "description": "Small description for sample-app goes here",
  "private": true,
  "directories": {
    "doc": "doc",
    "test": "tests"
  },
  "scripts": {
    "start": "ember server",
    "build": "ember build",
    "test": "ember test"
  },
  "repository": "",
  "engines": {
    "node": ">= 0.10.0"
  },
  "author": "",
  "license": "MIT",
  "devDependencies": {
    "broccoli-asset-rev": "^2.0.2",
    "ember-cli": "0.2.3",
    "ember-cli-app-version": "0.3.3",
    "ember-cli-babel": "^5.0.0",
    "ember-cli-content-security-policy": "0.4.0",
    "ember-cli-dependency-checker": "0.0.8",
    "ember-cli-htmlbars": "0.7.4",
    "ember-cli-ic-ajax": "0.1.1",
    "ember-cli-inject-live-reload": "^1.3.0",
    "ember-cli-qunit": "0.3.10",
    "ember-cli-uglify": "1.0.1",
    "ember-data": "1.0.0-beta.16.1",
    "ember-export-application-global": "^1.0.2",
    "ember-cli-mirage": "git://github.com/cibernox/ember-cli-mirage.git#make-broccoli-mergetrees-runtime-dependency"
  }
}
```

samselikoff:
really hm, ok let me investigate

samselikoff:
ah yep, I guess changing locally didn't tell ember CLI about the new package.json. Thanks!

PR #106, opened 2015-04-22
[WFA] Change status 204 to 200 when response is present

cibernox:
When a custom handler of an action that by default returns a 204 status
generates a non-blank response, a status 200 is used instead. 

Also added high-level acceptance tests for custom handlers

### DIFF
--- a/addon/controller.js
+++ b/addon/controller.js
@@ -2,12 +2,12 @@ import Ember from 'ember';
 import shorthandHandlers from 'ember-cli-mirage/shorthands/index';
 import Response from './response';
 
-const isArray = Ember.isArray;
-const isBlank = Ember.isBlank;
-const typeOf  = Ember.typeOf;
-const keys    = Ember.keys;
+var isArray = Ember.isArray;
+var isBlank = Ember.isBlank;
+var typeOf  = Ember.typeOf;
+var keys    = Ember.keys;
 
-const defaultCodes = {
+var defaultCodes = {
   put: 204,
   post: 201,
   delete: 204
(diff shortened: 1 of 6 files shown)


addon/controller.js
```diff
@@ -2,7 +2,12 @@ import Ember from 'ember';
 import shorthandHandlers from 'ember-cli-mirage/shorthands/index';
 import Response from './response';
 
-var defaultCodes = {
+const isArray = Ember.isArray;
```
cibernox: I also added a couple of ES6 stuff where it seemed appropiate.

addon/controller.js
```diff
@@ -10,16 +15,24 @@ var defaultCodes = {
 
 export default {
 
-  handle: function(verb, handler, db, request, code) {
-    code = code ? code : (defaultCodes[verb] || 200);
-
+  handle: function(verb, handler, db, request, customizedCode) {
+    var code, isEmptyObject;
     var handlerMethod = this._lookupHandlerMethod(verb, handler);
-    var response = handlerMethod(handler, db, request, code);
```
cibernox: The code was not being use at all in the handlers (function or object handlers)

addon/controller.js
```diff
@@ -10,16 +15,24 @@ var defaultCodes = {
...
+    var response = handlerMethod(handler, db, request);
 
     if (response instanceof Response) {
       return response.toArray();
 
     } else {
+      if (customizedCode) {
+        code = customizedCode;
```
cibernox: If a status code is specified by the user, it takes precedence over everything else.

samselikoff:
awesome. thanks for all your time + effort on this!

eventually I think we'll need to find a better way to isolate the acceptance testing, or the dummy app will eventually get out of hand.

great work

PR #107, opened 2015-04-22
[WFA] More clever ID parseInt

cibernox:
Mirage was parsing ids like "123-456" like "123". That is not correct.
Now only tries to parse the ID as an integer if its a String that looks like an Interger.

### DIFF
--- a/addon/db.js
+++ b/addon/db.js
@@ -1,5 +1,7 @@
 import Ember from 'ember';
 
+var allDigitsRegex = /^\d+$/;
+
 /*
   An identity map.
 */
@@ -74,7 +76,9 @@ export default function() {
 
   this._find = function(collection, id) {
     // If parses, coerce to integer
-    id = parseInt(id, 10) || id;
+    if (typeof id === "string" && allDigitsRegex.test(id)) {
+      id = parseInt(id, 10);
+    }
 
     return this[collection].filter(function(obj) {
       return obj.id === id;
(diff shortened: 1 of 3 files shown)


cibernox:
I (well, a coworker in fact) have another PR in the pipeline to improve the behaviour of mirage for records with non-numeric ids, but this can be reviewed on it's own.

samselikoff:
Yes sorry, you can't use ES6 within addons. I'll look at this (for some reason I thought someone else took care of this)

samselikoff:
I think this will be fine for now, as for your other PR you may want to hold off. I think eventually I'm going to treat all ids as strings within mirage, and leave decisions about int vs string to the serializer layer.

At the least open an issue to discuss before you sink a lot of work into it :)

cibernox:
Don't worry the other PR is a one liner too. No need to wait

samselikoff:
Can you add a test to prove it's not currently working? [This test](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/unit/shorthands/utils-test.js#L23-L26) seems to suggest the current functionality works, indeed I just checked with your example of '123-456' and it seems to work. What am I missing?

samselikoff:
What version were you on when you saw the bug?

samselikoff:
Ah, I think the bug is in db.find (I hadn't updated with the `isNaN` check). Your solution is fine, but can we get a test on `db.find`?

cibernox:
Test added.

samselikoff:
woot! thanks man

PR #110, opened 2015-04-23
[WFA] Fix update for records with string ids.

cibernox:
I found that `db.contacts.update('123-abc', { name: 'new-name' })` didn't work.

Now it does :)


samselikoff:
ah - and `remove`! mind adding one more?

cibernox:
sure

PR #116, opened 2015-04-30
Decode the failed URI before logging it

bdvholmes:
I was finding

screen shot 2015-04-30 at 2 04 43 pm

difficult for my brain. This pull request makes the message look like this instead

screen shot 2015-04-30 at 2 59 47 pm


samselikoff:
sweet, thx man!

PR #118, opened 2015-05-11
Update README example syntax

seanpdoyle:
* include `assert` as a `test` callback parameter
* use ES2015 fat arrow syntax where applicable


samselikoff:
:heart:

PR #123, opened 2015-05-15
Coalesce find requests in get shorthand handlers

cibernox:
Now shorthand methods can receive a POJO with options, ignored by all
shorthands except of the `get` one.

The `get` handler supports the `coalesce` option, which defaults to
false. When enabled that option will make the shorthand method look for
an `ids` attribute on the queryParams, and will return with a collection
containing only the models with those ids.

Examples:

```js
this.get("/contacts", { coalesce: true }); // "/contacts?ids=1,3]"
this.get("/friends", "users", { coalesce: true }) // "/friends?ids=2,4"
```

### DIFF
--- a/addon/controller.js
+++ b/addon/controller.js
@@ -16,10 +16,10 @@ var defaultCodes = {
 
 export default {
 
-  handle: function(verb, handler, db, request, customizedCode) {
+  handle: function(verb, handler, db, request, customizedCode, options) {
     var code, isEmptyObject;
     var handlerMethod = this._lookupHandlerMethod(verb, handler);
-    var response = handlerMethod(handler, db, request);
+    var response = handlerMethod(handler, db, request, options);
 
     if (response instanceof Response) {
       return response.toArray();
--- a/addon/server.js
+++ b/addon/server.js
@@ -5,6 +5,29 @@ import Db from './db';
 import controller from './controller';
 
 /*
+  Given a variable number of arguments, it generates an array of with
+  [path, handler, code, options], `path` and `options` being always defined,
+  and `handler` and `code` being undefined if not suplied.
+*/
+function extractStubArguments(/* path, handler, code, options */) {
+  var ary = Array.prototype.slice.call(arguments);
+  var argsInitialLength = ary.length;
+  var lastArgument = ary[ary.length - 1];
+  var options;
+  var i = 0;
+  if (typeof lastArgument === 'object') {
+    argsInitialLength--;
+  } else {
+    options = { colesce: false };
+    ary.push(options);
+  }
+  for(; i < 5 - ary.length; i++) {
+    ary.insertAt(argsInitialLength, undefined);
+  }
+  return ary;
+}
+
+/*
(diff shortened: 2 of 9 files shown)


achambers:
Cracking work @cibernox. Great idea ;) :+1:

cibernox:
@samselikoff Do you agree in the syntax? I'm not sure if coalesce is the best name for this, I used it for being the server-side sibling of this feature in ember-data, but probably it's not the most accurate term.

samselikoff:
Ah sorry, somehow missed this notification. Yes I think it's good! I need to look and make sure we're not closing out future improvements to the shorthand API - but I think this is correct, essentially, having the last param be an options hash.

One thought, I've seen coalescing the way you've implemented (`?ids=[1,2]`) but I've also seen it something like `?ids[]=1&ids[]=2`. Are either of these more "standard"? Should we have a default and allow users to change via a global config, or another option?

cibernox:
I think that `?ids=1,2,3` and `?ids[]=1&ids[]=2&ids[]=3` are equivalent but the first one it's not officially correct for not being url-safe, although it works in most browsers. 
In any case, the parsing of the queryParams its done automatically by pretender, by the time I access the `queryParams` in the request object I already have an array of IDs.

samselikoff:
oh awesome! Ok, great work, and thanks so much for your help!

samselikoff:
I really like this api, I'm thinking we'll add filtering, sorting and pagination some day as well, something like

```js
this.get("/contacts", {
  coalesce: true,
  filters: ['first_name', 'last_name'],
  sorting: ['age'],
  pagination: {
    per_page: 10
  }
});
```

cibernox:
What do you think about adding a configuration option like `this.coalesceFindRequests = true;` to enable this feature for all routes?

Just a convenience for people like me that has this behaviour on all the endpoints.

samselikoff:
Hesitant, since now you could define endpoints that inherit this behavior but that your app actually doesn't use. I'm planning on adding an "exporter" that describes your Mirage api, so you could hand that off to a backend engineer. I wouldn't want that to show that a certain route needs to coalesce ids if it doesn't.

I'll think more about this though. BTW could you gist your current project's mirage/config.js file? I'd just like to see some more real world examples.

cibernox:
Here: https://gist.github.com/cibernox/7501e86ad9982fa69f1d

We use mirage a lot because the frontend is going much faster than the backend.
Some times we add custom functions to simulate the real behaviour that the real API because pure CRUD is not enough. (p.e. in `this.put('/quiz_submission_questions/:id')`

samselikoff:
gotcha. cool, thx for showing!

cibernox:
btw, the wiki has just disappeared?

samselikoff:
www.ember-cli-mirage.com

seawatts:
I think this broke the shorthand
`this.get('/foo', ['friends', 'groups', 'foo'])`
Which should return all three of those sideloaded together in the response.

PR #124, opened 2015-05-17
Fix broken link to production config info

clekstro:


samselikoff:
thx! you just reminded me there are many links I need to fix in the docs.. :)

clekstro:
np! This is the only reference I found to localhost :)

PR #128, opened 2015-05-20
Where coherces query parameters to strings

cibernox:
Having this records in the db:
```js
[
 { id: 1, name: 'Bob', teacher_id: 123},
 { id: 2, name: 'Jane', teacher_id: 234}
]
```
performing a query like `/students?teacher_id=123` would return no
result because queryParameters in the requests are always treated as
strings, and `db.students.where({teacher_id: req.queryParams.teacher_id})`
will compare using `===`.

Now `where` coherces both operands to string before comparing them,
returning the expected student.

This also makes true === "true" and such, but that is very unlikely to cause
any problem.

### DIFF
--- a/addon/db.js
+++ b/addon/db.js
@@ -92,7 +92,7 @@ export default function() {
 
     Object.keys(query).forEach(function(queryKey) {
       records = records.filter(function(record) {
-        return record[queryKey] === query[queryKey];
+        return String(record[queryKey]) === String(query[queryKey]);
       });
     });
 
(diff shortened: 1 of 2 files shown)


samselikoff:
I'm a bit hesitant to merge this, I really think these situations only occur when dealing with ids. Once we get serializers, we will coerce all ids to strings internally. For other fields, i'm not so sure. (I'm thinking of weird things with 0, true/false, etc).

Thoughts?

cibernox:
It's true, this probably bites more often when comparing ids, but also with anything numeric like 
`/users?age=20` or `/addresses?post_code=28080`.

I think that occasions where a user legitimately wants `0` to be considered different than `"0"` and `true` different than `"true"` will be *extremely* rare.

samselikoff:
Ah right, I was thinking something like `.where({age: 0})` would return if age was `null` but I see now it won't. LGTM :+1:

PR #130, opened 2015-05-20
Handling urls types with slash before ?

knownasilya:
For example: '/api/properties/?ll=42.426894%2C-71.046524' 

That was not handled previously.


samselikoff:
awesome, thx! could you add a test [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/unit/shorthands/utils-test.js)?

knownasilya:
@samselikoff should be set

samselikoff:
:heart: on it

samselikoff:
Should be all set!

knownasilya:
Sweetness! Thanks a bunch!

PR #134, opened 2015-05-21
[closes #80] allow passthrough of non-defined routes

### LINKED ISSUE #80: Allow per-route and fallback passthrough routing
Currently I need to fetch certain SVG files hosted in public/ based upon some of my model data.  This system seems to make that impossible because it hijacks all my http requests even if they do not start with my chosen namespace.  So, I tried to move that ajax call into a mirage route but the same problem happens there.  It would be nice if all http requests that were not in the mirage namespace would behave normally.  After triple-checking your docs, I noticed the part about this.pretender.  According to the docs there, there is a this.passthrough function on pretender so it would seem I should be able to define something like 

```
this.get('/floorPlans/:id', this.pretender.passthrough)
```

I tried this with and without a dummy floorPlans fixture and neither way worked.  Is there a way to set up passthrough through mirage ?  If not, this becomes pretty useless to me :( because it is much better than other solutions I've tried for fixtures.  If so, I think it ought to be put in the config.js examples to save anyone else the trouble I went through .

samselikoff:


samselikoff:
I could use some help here. I verify the passthrough option works by making a failed ajax request. The tests pass in Chrome, but PhantomJS fails with this error:

> INVALID_STATE_ERR: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.

I could use some help, if anyone has any ideas!

samselikoff:
tracked it down, working on it here: https://github.com/trek/pretender/pull/74

constantm:
It sounds like [this merged PR](https://github.com/pretenderjs/pretender/pull/76) might fix this build, which might fix my [passthrough issue](https://github.com/samselikoff/ember-cli-mirage/issues/285). I'd like to help, but I'm not sure how to trigger a rebuild with the newest commits included?

samselikoff:
I need to think a little more about this api before coming back to it, and I want to finish the serializers first.

Have you tried upgrading your Pretender in your `bower.json`? You can do that independent of Mirage.

constantm:
Ah okay. I've tried upgrading Pretender yes - no luck. Would it not make sense to not actually touch any requests unless they are defined in the mirage config? From the look of it, it seems like this is what the PR does. However, I'll need to put some more effort in to fully understand how it works.

PR #138, opened 2015-05-27
make it obvious to import Mirage when using new Mirage.Response

ballPointPenguin:
It took me a minute to realize why I was getting "ReferenceError: Mirage is not defined". Maybe this will help the next person.

### DIFF
--- a/docs/latest/routes.md
+++ b/docs/latest/routes.md
@@ -62,22 +62,28 @@ Return the data you want as plain JS - it will be stringified and sent as the re
 Examples:
 
 ```js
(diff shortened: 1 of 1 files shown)


samselikoff:
Thanks!

jamesdixon:
@ballPointPenguin @samselikoff not sure if you guys think this would be overkill, but it may be worth making a special note to import Mirage in your `config.js` when using `Mirage.response`. It didn't even throw an undefined error for me - my authentication promise just kept resolving even though it shouldn't have. Just a thought!

PR #139, opened 2015-05-28
Add ability to have per environment configs.

cball:
First, here's the use case this is solving. The app is already existing and is set up with `--proxy`, so mirage needs to be explicitly enabled.

Workflow:
1. Create fixtures/factories for new feature development.
2. Stop using the fixtures in development when the API is ready, but keep using the factories in tests. This assumes a passthrough fallback is in place, either manually or from #134 so mirage doesn't complain about the route not being defined.

I can't think of a way to do this without requiring an update of some sort to consuming apps, and the approach here seems the least invasive. I can do this manually by adding `if (environment === 'test')` around test only routes, but it's messy. I also need to do this across at least 3 apps, and assume some others might be in a similar boat.

I added an integration test for this functionality because it's the only way I could think of to test that the initializer has actually been called.

Thoughts on the approach? Anything I'm missing?

### DIFF
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -1,5 +1,5 @@
 import ENV from '../config/environment';
-import userConfig from '../mirage/config';
+import baseConfig, { testConfig } from '../mirage/config';
 import Server from 'ember-cli-mirage/server';
 import readFixtures from 'ember-cli-mirage/utils/read-fixtures';
 import readFactories from 'ember-cli-mirage/utils/read-factories';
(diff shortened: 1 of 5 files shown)


tests/dummy/app/routes/friend.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+
+  model(params) {
+    return this.store.find('friend', params.friend_id);
+  }
+});
```
samselikoff: i think this file is unnecessary, no? isn't it the default behavior for that route

cball: Duh, my bad.

samselikoff:
I like this approach! Having `baseConfig` be shared is excellent, and it's right there next to `testConfig`, so you can see everything in one glance.

I tried some variations, having something like `this.testRoutes(() => ...` within the default function, but I think this is the cleanest. 

:+1: thanks for your help!

Issue #136, opened 2015-05-26
Use Factories in development


samselikoff:
Here's how I see this working. If Mirage finds a `/mirage/scenarios/default.js` file, it will use that to seed its database during development, rather than the fixture files.

This file will export a function that takes one argument, `server`:

```js
// app/mirage/scenarios/default.js
export default function(server) {
  var contact = server.create('contact');
  server.createList('comment', 20, {contact_id: contact.id});
}
```

Changes needed:

- [x] Add a line to the initializer [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/app/initializers/ember-cli-mirage.js#L13) to look for a function at `/app/mirage/scenarios/default.js`. Call this `var scenariosMap` (we plan to support multiple scenarios later on).
  - simply define a new `utils/read-scenarios` utility function. Alternatively, read-x should eventually be refactored.
- [x] Update the initializer [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/app/initializers/ember-cli-mirage.js#L25) (i.e. in non-testing environments) to check if `scenariosMap` exists. If so, invoke it, passing in the server; otherwise load the fixtures into the db.

g-cassie:
+1  Two thoughts below about future possibilities though I think this approach would work perfectly as a starting point.

If you had a huge fixture-set you may want to be able to split your scenarios among multiple files like you do right now with fixtures.  I presume because `scenarios/default.js` is just a function you can spread your code over subdirs and import - perhaps something like `/scenarios/default/*.js`?

My other thought is it would be nice to have an efficient way to build a realistic looking dataset.  For instance if you want 20 comments that have realistic text, it would be nice to be able to feed in a list of pojos with `comment.body` and `comment.title` values, but still be able to specify `comment.contact_id` once.  Again, because `default.js` is a script I guess you can just write some simple loops to get this kind of functionality.

samselikoff:
Yep, both great points. I want to use FakerJS to be able to write factories like

```js
export default Mirage.Factory.extend({
  first_name: Faker.firstName,
  last_name: Faker.lastName
});
```

and these would generate realistic fake data.

For now I think we would leave splitting the scenarios among multiple files to the developer. `/default.js` would be the function read in, and eventually I'd like to support loading different scenarios, perhaps via query params or some other mechanism.

g-cassie:
Anything I can do to help with this?

samselikoff:
[...]

PR #141, opened 2015-05-28
Implement fixture scenarios and refactor module loading

g-cassie:
Re https://github.com/samselikoff/ember-cli-mirage/issues/136#issuecomment-105636546

Need feedback on whether/how to test the new scenarios functionality. Hope that my refactor of the module loading makes sense.  Open to other ideas on how to structure it.


app/initializers/ember-cli-mirage.js
```diff
@@ -22,11 +23,13 @@ export default {
         server.loadConfig(testConfig);
       }
 
-      if (env === 'test' && factoryMap) {
-        server.loadFactories(factoryMap);
-
+      if (env === 'test' && hasFactories) {
+        server.loadFactories(modulesMap['factories']);
+      } else if (hasDefaultScenario && hasFactories) {
```
samselikoff: I think we also need a `env !== 'test'` here, since we don't want these running during testing

samselikoff:
module loading looks great, nice to get that cleaned up!

I'm not sure how to end to end test this either, since it's specifically not in `test` environment. I think what we should eventually do (not necessarily now) is create another object that understands the initializer logic, so we can unit test that.

For now if you make the one change I think we can ship this. I really appreciate your help!

g-cassie:
Right now the behaviour for tests is to use factories if they exist and if not then to use fixtures.  Correct?

If we are supporting fixtures in tests shouldn't we support scenarios as well?  Or alternatively should we disable both fixtures and scenarios for tests?

samselikoff:
Originally I conceived of scenarios as a development-only concept. You could boot the app in a default scenario, or a "new-user" scenario, or "pro-user" scenario, etc. Perhaps we'll use query params or a chrome extension to support loading different scenarios one day.

But I don't think scenarios makes sense for tests. Test themselves already are different scenarios, and I want to push for tests to set up their own state via factories. Not only is `/scenarios/default.js` far away from your tests (and thus make your tests less intention-revealing), I want to be able to set up a development scenario without it affecting all my tests. If we just loaded the scenario during tests, the tests wouldn't be atomic anymore and they'd all have to start from whatever data is in the scenario.

What do you think?

g-cassie:
Personally I think factories is best practice in tests so you shouldn't use fixtures or scenarios.  I see scenarios and fixtures as being substantially identical though so would just question why one works for tests but the other doesn't.  Do you support multiple fixture suites right now?

Let me know your decision one way or the other and I'll make the change.

samselikoff:
No we don't support multiple fixture suites, and I agree factories are best practice in tests. The whole point of scenarios is so people can use their factories to seed their database during development. I just want to keep those scenarios insulated from the test environment.

g-cassie:
Ok, change made.

samselikoff:
sweet, thanks!

g-cassie:
Where is the code for the docs website?

samselikoff:
On the `gh-pages` branch

PR #152, opened 2015-06-01
Use splice instead of insertAt

ZenCocoon:
Replace `Array.insertAt` by `Array.splice`.

Note: this make it possible to use `ember-cli-mirage` for addons.


samselikoff:
thx!

ZenCocoon:
My pleasure, thanks for this awesome tool :)

PR #133, opened 2015-05-20
Get babel working

cibernox:
Updated to the latest ember-cli (0.2.5) / Ember (0.12) / ED (beta17) and finally I managed to get ES6 working.

I haven't tested in a real app yet, so it's a WIP until I/someone does.

PR #155, opened 2015-06-03
Use namespace and glob route for passthrough requests.

cball:
This syntax should correctly use pretender passthroughs for any url (not just root level). It also automatically applies the namespace defined in the server config.


addon/server.js
```diff
@@ -54,10 +54,10 @@ export default function(options) {
     this.timing = environment === 'test' ? 0 : (this.timing || 0);
 
     if (this.passthrough) {
-      this.interceptor['get'].call(this.interceptor, '/:passthrough', this.interceptor.passthrough);
-      this.interceptor['post'].call(this.interceptor, '/:passthrough', this.interceptor.passthrough);
-      this.interceptor['put'].call(this.interceptor, '/:passthrough', this.interceptor.passthrough);
-      this.interceptor['delete'].call(this.interceptor, '/:passthrough', this.interceptor.passthrough);
+      this.interceptor['get'].call(this.interceptor, `${this.namespace}/*passthrough`, this.interceptor.passthrough);
```
samselikoff: pretty sure we can't use es6 yet

samselikoff:
did you see https://github.com/samselikoff/ember-cli-mirage/pull/133

addon/server.js
cball: Gotcha, will back that out, one sec.

cball:
@samselikoff I'll convert this to standard js, but oddly I didn't get the same failures #133 did on Travis. The only real difference I see is `^4.0.0` vs `^5.0.0` for babel.

samselikoff:
Sorry if I wasn't clear earlier. I had opened something similar here: https://github.com/samselikoff/ember-cli-mirage/pull/134 although I think your implementation handles some different cases.

The problem is Pretender has a bug in it which is causing a phantomjs test to fail: see [here](https://travis-ci.org/samselikoff/ember-cli-mirage/builds/65298883#L517). So we need to fix that before we merge in.

Once we do, I'm happy to merge this PR in and just kill mine.

samselikoff:
here's the bug: https://github.com/trek/pretender/pull/74

cball:
Totally agree. This was an intermediate step to give your PR the proper catchall syntax and use of namespace while I tried to work out the pretender bug. 

I'll check that out now, and combine into a new one.

samselikoff:
I see so we needed the star `*passthrough` for it to work - is that the problem we ran into the other day?

cball:
Exactly, that was our issue.

cball:
PR opened to fix the Pretender bug: https://github.com/trek/pretender/pull/76

cball:
Closing in favor of #188.

PR #161, opened 2015-06-07
Ember CLI 0.2.7.

blimmer:
Most of the changes are a result of running an `ember init` after the upgrade. Any other meaningful changes are noted inline.


tests/unit/addon-tree-test-node.js
```diff
@@ -47,7 +49,7 @@ describe('Addon', function() {
       var addon = getMirageAddon({ configPath: 'tests/fixtures/config/environment-production-enabled' });
       var addonTree = addon.treeFor('addon');
 
-      expect(addonTree.inputTrees.length).to.be.equal(1);
+      expect(addonTree.inputTrees.length).to.not.equal(0);
     });
```
blimmer: The new version of `ember-cli-babel` made this input tree not exactly equal 1. It now exactly equals 2. 

Instead of changing this to be "2" and we're checking that it equals 0 in the case where mirage shouldn't be included, I just check that an inputTree structure exists.

tests/unit/addon-tree-test-node.js
```diff
@@ -19,6 +19,8 @@ function findMirage(app) {
 }
 
 describe('Addon', function() {
+  this.timeout(15000);
```
blimmer: These tests run significantly slower than they used to. Ember-CLI 0.2.7 seems to be much much slower than 0.2.0.

samselikoff:
v. nice! @cibernox was having trouble in #133 but I think you figured out the issues with the node/addon tests. Thanks a lot!

PR #175, opened 2015-06-18
[closes #174] Check if `modulesMap['factories']` is empty via Ember.keys

### LINKED ISSUE #174: Does not properly detect when there are no factories
In the initializer, mirage checks to see if the user has defined factories [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/app/initializers/ember-cli-mirage.js#L14). However, the default empty value for `modulesMap['factories']` is an empty object `{}`. The return value of `Ember.isEmpty({})` is [actually `true`](http://emberjs.com/api/classes/Ember.html#method_isEmpty), so the value of the `hasFactories` variable in the initializer appears as though it will always be `true`. This means that in the 'test' environment it isn't possible to [load fixtures](https://github.com/samselikoff/ember-cli-mirage/blob/master/app/initializers/ember-cli-mirage.js#L32).

bantic:
This fixes #174. It relies on the now-private `Ember.keys` api, unfortunately, but this seems as though it should do the job.

### DIFF
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -4,6 +4,12 @@ import baseConfig, { testConfig } from '../mirage/config';
 import Server from 'ember-cli-mirage/server';
 import readModules from 'ember-cli-mirage/utils/read-modules';
 
+function hasModulesOfType(modulesMap, type) {
+  var modulesOfType = modulesMap[type] || {};
+
+  return Ember.keys(modulesOfType).length > 0;
+}
+
 export default {
   name: 'ember-cli-mirage',
   initialize: function(container, application) {
@@ -11,7 +17,7 @@ export default {
 
     if (_shouldUseMirage(env, ENV['ember-cli-mirage'])) {
       var modulesMap = readModules(ENV.modulePrefix);
-      var hasFactories = !Ember.isEmpty(modulesMap['factories']);
+      var hasFactories = hasModulesOfType(modulesMap, 'factories');
       var hasDefaultScenario = modulesMap['scenarios'].hasOwnProperty('default');
       var server = new Server({
         environment: env


app/initializers/ember-cli-mirage.js
```diff
@@ -11,7 +17,7 @@ export default {
 
     if (_shouldUseMirage(env, ENV['ember-cli-mirage'])) {
       var modulesMap = readModules(ENV.modulePrefix);
-      var hasFactories = !Ember.isEmpty(modulesMap['factories']);
+      var hasFactories = hasFactories(modulesMap);
```
samselikoff: build error was

> 'undefined' is not a function (evaluating 'hasFactories(modulesMap)')

I wonder if it's because `var hasFactories` here overrides the function declaration above?

samselikoff:
Ah, I recall someone else having trouble with fixtures but it got lost in the pile. Thanks for tracking this down!

bantic:
Oops, sorry about that test error, I believe it is fixed now. `npm test` for me locally shows passing on default and ember-release but failing on beta and canary with what appear to be unrelated reasons. If that doesn't seem normal I can take another look tonight.

bantic:
I've confirmed that the build failures here are for unrelated reasons (they are due to `ember#beta` now pointing to a different version (2.0) than it was the last time the ember-try beta scenario was run (at that time beta was 1.13), and ember-data is still using `Ember.EnumerableUtils` which was removed from ember beta). Those are mentioned in more detail in #177

samselikoff:
if you rebase against master the test suite should pass correct?

bantic:
I believe so. I'll give that a shot.

bantic:
Pushed a rebase, hopefully travis is happy now/soon

samselikoff:
woot!

PR #177, opened 2015-06-19
Add ember 1.13.2 scenario

bantic:
I was trying to track down the build failures for #175 and I discovered that on master the "ember-beta" scenario is also failing. It appears to be failing because in that scenario ember-data attempts to read off `indexOf` from `Ember.EnumerableUtils.indexOf`, which no longer exists in ember beta.

This makes a few changes:

  * adds the ember 1.13.2 scenario which also uses ember-data 1.13.2
  * I had to make a small change to the dummy app contacts route error
    handler to account for a difference in the way ember-data 1.13.2
    handles exposes errors (it no longer passes through the jqXHR)
  * Change travis.yml to allow the ember-beta scenario to fail. It is failing anyway because `ember#beta` now points at ember 2.0, whereas when v0.1.4 of mirage was released beta was pointing at 1.13, and that's not incompatible with ember-data.

When I have a chance (in a couple days) I'm going to look at this again and see if I can make it pass w/ ember beta as well. It should be able to do so, it's just a discrepancy between the versions used in the 'ember-beta' scenario. I tried with both ember and ember-data pointing at 'beta' but that didn't work either. The fixes to ember-data's fixture adapter (removing use of `Ember.EnumerableUtils` and `Ember.ArrayPolyFills`) doesn't seem to be on beta yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
 
 env:
   - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=ember-1.13.2
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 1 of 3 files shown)


samselikoff:
Hey, thanks so much for tracking this down! I'm sure it was a pain - you're the man.

PR #188, opened 2015-07-04
Use namespace and glob route for passthrough requests.

cball:
Replaces #155.


samselikoff:
also replaces #134 yes?

PR #190, opened 2015-07-07
Shorthand method with id return 404 when record is missing

cibernox:
p.e: `/users/999` now returns `new Mirage.Response(404, {}, {})` when
the user with that id is not in the database


samselikoff:
thx dude!

seawatts:
Thanks :+1:

PR #191, opened 2015-07-07
Reuse code between undefined and string get shorthands

cibernox:
The undefined handlers can infer the type and just invoke the string handler with it, since the behaviour is the same.


cibernox:
@samselikoff Any feedback on this?

samselikoff:
yep yep - looks great!

PR #193, opened 2015-07-09
replaces usage of Ember.keys with Object.keys to avoid ember@1.13.3 deprecation warnings

makepanic:
Replaces usage of Ember.keys with Object.keys to avoid ember@1.13.3 deprecation warnings:

see https://github.com/emberjs/ember.js/pull/11511


jherdman:
This would effectively break anyone not ready to upgrade yet. Perhaps the following is more prudent for now:

```javascript
var getKeys = Object.keys ? Object.keys : Ember.keys;
```

makepanic:
Currently mirage already uses `Object.keys` in some places: i.e. https://github.com/samselikoff/ember-cli-mirage/blob/afedb357c5d756897da310948700b176e4c0e177/addon/db.js#L93

That means the pull request wouldn't break mirage in browsers where it's already broken. 
I agree we should either feature check `Object.keys` or put a warning message in the readme or something else.

jherdman:
:+1:

samselikoff:
We could also use `_.keys` since we already have lodash

makepanic:
> We could also use _.keys since we already have lodash

Sounds like the best solution. I rebased the commit and it now changes every `*.keys` to `_.keys`

samselikoff:
you're the man!

Issue #46, opened 2015-03-07
ability to use fixtures/factories in dev/tests


samselikoff:
whichever you prefer. shouldn't be too difficult.

seawatts:
What does this mean? Like in unit tests?

samselikoff:
Currently fixtures work in dev (actually in tests too, if you don't have any factories). You only use factories in tests. I want to have a clear path if, e.g. you only want to use factories. So you've set them up for your tests, and you want to be able to set up a default scenario for development too, using your factories to generate some data.

seawatts:
That makes sense. Is there any plan to be able to use this for unit tests?

samselikoff:
ohhh using mirage in unit tests. Hm, hadn't thought about that, what unit test would you need to mock out HTTP requests, your adapter or something?

seawatts:
Lol actually that is exactly what I was talking about. But I could also see it useful just to use the fixtures for other unit tests as well. Not necessarily mocking the http requests, but since we already have the fixtures defined, I would rather not have to define them again.

samselikoff:
hm what would you use the fixtures for, if not http mocks?

varblob:
I'm actually thinking about this at this very moment.  I would really like to use the factories I created to  generate ember data models with default values.  I guess I could use something like https://github.com/danielspaniel/ember-data-factory-guy, but I would like to avoid having to create factories twice.

samselikoff:
yes. The problem is, Mirage is really about http mocks. It actually has nothing (very little) to do with ember, it's just using Ember CLI to wire up the server. I think it's important to keep it that way because core Mirage could very easily be extracted and used e.g. in a react app.

I could be persuaded otherwise, but right now I don't think Mirage has any business creating ember data models. I know it's a drag and perhaps there's room for another layer that will simplify this, but mirage in testing is really about "given this data in the server, here's what my ember app does". The factories are about simply reproducing your API responses. I know you guys know this but again, my current thinking is using the factories for something like making ember data models is out of scope.

varblob:
That makes sense to me to focus on the core of pretending to be a server first.

Issues like generating ember-models could exist in a separate addon or perhaps core would be pushed to a seperate repo and ember bits would remain in this repo?  It would be great to also have some code generation from server languages to mirage especially given that mirage is moving towards having schema/validations/serializer.  I pretty much have factories, serializers, routes etc all specified on my server side anyways.  

[...]

PR #195, opened 2015-07-10
[WIP] testing helpers for using factories in unit tests

ronco:
This provides several test helper methods which can be used in unit & component integration tests.  This will populate the ember data store with the created objects.

Sample usage as follows:

``` javascript
        setStore(store);
        let task = make(contactFactory, 'contact', {
          name: 'foo'
        });
```

If there's interest in bringing this in I can add doc updates as appropriate.

### DIFF
--- a/test-support/helpers/mirage-unit-helpers.js
+++ b/test-support/helpers/mirage-unit-helpers.js
@@ -1,0 +1,71 @@
+import Ember from 'ember';
+
+let sequenceMap = {};
+
+var store;
+var setStore = function(passedStore) {
+  store = passedStore;
+};
+
+var _factoryCreate = function(factory, modelName) {
+  let obj = _factoryBuild(factory, modelName);
+  let record;
+
+  Ember.run(function(){
+    let data = _.defaults(obj, {id: sequenceMap[modelName]});
+    record = store.push(
+      modelName,
+      data
+    );
+  });
+  return record;
+
+};
+
+var _factoryBuild = function(factory, modelName) {
+  if (!sequenceMap.hasOwnProperty(modelName)) {
+    sequenceMap[modelName] = 1;
+  }
+  return factory.build(sequenceMap[modelName]++);
+};
+
+var _getFactory = function(Factory, overrides={}) {
+  let FactoryWithOverrides = Factory.extend(overrides);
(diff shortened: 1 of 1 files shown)


test-support/helpers/mirage-unit-helpers.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+  return factory.build(sequenceMap[modelName]++);
+};
+
+var _getFactory = function(Factory, overrides={}) {
+  let FactoryWithOverrides = Factory.extend(overrides);
+  return new FactoryWithOverrides();
+};
+
+var build = function(Factory, modelName, overrides={}) {
```
ronco: Since `Factory` has no knowledge of the model type the name needs to be provided.  I'm very open to suggestions for improvement here.

ronco:
@blimmer Any interest?

samselikoff:
I'd like to hold off a bit on this, after I finish serializers I will be updating the factory layer. At that point the factory will delegate to `schema`, which knows what models are registered, will be able to look up relationships etc. I think it will simplify things here when that's finished.

ronco:
Gotcha, makes sense.

musaffa:
@samselikoff Is there anyway to use mirage factories in unit and integration tests now?

samselikoff:
If you [manually start mirage](http://www.ember-cli-mirage.com/docs/v0.1.x/manually-starting-mirage/) you can use `server.create` to use the factories to generate mirage db data, just like in an acceptance test. There's no way to get POJOs or Ember Data models from Mirage factories, right now.

musaffa:
@samselikoff Starting an application instance for integration and unit tests is an overkill. As discussed in #46, we would love to reuse the factory-guy factories with mirage.

samselikoff:
You're talking about creating ED models in unit/integration tests, correct? This is planned behavior but is lower priority than the rest of the HTTP mocking stuff, since that's the problem Mirage was built to solve. I do understand it's frustrating, but I can't really give you an estimate on when it will be ready.

samselikoff:
I'm going to close this for now until we can flesh it out a bit more.

If somebody wants to take charge on this issue, let me know (here or in #ec-mirage) and I can help move this forward.

PR #202, opened 2015-07-16
Update to Faker 3.0.0.

### LINKED ISSUE #201: Upgrade to Faker 3.0
I ran into this [error](https://github.com/Marak/faker.js/issues/241) when using `faker.internet.email`, and their solution was to upgrade to faker 3. I upgraded locally in my `bower.json` and don't see any direct problems from the upgrade. Is it safe to upgrade this dependency?

blimmer:
Resolves #201


samselikoff:
:+1:  thanks!!

PR #204, opened 2015-07-17
Expose all of faker via shim.

### LINKED ISSUE #203: not all faker methods are exposed
e.g. `faker.commerce` from the [API Docs](https://github.com/Marak/faker.js#api-methods)

Seems like we should be able to just add our custom `.list` to the main faker object and export everything else.

blimmer:
Resolves #203

### DIFF
--- a/addon/faker.js
+++ b/addon/faker.js
@@ -16,18 +16,6 @@ var list = {
   }
 };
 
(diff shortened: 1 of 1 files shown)


samselikoff:
awesome, thanks dude !

PR #207, opened 2015-07-19
Serializers (nested + sideloaded responses)

samselikoff:
**Attributes**

- [x] whitelist plain attrs
- [x] no root (default), root of model type
- [x] custom #serialize function
- [x] formatting model attribute keys (e.g. camelize)

**Associations**
- [x] whitelist associated attrs
- [x] embedding responses
- [x] sideloaded responses
- [ ] ~~formatting relationship keys~~

**Other**
- [x] lookup appropriate serializer for response
- [x] use application_serializer if exists
- [ ] ~~update shorthands to use serializers, if they exist~~ (just need to update shorthands to return models/collections, if orm is in use
- [ ] ~~json api serializer~~ (separate issue/pr)


addon/serializer.js
```diff
@@ -0,0 +1,37 @@
...
+    } else if (response instanceof Collection) {
+      return this.serializeCollection(response);
+
+    } else {
+      return response;
+    }
+  }
+
+  serializeModel(model) {
+    return this.constructor.root ? { [model.type]: model.attrs } : model.attrs;
```
seawatts: When will this ever be false?

samselikoff: If you override in a subclass, see the [test](https://github.com/samselikoff/ember-cli-mirage/pull/207/files#diff-13e6bf460e46a3755ef874d2806b55e0R68) below

samselikoff:
@seawatts thx for keeping your eye on this! :)

addon/serializer.js
cibernox: Just highlight that this patterns only applies to AMS-like serializers. 
JSONAPI serializer never include the root (or always include the root `data`, depends on how you interpret it). If this class intends to be a superclass of other serializers maybe it's not generic enough.

samselikoff: Excellent point - I was going to focus on getting AMS-like serializers working first, then hopefully use something like [jsonapify](https://github.com/alex94puchades/jsonapify) to convert the data to json api-like response.

I'll think a little more about this.

seawatts:
:+1: No problem!

addon/serializer.js
seawatts: Oh I see! Yeah that makes sense.

ibroadfo:
:heart_eyes_cat:

jsmestad:
What is left with this?

Issue #194, opened 2015-07-10
Support single-use server endpoint overrides


ronco:
It would be ideal to temporarily override server endpoints to return an error, and then reset back to the default behavior provided in `config.js`.

The use case is writing acceptance tests to assert the UI recovers from intermittent server errors.

An example acceptance test would be:

```
visit('/posts/1');
server.forceError('post', '/comments');
fillIn('#comment', 'Hello world');
click('#submit');
andThen(function() {
  assert.equal(find('.error-field').text(), 'Sorry something went wrong, please try again.');
});
click('#submit');
andThen(function() {
  assert.equal(find('.success-field').text(), 'Thanks for your comment')
});
```

ronco:
We are doing this with some home grown helpers (seen in this [gist](https://gist.github.com/ronco/ffaa1c7256990529068b)) that capture the default pretender handler for resetting after use.

The api could probably be cleaned up a bit if directly integrated into `Server`.

samselikoff:
Thanks for kicking off this discussion! I'm wondering if `forceError` is too specific - maybe you want to force something else temporarily, and also have more control over the error code. Maybe we should generalize it to something like

```js
server.override.post('/comments', {error: 'message'}, 404);
server.override('post', '/comments', {error: 'message'}, 404); // or this syntax
```

That way you can use the normal routing api in your temp override. Default would be once, need to think of an api to allow multiple overrides.

```js
server.override('post', '/comments', {error: 'message'}, 404); // single override
server.override(3, 'post', '/comments', {error: 'message'}, 404); // next 3 requests error. is this api necessary?
server.override.twice('post', '/comments', {error: 'message'}, 404);
```

samselikoff:
Only problem is `override` doesn't suggest the old route handler will be restored. Need to think of another name.

```js
server.temp('post', '/comments', {error: 'message'}, 404);
server.tempOverride('post', '/comments', {error: 'message'}, 404);
server.handleNextRequest('post', '/comments', {error: 'message'}, 404);
server.nextRequest('post', '/comments', {error: 'message'}, 404);
server.next('post', '/comments', {error: 'message'}, 404);
server.forNext('post', '/comments', {error: 'message'}, 404);
```

Thoughts? Other ideas?

ronco:
I agree override could give the wrong impression. Maybe just 'force', but
I'm ok with your other ideas.
I'm not sure of the need for multiple, but it might be nice to set
[...]

PR #210, opened 2015-07-21
[wip] Allow handlers to be temporarily overridden

elucid:
Extends the server to accept temporary handlers to be used in place of the default handlers defined in `mirage/config.js`. Provides a simple way to stub out server responses within specific
tests to e.g.
- avoid defining complex default handlers for search, sort, filter, pagination, etc.
- test transient error states (see #194)

Temporary handlers are set via `server.queueResponse(verb, path,
handler)`. They are used only once, on the next request to the
matching verb/path pair.

A sequence of temporary handlers can be queued via successive calls to `server.queueResponse()` using the same verb/path pair. They will be run in the order they were queued. Once all temporary handlers in a verb/path queue have been called, subsequent requests to the verb/path pair will be intercepted by the default handler.

**NOTE: Currently a work in progress**

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -41,6 +41,8 @@ export default class Server {
     });
     this.pretender = this.interceptor; // alias
 
+    this.responseQueues = {};
+
     /*
       Db instance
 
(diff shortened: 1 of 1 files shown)


ronco:
@elucid @samselikoff I'm very interested in seeing this functionality pulled into mirage, any updates?

elucid:
I'm going to have to take a look at the recent changes in mirage and get this working again.

g-cassie:
+1
I would also want to be able to make testing assertions in the temporary handlers to check request data is as expected.  Looking at your code I think this would work without any modifications.

samselikoff:
@rondale-sc expressed renewed interest in this feature. I'm all in favor of adding.

I think the implementation here is sound. I'm wondering if instead of introducing a new API for defining handlers, we can just use the `options` argument that already exists (currently, it only takes `timing`):

```
test('it can recover from an authentication error', function() {
  server.get('/users/current', { some: 'data' }, 500, {
    transient: true
  });

  ...
});
```

samselikoff:
I'm going to close this for now until we can flesh it out a bit more.

If somebody wants to take charge on this issue, let me know (here or in #ec-mirage) and I can help move this forward.

nruth:
@samselikoff I'd like to work on this feature, because we've got tests written like that already that are probably not running the way we thought they were. What needs changing or fleshing out? Docs & using the transient option?

localpcguy:
@samselikoff Like @nruth this is something that is coming up in our app, and curious what needs to happen in order to get this to the point where it would be acceptable? I posted in #ec-mirage also, feel free to respond in whichever place is easier.

PR #218, opened 2015-07-29
Updated faker method for 0.1.6

tomoguisuru:
Resolves ```faker.random.arrayElement is not a function``` error after upgrading from 0.1.5 and using Faker 2.12.1


tomoguisuru:
ignore, npm wasn't updating to the correct faker version

PR #220, opened 2015-07-29
Update contact.js

gogogarrett:
When uncommenting this file the lack of commas would cause a build error. This simply adds the commas in.


samselikoff:
ty!

PR #227, opened 2015-08-04
Depend on pretender 0.9

cibernox:
This should fix most of the memory leak happening in acceptance tests.
Tomorrow I'll perform some more checks to see if there is still
something leaking in my test suite, and is so apply some voodoo I've
learnt today.


samselikoff:
thanks!

PR #229, opened 2015-08-05
Release window.server global on shutdown

cibernox:
I keep my fight agains memory leaks. This one removes the server (and the internal pretender too) when the server is shut down.


samselikoff:
ty brother

PR #234, opened 2015-08-07
Install ember-inflector from NPM

### LINKED ISSUE #221: Replace ember-inflector(bower) with ember-inflector(npm)
Bower package [ember-inflector](https://github.com/fivetanley/ember-inflector-bower) is out of date and seems discontinued.
It causes warnings about deprecation of Ember.Handlebars.helper" (at least for ember 1.13) and will not work after ember 2.0.
In fact, ember-data has bundled npm package [ember-inflector](https://github.com/stefanpenner/ember-inflector), and it's available in browser as `Ember.Inflector.inflector`.
Please use the npm version instead.

alexlafroscia:
- Also updates ember-inflector from version 1.3.1 to 1.9.0
- Prevents installing `ember-cli-mirage` from breaking the build with the latest Ember canary (the reference to `Ember.Handlebars` in the old version of `ember-inflector` was messing with things)

Closes #221

### DIFF
--- a/addon/utils/inflector.js
+++ b/addon/utils/inflector.js
@@ -1,5 +1,4 @@
 import Ember from 'ember';
+export { singularize, pluralize } from 'ember-inflector';
 
-export var singularize = Ember.String.singularize;
-export var pluralize = Ember.String.pluralize;
 export var capitalize = Ember.String.capitalize;
--- a/bower.json
+++ b/bower.json
@@ -13,7 +13,6 @@
     "ember-qunit-notifications": "0.0.7",
     "qunit": "~1.17.1",
     "pretender": "~0.9.0",
-    "ember-inflector": "~1.3.1",
     "lodash": "~3.7.0",
     "Faker": "~3.0.0"
   }
--- a/index.js
+++ b/index.js
@@ -18,7 +18,6 @@ module.exports = {
         type: 'vendor',
         exports: { 'pretender': ['default'] }
       });
-      app.import(app.bowerDirectory + '/ember-inflector/ember-inflector.js');
       app.import(app.bowerDirectory + '/lodash/lodash.js');
       app.import(app.bowerDirectory + '/Faker/build/build/faker.js');
     }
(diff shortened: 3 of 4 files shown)


alexlafroscia:
I noticed that the tests don't pass due to the checks like [these](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/unit/import-files-test-node.js#L22) that look for Inflector.  With the NPM import, should I just remove the references to Inflector from that test?

samselikoff:
Hm, the only reason we're able to export in `utils/inflector.js` is because `ember-inflector` is an ember addon, correct? If so, I don't want to merge this, because everything in `/addon` has nothing to do with ember (i.e. we could extract it to a client-only "Mirage.js" lib). What was the original reason to use npm package instead of bower?

samselikoff:
Ah, because of #221. Hm, I'll have to think more about this, I'd prefer a bower package (for now), or at least something that doesn't get automatically imported bc of ember cli.

alexlafroscia:
Gotcha. I made this PR because of that issue, I'm making a demo app of Ember 2.0 for a meetup and wanted to use Mirage for some sample data.  Turns out that #221 is the reason that it wasn't working for me on Ember 2.0, but grabbing a newer version from NPM fixes it.  I think that installing a newer version of `ember-inflector` from Bower would fix it as well.

joeruello:
Had to turn off mirage for one of our apps because of this. Trying to get 2.0 ready.

samselikoff:
Using the npm package for now, hopefully we can get it as a bower/client package and not depend on Ember's build process eventually.

Thanks dude!

joukevandermaas:
Will this go into a (minor) release, so we can update to ember 2.0? More accurately, approx. when will this be in a release?

Thanks!

samselikoff:
@joukevandermaas yep, going to cut as soon as I get tests passing on 2.0

samselikoff:
update: need to get to work and will try to get this working later tonight. Ran into some more test failures updating to 2.0

PR #239, opened 2015-08-14
Update factory blueprint to pass JSHint

blimmer:
JSHint complains about unused {faker} import in default factory.


samselikoff:
thx!

PR #241, opened 2015-08-15
Upgrade Ember and Ember CLI

blimmer:


.jshintrc
```diff
@@ -30,5 +30,5 @@
   "white": false,
   "eqnull": true,
   "esnext": true,
-  "unused": "vars"
+  "unused": true
```
samselikoff: can we keep this `vars`?

.npmignore
```diff
@@ -1,10 +1,10 @@
 .sass-cache/
 _site/
 bower_components/
-jekyll-tmp/
 tests/
 tmp/
 dist/
+jekyll-tmp/
```
samselikoff: this is for the gh-pages site, ok to keep?

blimmer: Yep, it's still there. My editor must have moved it down.

.jshintrc
blimmer: Yep, was just matching ember-cli standards, but I think it's totally acceptable to do this.

blimmer:
We probably shouldn't merge this until we figure out what changed in Ember-CLI that [caused the node tests to break](https://travis-ci.org/samselikoff/ember-cli-mirage/jobs/75733531#L1026).

It looks like calling `EmberAddon()` without any options now throws an exception:

```
TypeError: Cannot read property 'project' of undefined
      at getMirageAddon (tests/unit/addon-tree-test-node.js:7:18)
      at Context.<anonymous> (tests/unit/addon-tree-test-node.js:30:23)
```

samselikoff:
started to debug this but, the first Addon#treeFor test isn't even passing for me on master. Do all tests on master pass for you?

blimmer:
all except the first node test work on master for me:
```
  Addon
    #treeFor
      1) returns an empty tree in production environment by default //fails
```

samselikoff:
ditto. it's passing on travis tho, strange

blimmer:
Still looking into this. It looks like with the new `ember-cli-build.js` file, it gets passed a set of defaults (which were previously "just there"). I am trying to track down how those get passed in because that should fix the problem that exists here.

tests/unit/addon-tree-test-node.js
```diff
@@ -29,7 +35,11 @@ describe('Addon', function() {
       process.env.EMBER_ENV = 'production';
       var addonTree = getMirageAddon().treeFor('addon');
 
-      expect(addonTree.inputTrees.length).to.be.equal(0);
+      if (addonTree._inputNodes) {
+        expect(addonTree._inputNodes.length).to.equal(0);
+      } else {
+        expect(addonTree._inputNodes).to.be.undefined;
+      }
```
blimmer: This is very strange, but goes along with the behavior you and I were seeing early on master (where `length` was undefined on Mac, but not on Travis' Ubuntu version).

blimmer:
@samselikoff this is passing now.

samselikoff:
awesome, looks great!

samselikoff:
could you squash? makes it easier when updating the changelog if I can look back through history quickly. (I'm fine with multiple commits on more complex PRs)

blimmer:
Makes sense

samselikoff:
perfect, ty!

PR #236, opened 2015-08-10
Adds support for urlPrefix to allow requests to external resources; fixes #228

jamesdixon:
With Pretender v0.9.0 now supporting external resources, this update adds a new `urlPrefix` parameter that allows you to configure Mirage accordingly. For example, if your API lives at `http://localhost:8080`, you can set `urlPrefix` to that url and it will be prepended to any routes configured in `mirage/config.js`.

PR #245, opened 2015-08-16
Allow FQDN in config

blimmer:
This should unblock our release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,6 @@ Check out the [Docs](http://www.ember-cli-mirage.com/docs/latest/)!
 
 ## FAQ
 
-### Known issues
-
-- Pretender doesn't handle other-origin requests (e.g. api.twitter.com)
-
 ### Support
 
 Having trouble? Open an issue!
--- a/addon/server.js
+++ b/addon/server.js
@@ -67,9 +67,16 @@ export default class Server {
   // TODO: Move all this logic to another object (route?)
   stub(verb, path, handler, code, options) {
     var _this = this;
-    path = path[0] === '/' ? path.slice(1) : path;
 
-    this.interceptor[verb].call(this.interceptor, this.namespace + '/' + path, function(request) {
+    var route;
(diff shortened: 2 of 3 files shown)


blimmer:
@samselikoff i'm going to hold off on merging this until you have a chance to confirm the fix.

blimmer:
This is also very similar to @jamesdixon 's #236 , though they are two different implementations. we should figure out which one we want.

jamesdixon:
After looking at your solution, I realized I didn't build in support for using a FQDN directly in the route. That said, I added support for that as well. I'll leave it up to you guys to decide. The one nice thing about being able to set a global prefix is it prevents you from having to set a prefix manually on each route if you happen to be using an external API and don't want to proxy.

Regarding my recent addition, the Travis CI build is still failing on something seemingly unrelated to the updates I made even though everything is passing locally.

samselikoff:
I think being able to have a global prefix is nice, @blimmer any objections?

@jamesdixon could you squash + rebase and we'll investigate the test failures?

blimmer:
As long as we also support a fully qualified url in the config (and that overrides the global prefix), that sounds great! I can imagine a situation where you are mocking an internal api (perhaps with cors) and an external API as well.

jamesdixon:
@blimmer @samselikoff squashed and rebased. The Travis CI build is now passing as well :+1: I had to create a [new PR](https://github.com/samselikoff/ember-cli-mirage/pull/247) as I deleted the master branch of my fork when attempting to squash + rebase and unfortunately, I couldn't reopen the existing PR. Admittedly, it was my first time doing a rebase and ran into a couple of bumps. Let me know if there's anything else you need from me.

samselikoff:
@blimmer yep looked like the test cases supported this case

@jamesdixon rebases are fun :)

jamesdixon:
@samselikoff yeah, definitely a thrilling part of my day :+1:

PR #251, opened 2015-08-19
Implemented addon blueprint for ember-cli-mirage that installs bower …

### LINKED ISSUE #108: Use within development of an addon
Is Mirage usable when developing an addon? We are developing a location service that makes API calls to google, testing it has been difficult because we are making ajax calls and we wanted to override those with mirage.

Samsinite:
…deps and copies default structure into tests/dummy/app. Fixes #108


Samsinite:
Looks like this now creates it in `app/mirage` (from the `ember-cli-mirage` blueprint) as well as in `tests/dummy/app/mirage` (from the `ember-cli-mirage-addon` blueprint). It needs to only install into `tests/dummy/app/mirage` :/. Let me see if I can think of a solution.

Samsinite:
Fixed the issue noticed ^^^ by switching to __root__ and adding the attr under fileMapTokens.

Samsinite:
Also, `tests/dummy/app/(fixtures | factories)/__path__.js` can now be created by specifying the `--dummy` flag. (Ex: `ember g fixture user --dummy` creates `tests/dummy/app/fixtures/user.js`).

blimmer:
ping @samselikoff

samselikoff:
lgtm, thanks so much!

samselikoff:
so, how would one install this in an addon?

Samsinite:
You can run the blueprint like normal, `ember g ember-cli-mirage`, or you can install the addon using `ember install ember-cli-mirage`. Either way will work. For factories and fixtures you'll probably want the `--dummy` option.

Issue #219, opened 2015-07-29
Passthrough Defaults


dgavey:
I looked at issue #131 and #80 and I see that some commits have been made to enable passthrough defaults.  I was wondering when/if those will be merged?  

Also was wondering if there is anyway to enable this now without defining every single route I need to pass through?  I tried the suggestion in #131 but wasn't sure if I should be putting something else in for :fallback or if that was a keyword of some sort.

Thanks for the great addon!

dgavey:
Actually, I was able to enable a default passthrough by doing something like this:

this.pretender.get('/*catchall', this.pretender.passthrough);

However there still seems to be another problem, where only the first url resolves and the Ember promises are stuck on 'resolving'

samselikoff:
Yes `this.pretender.get('/*catchall', this.pretender.passthrough)` should work for now.

I haven't heard of the pending thing..are you putting the `catchall` route at the bottom of your `config.js` file (so your other route definitions match first)?

dgavey:
I don't have any others defined right now.  So even if I don't do a 'catchall' if I put my first two calls  each in a passthrough specifically, the first api call that gets hit, gets the data from my local server. But it never reaches the second api call.  It almost looks like for some reason the first ajax promise is not fully being resolved according to my Ember inspector. Even though I see in the network tab it is being returned.

I'm still trying to determine why that might be.  We don't use Ember-data but an old (abandoned, much hacked) Ember model.  I'm still digging, but any suggestions you have would be appreciated.

dgavey:
And I have the answer! Apparently using passthroughs with pretender requires using Jquery 2 and up.  Found in this [issue](https://github.com/trek/pretender/issues/85) on pretender's github issues.

samselikoff:
bizarre! Great investigative work :mag:

PR #254, opened 2015-08-20
Warn that passthrough only works with jQuery > 2.x

LevelbossMike:
This took me some time to figure out and find in the issues. I suppose this is something that a lot of people need (passthrough option) and might run into.

### DIFF
--- a/docs/v0.0.27/server-configuration.md
+++ b/docs/v0.0.27/server-configuration.md
@@ -50,6 +50,8 @@ export default function() {
 
 Refer to [pretender's docs](https://github.com/trek/pretender#mutating-the-body) if you want to change this or any other options on your pretender instance.
 
+Be aware that the `passthrough`-option wil currently only work with jQuery versions >= 2.x. See this [issue](https://github.com/pretenderjs/pretender/issues/85) for details.
+
 ## Environment options
 
 <a name="enabled" href="#enabled">#</a> ENV['ember-cli-mirage'].<b>enabled</b>
--- a/docs/v0.0.28/server-configuration.md
+++ b/docs/v0.0.28/server-configuration.md
@@ -50,6 +50,8 @@ export default function() {
 
 Refer to [pretender's docs](https://github.com/trek/pretender#mutating-the-body) if you want to change this or any other options on your pretender instance.
 
+Be aware that the `passthrough`-option wil currently only work with jQuery versions >= 2.x. See this [issue](https://github.com/pretenderjs/pretender/issues/85) for details.
+
 ## Environment options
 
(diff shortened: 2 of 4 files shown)


samselikoff:
hm is that true? I just created a brand new 1.13.8 app thats using jQuery 1.11.3 and passthrough works

samselikoff:
maybe it should be `this.pretender.get('/api/v1', this.pretender.passthrough);`

LevelbossMike:
I had this problem in an app of mine today. I had problems with ember-simple-auth + pretender and the promise (for the token endpoint) that should have resolved didn't with jquery 1.x and with 2.x it worked.

Maybe simple-auth is doing something special that doesn't work with passthrough and jquery 1.x but as far as I debugged it it only waits for a successful response and resolves in the hook that wasn't called for me.

I updated jquery after finding this: https://github.com/samselikoff/ember-cli-mirage/issues/219#issuecomment-126106993 which links to this [issue in the pretender repo](https://github.com/pretenderjs/pretender/issues/85) and the issue went away

The use case I have is migrating an app from http-mocks via ember-cli to mirage and we want to do this resource by resource. So I figured other people might run into this as well.

samselikoff:
yah definitely. I see, post, let me try that. thanks for investigating mike!

LevelbossMike:
I have created a repository that demonstrates the problem. I am not sure if I am doing something wrong here but when using passthrough the app does not load unless I install jquery >= 2.x

https://github.com/LevelbossMike/mirage-test

This also fails for a get request and does not seem to be a problem of post requests.

sglanzer:
I hit this problem as well - using mirage for all but one API call (to an ember-cli http-mock that reads from the file system) - passthrough wouldn't resolve the promise until I upgraded to jquery 2.x

samselikoff:
Sorry, I need to review this. Was trying to wrap up other things first but I should carve out some time for this.

sglanzer:
No sweat, easy workaround (for me) - by the way, this addon is so ridiculously useful; really great work!

jkeen:
Ran into the same issue: passthrough wasn't firing the callbacks on jQuery 1.11.3, and upgrading to 2.1.4 after finding the pretender issue fixed the problem. A console warning would be super helpful. :+1:

olivierlesnicki:
having the same issue ... @samselikoff any progress on the review?

samselikoff:
Seems like y'all verified it's true, though I swear I had it working on < 2. I'll merge this in, sorry yall! And thanks for opening issues.

olivierlesnicki:
@samselikoff the root cause is actually a pretender issue: https://github.com/pretenderjs/pretender/issues/85 (for which I'm offering the following fix https://github.com/pretenderjs/pretender/pull/114)

PR #262, opened 2015-09-01
Rename `extend` to `utilsExtend`

seanpdoyle:
Running a project's CI with `phantomjs@~1.9.0` raises the following
error:

```
Exception while running qunit-bdd `before` hook:
`undefined` is not a function
  (evaluating `extend['default'].bind(Model, null)`)
```

`extend` seems to be a pre-existing variable (and will probably be a
keyword in future browsers). Therefore `extend['default']` is
`undefined`.

To work around this, export the `utils/extend` to a different name.

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -1,5 +1,5 @@
 import { pluralize } from '../utils/inflector';
-import extend from '../utils/extend';
+import utilsExtend from '../utils/extend';
 
 /*
   The Model class. Notes:
@@ -197,6 +197,6 @@ class Model {
   }
 }
 
-Model.extend = extend;
+Model.extend = utilsExtend;
 
 export default Model;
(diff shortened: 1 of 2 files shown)


seanpdoyle:
On further inspection, the issue seems to be with `.bind()`.

Either way, `extend` will probably be a reserved word, so I'll leave this open to discussion.

blimmer:
Sounds like you got the `.bind` issue figured out - I'm not sure if we should change this because it might be a problem in the future.

seanpdoyle:
@blimmer could you elaborate on the details of why this rename would be a problem in the future?

blimmer:
Sorry, I think there was a miscommunication. Let me rephrase - I don't think there's a strong case to make this change just because we think it could maybe be a reserved word in the future.

Issue #216, opened 2015-07-27
Pull in a .bind polyfill for PhantomJS


brancusi:
Hi,

Getting a strange error on acceptance tests only in phantomjs. All works fine in chrome. Here are the steps to reproduce:

Using ember-cli: version: 1.13.1

```bash

$ ember init
$ ember g acceptance-test something

```

Update the test and force a simple assert:

```javascript
test('visiting /something', function(assert) {
  assert.ok(true);
});
```

Run the test

```bash
$ ember test
```
Notice everything passes

```bash
$ ember install ember-cli-mirage 
$ ember test --server
```

Phantomjs fails now but chrome is ok.

Am I missing some config or helper? I looked through all the docs but could not find any special setup.

Thanks

samselikoff:
Hmm I wasn't able to reproduce. Node version?

brancusi:
Strange:

ember -v outputs:

```
version: 1.13.1
node: 2.2.1
npm: 2.12.1
```

Thanks

samselikoff:
node 2.2.1? is that iojs or something?

brancusi:
Yes, good point I will try with node. Strange that it would work without mirage though and also with chrome.

The actual error message is:

```
Acceptance: Something: first test
    ✘ beforeEach failed on first test: 'undefined' is not a function (evaluating 'extend['default'].bind(Model, null)')
        
    ✘ afterEach failed on first test: 'undefined' is not an object (evaluating 'target[method]')
```

```javascript
import { module, test } from 'qunit';
import startApp from 'testproject1/tests/helpers/start-app';

var application;

module('Acceptance: Something', {
  beforeEach: function() {
    application = startApp();
  },

  afterEach: function() {
    Ember.run(application, 'destroy');
  }
});

test('first test', function(assert) {
  assert.ok(true);
});
```

Does that shed any light on the issue? Will report back.

samselikoff:
Ahh yes it does, it's that Phantom doesn't support `.bind`. You can pull [this](https://github.com/kdimatteo/bind-polyfill) in for now

mattmcmanus:
The polyfil is only needed for PhantomJS 1.9. Bind is present in 2.0. Unfortunately circlecii comes with 1.9 installed.

brancusi:
Nice, thank you guys! I was indeed running phantom 1.9, all good once upgraded to 2.

samselikoff:
[...]

PR #263, opened 2015-09-01
Include `es5-shim` for PhantomJS/IE8

seanpdoyle:
Versions of `phantomjs` < `2.0.0` [don't include
Function.prototype.bind][bind].

To deal with this, the library could include `es5-shim` in test
environments.

Related to [#262].

[#262]: https://github.com/samselikoff/ember-cli-mirage/pull/262 
[bind]: https://github.com/ember-cli/ember-cli/issues/2634#issuecomment-65326419


jherdman:
See #216

seanpdoyle:
@jherdman good point.

Issue #270, opened 2015-09-03
Make belongsTo and hasMany accept optional types


2468ben:
for example,
`responses: Em.hasMany('comments')`
`user_like: Em.belongsTo('like')`

I'll make a PR if I have energy left after debugging.
In some places around the orm code, (mostly in belongs-to.js and has-many.js), when I want the string 'response_ids' I'll get 'comment_ids' instead, or instead of 'user_like_id' I'll get 'like_id'.

2468ben:
This is definitely supported now, although sometimes setting something like 'responseIds' won't set 'responses' or vice versa. If I can write tests to confirm it's not something dumb on my part I'll make another issue, but I'll close this one.

Issue #266, opened 2015-09-02
object.defineproperty calls in schema.registerModels prevent server re-initialization


2468ben:
Happens when running more than one test in qunit, with or without window.server.shutdown in between.
 
In `schema.registerModels`, some Object.defineProperty getters are created on the model prototypes during the first initialization. If the server's re-created for a second test on the same page, those getters override some calls in `schema.registerModels`'s second runs.

https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/schema.js#L36 was getting picked up by one of these three getters:

https://github.com/samselikoff/ember-cli-mirage/blob/6753698ec3818b6b5e99594a74f764230cdc0d01/addon/orm/associations/has-many.js#L41 (looking for undefined attrs)
https://github.com/samselikoff/ember-cli-mirage/blob/6753698ec3818b6b5e99594a74f764230cdc0d01/addon/orm/associations/has-many.js#L38 (calling isNew() which looks for attrs)
https://github.com/samselikoff/ember-cli-mirage/blob/6753698ec3818b6b5e99594a74f764230cdc0d01/addon/orm/associations/has-many.js#L83
(calling isNew() which looks for attrs)

If I add `if(!this.attrs) { return; }` to the top of those getters, everything runs fine again, but that's not a great solution so I skipped the PR.

samselikoff:
Hm, still seeing this?

Should we do some more cleanup in `server.shutdown`? Something to ensure all the objects are GC'd so we start fresh in new tests?

2468ben:
I was waiting to retry models & serializers due to some work busy-ness, but I'll try to check out in the next few days.

Yeah, the issue was that the defineProperty properties were being accessed on a second run before they could be redefined. I don't know if that's still the case since it's been a few months, but if it is then having a list of keys that mirage can push to whenever it uses defineProperty, then cleaning out that list on shutdown.

2468ben:
I just finally finished getting my whole suite running in the ORM and not seeing this at all, so closing it. Thank you!

samselikoff:
Awesome!

PR #271, opened 2015-09-03
Custom association keys

2468ben:
Here is the code I've changed so far to get a bunch acceptance tests working with #266 and #270.
Most of it is allowing the custom association keys, and the rest are little bits I'll comment on in the code.

I know you're moving and refactoring heavily right now, hopefully some of this is useful if you wanted to fix these things a better way.

### DIFF
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -18,21 +18,23 @@ class BelongsTo extends Association {
     let modelPrototype = ModelClass.prototype;
     var association = this;
     var foreignKey = this.getForeignKey();
+    var relationshipIdKey = key + '_id';
 
     var associationHash = {};
     associationHash[key] = this;
     modelPrototype.belongsToAssociations = _.assign(modelPrototype.belongsToAssociations, associationHash);
     modelPrototype.associationKeys.push(key);
-    modelPrototype.associationIdKeys.push(foreignKey);
+    modelPrototype.associationIdKeys.push(relationshipIdKey);
 
-    Object.defineProperty(modelPrototype, this.getForeignKey(), {
+    Object.defineProperty(modelPrototype, relationshipIdKey, {
 
       /*
         object.parent_id
           - returns the associated parent's id
       */
       get: function() {
-        return this.attrs[foreignKey];
+        if(!this.attrs) { return; }
+        return this.attrs[relationshipIdKey];
       },
 
       /*
@@ -44,7 +46,7 @@ class BelongsTo extends Association {
           throw 'Couldn\'t find ' + association.target + ' with id = ' + id;
         }
 
-        this.attrs[foreignKey] = id;
+        this.attrs[relationshipIdKey] = id;
         return this;
       }
     });
@@ -55,7 +57,7 @@ class BelongsTo extends Association {
           - returns the associated parent
       */
       get: function() {
-        var foreignKeyId = this[foreignKey];
+        var foreignKeyId = this[relationshipIdKey];
         if (foreignKeyId) {
           association._tempParent = null;
           return schema[association.target].find(foreignKeyId);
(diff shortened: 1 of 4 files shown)


addon/orm/associations/belongs-to.js
```diff
@@ -72,15 +74,15 @@ class BelongsTo extends Association {
           - sets the associated parent (via model)
       */
       set: function(newModel) {
-        if (newModel && newModel.isNew()) {
-          this[foreignKey] = null;
+        if (newModel && newModel['isNew'] && newModel.isNew()) {
```
2468ben: the `newModel` that's getting sent in (at least sometimes) is a stripped down POJO, so it couldn't call isNew().

addon/orm/associations/has-many.js
```diff
@@ -114,11 +114,11 @@ class HasMany extends Association {
           schema[association.target].where(query).update(foreignKey, null);
 
           // Save any children that are new
-          models.filter(model => model.isNew())
+          models.filter(model => model['isNew'] && model.isNew())
             .forEach(model => model.save());
 
           // Associate the new children to this model
-          schema[association.target].find(models.map(m => m.id)).update(foreignKey, this.id);
+          schema[association.target].find(models.map(m => m.id || m.attrs.id)).update(foreignKey, this.id);
```
2468ben: Some times that it was called I had to grab the attrs hash for the id. I'm really sorry but I didn't dig back to find the cause of it.

addon/orm/model.js
```diff
@@ -178,7 +180,7 @@ class Model {
     Object.keys(this.belongsToAssociations).forEach(key => {
       var association = this.belongsToAssociations[key];
       var parent = this[key];
-      if (parent.isNew()) {
+      if (parent && parent.isNew()) {
```
2468ben: null check (for when the belongsTo relationship is empty/async).

addon/serializer-registry.js
```diff
@@ -96,21 +96,22 @@ export default class SerializerRegistry {
     // Traverse this model's relationships
     let serializer = this._serializerFor(model);
     serializer.relationships
-      .map(key => model[key])
-      .forEach(relationship => {
+      .filter(key => model[key])
+      .forEach(relationshipKey => {
+        let relationship = model[relationshipKey];
```
2468ben: changed so I can access the relationshipKey and pass it through to the functions below.

2468ben:
I ran into another definedProperty taking over when the orm is re-initialized, and couldn't track down the cause so I had to move back off the orm layer for now. I'd be glad to slack chat on anything I found over the past ~week squeezing my API into it, and/or help out with wherever you plan to go.
I've spent way too much of my ember life on API serializing/deserializing before mirage, so my biggest feature request is a time machine so we could use it in 2013.

samselikoff:
Thanks for looking into this, I'll circle back after I tie up the serializer stuff. I didn't expect anyone to be using the orm yet, I wanted to finish serializers/factories first and then document it. Thanks for giving it a try.

2468ben:
No problem, if you're still ironing out serializers there's only one 'feature' would love to sneak in:

Right now if I add a custom `Serialize.serialize(record, request)`, the request doesn't get passed in yet. If I had that AND a similar function to pick which relationships are included (like `Serialize.getRelationshipKeys(record, request)1, I could toss out all my homegrown serializers.

For example an endpoint like this:
`GET /comments?include_user=true`
```
//serializers/comment.js
getRelationshipKeys(record, request) {
   return request.queryParams.include_replies ? ['user','replies'] : ['replies'];
}
```

samselikoff:
I definitely want to an api that supports inline serializers, which (I think) would let you accomplish that:

```js
this.get('/comments', function(schema, request) {
  let includeReplies = request.queryParams && request.queryParams.include_replies;
  let comments = schema.comment.all();

  return new Mirage.Response({
    content: comments,
    serializer: CommentSerializer.extend({
      relationships: includeReplies ? ['user', 'replies'] : ['replies]
    });
});
```

Another option will always be to just return JSON from your route, which will pass through the serializer layer:

```js
this.get('/comments', function(schema, request) {
  let includeReplies = request.queryParams && request.queryParams.include_replies;
  let comments = schema.comment.all();
  let SpecialSerializer = CommentSerializer.extend({
    relationships: includeReplies ? ['user', 'replies'] : ['replies]
  });

  return new SpecialSerializer().serialize(comments);
});
```

Finally we could also add the option for `relationships` to take a function, which gives you the request:

```js
// serializers/comment.js
  relationships(request) {
    let includeReplies = request.queryParams && request.queryParams.include_replies;
    relationships: includeReplies ? ['user', 'replies'] : ['replies];
  }
});
```

2468ben:
Closing this because #266 and #270 are fixed and these changes aren't needed

PR #272, opened 2015-09-04
Fixed fatal regex error in older Firefox and IE8

jerel:
Upstream: https://github.com/stefanpenner/ember-inflector/pull/89


samselikoff:
thanks!

PR #273, opened 2015-09-09
Remove reliance on lodash global

HeroicEric:
This removes the usage of the lodash `_` global throughout the codebase and imports used modules instead.


blimmer:
I think you can also remove the dependency in [bower.json](https://github.com/samselikoff/ember-cli-mirage/blob/master/bower.json#L16), [index.js](https://github.com/samselikoff/ember-cli-mirage/blob/master/index.js#L21), [blueprints index.js](https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/ember-cli-mirage/index.js#L37).

Additionally, you'll need to modify the blueprint's index to install `ember-lodash` in the consuming application.

I find it helpful to create an example app ([feel free to use mine](https://travis-ci.org/blimmer/ember-cli-example-app-for-github)), [npm link](https://docs.npmjs.com/cli/link) and test everything works there first.

samselikoff:
I'm for removing globals, but I don't know about the explicit imports of each function. `_.[func]` is super convenient - any good arguments for explicit utility imports? Seems like it would make things annoying to develop (need new function, go look up path, add import statement etc.)

john-kurkowski:
I don't know if Babel does this now, but explicit imports could fail faster. In other words, fail at import time, rather than waiting until call time. 

It does also seem to be the Ember codebase's preference when importing other parts of Ember, versus referencing everything on the `Ember` global. Not sure of their reasons.

¯\_(ツ)_/¯

samselikoff:
Good point, and it does seem to be the direction Ember and others are headed.

@HeroicEric, I'm happy to merge this in if you rebase + address blimmer's points.

samselikoff:
@HeroicEric what do you think?

HeroicEric:
@samselikoff sure. I'll make the updates

HeroicEric:
@samselikoff should be all set

samselikoff:
thank you very much!!

HeroicEric:
@samselikoff I think there was some code that you merged in yesterday that will fail now with these removals :panda_face: I'm not sure why CI didn't warn about this.

I just ran locally and some of them are failing but I'm going to make another PR with the fixes

samselikoff:
Yah I saw -I'm taking care of it locally, no problem man!

PR #277, opened 2015-09-10
Fix WHERE db method example

flexyford:
The db.users.where clause was not returning an existing user in the db according to the 'Givens'. We want the documentation to match the 'Given' users for a clear example.


samselikoff:
thank you!

PR #282, opened 2015-09-21
Avoid deprecation warning in Ember.js 2.1

mixonic:
Ember 2.1 introduces the following deprecation message:

https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/application.js#L607

This avoids that message in a backward compatible manner.


samselikoff:
somehow missed this - thanks!

PR #287, opened 2015-09-28
(Docs) Function declarations should not be followed by semi-colons

joeegan:
This caused my jshint to warn me when I dropped it into my app.
The [spec](http://ecma262-5.com/ELS5_HTML.htm#Section_13) to back this up.


samselikoff:
thank you!

Issue #67, opened 2015-03-26
db's collection.find should treat ids as strings not numbers


lukemelia:
For some background info on why treating IDs as strings instead of numbers is a good ideas, this is a good read: https://dev.twitter.com/overview/api/twitter-ids-json-and-snowflake

From an implementation perspective in mirage, we should coerce ids to strings when adding to db, and convert ids to strings when calling `db[collection].find(someId)`

samselikoff:
Thanks for the background, and this totally makes sense to me.

What if someone's API sends up json with an `id` field as an int? Would they not be able to mock our their server using mirage?

lukemelia:
Good question. I suppose another option is to simply not transform IDs at all, and put the onus on users to be consistent.

samselikoff:
I think the default should be, coerce everything to string, and have a config option that disables coercion.

samselikoff:
although, looking at all our rails apps, they definitely serialize ids as ints. I'd imagine this would be the default for most frameworks?

maybe we should default coerce to ints, but have option for coerce to strings, or disable coercion completely.

lukemelia:
@samselikoff an option seems OK to me.

lukemelia:
I think conceptual problem here is that when a `find` operation occurs in a back-end framework, the framework is aware of the stored data type of the ID, and can coerce the incoming string query param value accordingly. That coercion may even vary from model to model. But mirage has no such understanding of the underlying storage type today and so cannot intelligently coerce.

IMO, coercing IDs to strings in all cases would be the most robust solution. Developers who want numeric IDs in their serialized responses can customize the response (eventually through a serializer).

samselikoff:
Makes sense. Seems like serializers are the answer, and when those land
mirage can do whatever it wants internally (coerce to string). Thanks for
input!

ksnyde:
I believe I'm running into this too. What's the best way for us string-id'd folks to proceed?

ksnyde:
BTW, I also wonder if the shorthand should be a bit more robust for singular find operations because when it search for a numeric id and doesn't find it returns `undefined` and ember-data doesn't handle this gracefully. See below:

![2015-06-07_22-43-53_05](https://cloud.githubusercontent.com/assets/980757/8026239/bc65b578-0d66-11e5-9b6e-404ae36d5170.png)

Obviously if I went with the long form I could handle this quite easily myself but I'm a sucker for the short form. :)

samselikoff:
[...]

PR #293, opened 2015-10-01
Fix _findRecord to account for IDs which are numeric but saved as strings

mydea:
When having a database with IDs which are strings but possibly numeric, the _findRecord(id) function would not find a record because the passed in ID is converted to an integer but the record's ID is not.

e.g. if you have a record: 

```json
{ 
    id: "20", 
    name: "Hello" 
}
```

It was impossible to fetch it via its ID.
This pull request will also convert the record ID to an integer the same way the id parameter is converted, which fixes this issue.


samselikoff:
Thanks for the PR! I'm not sure if you've seen https://github.com/samselikoff/ember-cli-mirage/issues/67 for some background on this issue.

Question, how are you getting string ids into your db? Are you using fixture files?

mydea:
I am using factories & fixtures together like this:

```js
// app/mirage/scenarios/default.js

import fixtures from "my-app/fixtures/dummy-data";

export default function() {
    fixtures.users.forEach((user) => {
        server.create("user", user);
    });

    // ...
}
```

And in my fixtures file, I have mixed ids (some strings, some numbers), but I always use strings (e.g. `"1"`).

I am also having issues with the default shorthands, e.g. when using the ´put` shorthand updating does not work correctly because of the string ids.

samselikoff:
yeah. So I think what we should do is coerce all ids to strings when storing in db, and then in your serializer you can change ids to ints if you so choose.

I can make this change after JSONAPISerializer lands, and I document the orm/serializer stuff.

mydea:
Yeah, this sounds good. Looking forward to the Serializer API!

jherdman:
@samselikoff I believe this should be resolved since #462 , aside from the HEAD alias.

samselikoff:
Yes, you're right. Thanks!

PR #295, opened 2015-10-01
Refactor Stringifying in DB-Collection

jherdman:
A common "parse, and then stringify" pattern was repeated everywhere.
This commit extracts this pattern into a helper function.


samselikoff:
beauteous. in fact this is what I was doing to return copies of the data - I wonder if there's a better way?

jherdman:
Totes! We could use [`Ember.copy()`](http://emberjs.com/api/#method_copy).

PR #298, opened 2015-10-05
[FEATURE] `DbCollection#firstOrCreate`

jherdman:
A common pattern I was seeing in our Mirage config was the need to
either find the first record matching a certain query, or to create it.
This change introduce `DbCollection#firstOrCreate`. It returns the first
record that matches a query, or creates it if need be.

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -76,6 +76,20 @@ class DbCollection {
     return records.map( r => stringify(r) );
   }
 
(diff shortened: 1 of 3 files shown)


jherdman:
I cannot replicate this fail locally. Would someone mind punting the build?

samselikoff:
I think its an npm thing, I'll rebuild it.

Btw thanks for this, awesome api!

PR #300, opened 2015-10-05
Update Pretender version to ~0.10.1

bantic:
Changes from version 0.9 include better synchronous xhr support, some fixes for content-type header and event firing on `passthrough`'d requests, better error messages.

See the changes from Pretender 0.10.0 -> 0.10.1 [here](https://github.com/pretenderjs/pretender/blob/master/CHANGELOG.md).


samselikoff:
Yes sir, thank you!

PR #304, opened 2015-10-07
Cookbook example for setting cookies

ronco:


blimmer:
Interesting solution - this lgtm!

samselikoff:
Yes, ditto!

Could you help me understand the part about the XMLHttpRequest spec forbidding cookie headers? Does this mean that an actual ajax request in a real application could never set cookies via headers?

ronco:
@samselikoff Setting headers from ajax requests definitely does work.  I think the spec restriction is just for the XMLHttpRequest javascript implementation.  The browser will read those cookie headers and create appropriate cookies, but XMLHttpRequest itself is forbidden from accessing them/returning them.  I'm not sure if browsers throw errors if XMLHttpRequest returns the cookie headers, or if the pretender.js guys were just being good citizens and obeying the spec.

samselikoff:
I see. I'd like to check with trek et. al. about this first, I wouldn't want to encourage people to do this if it it's bad practice in a real environment. I'll open an issue over there.

samselikoff:
I can't find the part of pretender that ignores this, wasn't there some code in there that regex'd on cookie?

ronco:
@samselikoff Here are the spots it's ignoring it:
 - https://github.com/pretenderjs/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js#L343
 - https://github.com/pretenderjs/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js#L317

samselikoff:
Ah ok, I understand now. Thanks for explaining + diving into this, great work!

PR #305, opened 2015-10-08
Feature: Extended factory type is parent type

jsncmgs1:
To start, let me say that I'm a rails developer learning Ember and JS (also my first OS contribution attempt), so the code I've written could be terrible, please feel free to let me know :).

I wanted to be able to create extended factories which would still be saved in the DB as the parent type.  For example, if I'm on the '/pets' route, I'd like to be able to have factories 'cat' and 'dog' in order to make difference types of pets in my scenario, which would still end up in the 'pet' collection.

If this is a bad idea, cool, if its a good idea with bad implementation, cool, let me know what I should do and I'll fix it.  I'd prefer to write some tests as well but it wasn't super obvious to me where the factory tests should belong.  Thanks!

### DIFF
--- a/addon/factory.js
+++ b/addon/factory.js
@@ -20,7 +20,6 @@ var Factory = function() {
 Factory.extend = function(attrs) {
   // Merge the new attributes with existing ones. If conflict, new ones win.
   var newAttrs = _.assign({}, this.attrs, attrs);
-
   var Subclass = function() {
     this.attrs = newAttrs;
     Factory.call(this);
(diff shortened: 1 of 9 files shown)


samselikoff:
Hey there- thanks for opening this!

Are you familiar with factory_girl? Factory traits are on the roadmap - would that solve your need here? I believe traits is how you'd solve this there. Although maybe this makes sense too.

jsncmgs1:
No problem.  I've used it before but I used Fabrication primarily, I just
took a look at them though, I do think that would solve my needs.

PR #315, opened 2015-10-14
Fixed a typo in setting this.urlPrefix

tomoguisuru:
# What?
A typo fix

# Why?
Because it resulted in using an incorrect setting for the site URL prefix


samselikoff:
thx!

PR #322, opened 2015-10-19
Remove usage of lodash global

HeroicEric:
This one was added in between merging [my other PR](https://github.com/samselikoff/ember-cli-mirage/pull/273) and will fail since the lodash package is no longer included

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -1,8 +1,9 @@
-/* global _ */
 import extend from '../utils/extend';
 import { dasherize, pluralize } from '../utils/inflector';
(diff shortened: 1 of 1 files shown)


samselikoff:
haha, this is in master :) great minds and such.

interestingly I tried to use `Object.assign` first, which worked in Chrome but failed in phantom. I thought this was polyfilled by babel?

ef4:
Yes, Object.assign is in babel's polyfill, but ember-cli-babel only includes the runtime polyfill if you explicitly tell it to. And I don't think addons can easily do that.

samselikoff:
Ah, good to know. Thanks!

Issue #183, opened 2015-06-24
Extract initializer


samselikoff:
Right now Mirage's server boots up in an application initializer. We should extract this logic so users can start their server e.g. in a unit or integration test.

tomoguisuru:
This would help out a lot :+1:

chrisgame:
This would be great for components that access the store!

chrisgame:
Just been speaking to @rwjblue and he suggested the following work around if you need this now...

Create the helper below
```
// tests/helpers/setup-mirage-for-integration.js

import mirageInitializer from '../../initializers/ember-cli-mirage';

export default function setupMirage(container) {
  mirageInitializer.initialize(container);
};
```

Then add the following to your test
```
// tests/integration/components/your-test.js

import mirage from '../../../helpers/setup-mirage-for-integration';

moduleForComponent('your-component', 'Integration | Component | your component', {
  integration: true,
  setup: function() {
    mirage(this.containter);
  }
});
```

samselikoff:
Documented here: http://www.ember-cli-mirage.com/docs/v0.1.x/manually-starting-mirage/

I think this is fine for now

PR #323, opened 2015-10-22
Extract startMirage from initializer

mitchlloyd:
Even though this was partially resolved in https://github.com/samselikoff/ember-cli-mirage/issues/183, I had a situation where I only wanted to start Mirage in specific tests as it was being introduced to a project.

Let me know what you think:
  * Is there a better place to put `startMirage` rather than in the initializer file?
  * Is `startMirage` a good name?
  * Any other testing approaches?

Once this is resolved I'd be happy to update the cookbook documentation as well.

### DIFF
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -11,16 +11,20 @@ export default {
       var container = arguments[0],
           application = arguments[1];
     }
-    let environment = ENV.environment;
 
-    if (_shouldUseMirage(environment, ENV['ember-cli-mirage'])) {
-      let modules = readModules(ENV.modulePrefix);
-      let options = _assign(modules, {environment, baseConfig, testConfig})
-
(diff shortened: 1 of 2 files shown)


samselikoff:
interesting, looks pretty good! I'd like to check with @rwjblue about this, to see if there's a convention going forward for doing this kind of thing.

samselikoff:
I like this PR, any other thoughts on it? Not sure why I punted on it, I think it's better than what we have.

Shall I merge?

mitchlloyd:
Just reviewed this again. Looks like a pretty low risk change, so I'm good with merging it. I should probably follow up with a PR for the cookbook documentation.

samselikoff:
Mucho appreciated, amigo

bargar:
We had the same situation as @mitchlloyd. I had made a similar change locally and started forking when I discovered this PR. In other words, thank you :+1:

Issue #326, opened 2015-10-23
Show routes managed by Mirage


benoittgt:
Hello

I have a rails api, and an ember app that use ember-cli-rails. I want to mock the api only when testing.

Console in browser when running test with `ember test --server` return me this :

```
Mirage: Your Ember app tried to GET 'http://www.lvh.me:3000/api/front/696/users?page=1&per=25', but there was no route defined to handle this request. Define a route that matches this path in your mirage/config.js file. Did you forget to add your namespace?
```

My _config.js_ look like this : 

``` javascript
export default function() {
  this.urlPrefix = 'http://lvh.me:3000';
  this.namespace = '/api/front'; 

  this.get('/696/**', function(db, request) {
    return {
      data: {
        type: 'users',
        id: id,
        attributes: db.users
      }
    };
  })
}
```

I try also with `this.get('/696/users/**', function..` and a really simple route handler :

``` javascript
export default function() {
  this.get('http://www.lvh.me:3000/api**');
}
```

I just have this in my `environment.js` : 

``` javascript
  if (environment === 'development') {
    ENV['ember-cli-mirage'] = {
      enabled: false
    }
```

I've tried also to add `this.pretender.get('/*passthrough', this.pretender.passthrough)
`in my `config.js` like [mention on an another post](https://github.com/samselikoff/ember-cli-mirage/issues/285#issuecomment-142652834) but it didn't change my problem.

I've ask the question [to ember-cli-rails community](https://github.com/thoughtbot/ember-cli-rails/issues/223) because I thought it was partially related but it's not. 

PS: I'm using `"ember-cli-mirage": "0.1.11",`

benoittgt:
I found a way to use it : 
```javascript
this.get('http://www.lvh.me:3000/api/**')
```

and not without the `/` after `api`
```javascript
this.get('http://www.lvh.me:3000/api**')
```
But now I got `application.js:13 Uncaught ReferenceError: I18n is not defined` : 
```
Acceptance | user list: visiting /users (2, 0, 2) 60152 ms
Uncaught ReferenceError: I18n is not defined@ 165 ms
Source: 	
http://localhost:7357/assets/front.js:309
Test timed out
```

So I add to my `config.js` 
```javascript
this.pretender.get('http://localhost:7357/assets/**', this.pretender.passthrough);
```
But I still have this error.

JamieWohletz:
[...]

PR #329, opened 2015-10-23
Throw an Error when Mirage route is not defined

seanpdoyle:
While TDDing, test failure feedback is very important. We prefer loud
failures of quiet ones.

Unfortunately, `console.log` won't cause a test failure while running
`ember test --serve`. If a Mirage route is undefined, tests tend to
hang.

Upon further investigation in a UI-backed browser like Chrome or
Firefox, the `console.log` message is visible.

This feedback is still useful, but we'd prefer to see it in the console,
without having to resort to a visual browser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Ember CLI Mirage Changelog
 
+Changes:
+
+  - [BREAKING CHANGE] missing routes will now throw an Error instead of logging
+    to the Logger's `error` channel.
+
 ## 0.1.9
 
 Update notes:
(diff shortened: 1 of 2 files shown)


seanpdoyle:
@samselikoff I've described this as a `[BREAKING CHANGE]` in the CHANGELOG. Is that the case, or am I being too pessimistic?

samselikoff:
Nope, I think that's right on, as it could break previously passing tests.

Thank you!

geoffharcourt:
LGTM. This is a great step forward in developer-happiness.

benoittgt:
Thanks for this. It should help with [recent issues](https://github.com/samselikoff/ember-cli-mirage/issues/326).

PR #349, opened 2015-10-30
remove dev dependency on es5-shim

csantero:
Removing usage of `function.prototype.bind` means we can get rid of es5-shim.


samselikoff:
what a wonderful PR, thank you very much!

PR #352, opened 2015-11-02
Make lodash a runtime dependency

ef4:
ember-lodash is always needed -- not just when ember-cli-mirage is under development.


samselikoff:
:+1: thank you!

ef4:
I think this ember-lodash PR is also necessary: https://github.com/mike-north/ember-lodash/pull/7

PR #354, opened 2015-11-02
fix shorthand implementation when model has a dash in name

ef4:
Models with dashes in their names end up as camelized properties on the
`schema`, so `my-model` becomes `schema.myModel`. But the shorthand GET
implementation was trying to find them at `schema.my-model`.


ef4:
Oh, I just noticed a better way to fix this, since there is already a `getTypeFromUrl` method. I should fix that instead and make sure it's used here. That will also fix all the other method types.

samselikoff:
Yes, there's a bit more involved here too, since in 0.1.x Mirage used the name of a module to format the result. E.g., if you had `/fixtures/blog_posts.js` your response would be

```
blog_posts: [
 ...
]
```

I think the solution is in the `read-modules` utility, to convert everything there to camelCase. This is a breaking change, but the upgrade path would be to tell everyone to use dasherized filenames. I'm okay with this, because the decision to infer formatting logic from filenames was a bad one. But we'll also need to make sure there's hooks in the serializer layer that will let users get back to their old formatting.

I'm busy right now but this is a good first step.

ef4:
This particular bug doesn't seem to be related to what read-modules is doing, because the model module was already correctly mapped to camel case. It's about mapping URLs to camel case.

I realized that my original change above is still needed to handle the case where the user provides an explicit type name in their handler definition, so it will be additive with a fix to `_getTypeFromUrl`.

I will keep pushing changes to this branch as I get it working for my app, and we can collaborate on making it mergeable when you're ready.

samselikoff:
sounds wonderful. thanks edward!

PR #361, opened 2015-11-03
Apply serializer to any array of Model

ef4:
This ensures that we will correctly serialize any array of `Model`. It's helpful because many natural transforms of `Collection` do not result in a new `instanceof Collection`. For example:

    schema.posts.all().filter(post => post.upVotes > 100)

Without this change, an example like the above returns unserialized results.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -238,7 +238,7 @@ export default class SerializerRegistry {
   }
 
   _isCollection(object) {
-    return object instanceof Collection;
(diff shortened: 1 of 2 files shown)


ef4:
This doesn't cover the empty array case. I suspect the cleanest way to do that will be explicitly passing the top-level type into `serialize`, rather than trying to thread it through as a property on the array.

samselikoff:
I like this. There's also the `where` API off a schema type, but it currently only takes a POJO query:

```js
schema.post.where({ isAdmin: true });
```

I added issue #365 to allow it to take a filtering function.

Issue #356, opened 2015-11-03
Format for fixtures


samselikoff:
With the ORM, Mirage's database has become "opinionated"; that is, it expects keys to be camelCased, foreign keys to be `authorId` and so on. This is so the ORM can work.

It raises an interesting question around what the format of fixtures should be. If someone wants to create a fixture file just to seed their db quickly, we'd probably want it to be a simple array of objects with camelCasing, just like the db expects. If someone wants to copy a response from their actual api - say, a complicated JSON:API document - we'd want the fixture format to be ingested in the same format as their `create` calls (e.g. we'd want to run it through their serializer's `normalize` method).

Which should we do? Should we have an optional toggle so users can use both formats?

ef4:
We could distinguish jsonapi formatted fixtures by requiring them to be valid jsonapi responses, so the top object is a pojo instead of an array.

But to make it extensible, we'd probably need to invoke the appropriate serialized for each type and let it implement a deserialize method.

tomoguisuru:
FWIW, I believe forcing camelCase is a mistake. We use snake case for our model properties and relations because our api is setup to return data in snake case. We made a decision that it was easier to just use what the DB was giving us rather than go through the mess of coveting them when serializing and deserializing.

Forcing the user to use a code style because you believe it is the "correct" way is the wrong way to approach this.

tomoguisuru:
At the very least, there should be a configuration option to use snake_case instead of camelCase.

samselikoff:
Closing due to activity. If someone is facing this issue and motivated to help, let me know!

PR #362, opened 2015-11-03
camelize module names

ef4:
This doesn't break any tests, although [this discussion](https://github.com/samselikoff/ember-cli-mirage/pull/354#issuecomment-153162997) suggests it is probably a breaking change.

This change is necessary to load fixtures for dash-containing model names.

### DIFF
--- a/addon/utils/read-modules.js
+++ b/addon/utils/read-modules.js
@@ -5,6 +5,7 @@
 
 import Ember from 'ember';
 import _keys from 'lodash/object/keys';
+import { camelize } from 'ember-cli-mirage/utils/inflector';
 
 /*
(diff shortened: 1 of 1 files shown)


samselikoff:
Yes, I believe this is the right thing to do though. If anyone ever wanted to instantiate a Mirage server outside the initializer, they should expect to be able to

```js
new Server({
  fixtures: {
    wordSmiths: []
  }
});
```

so I think this is necessary. It will mean users will need to write a serializer for their formatting logic instead of relying on the names of their modules - which I think is a good thing.

samselikoff:
Also I'd like your opinion on #356

PR #364, opened 2015-11-03
Don't mutate model classes

ef4:
Model classes with associations can only be correctly registered with a Schema once, because the registration process mutates them. This change extends them instead so that we don't mutate the originals.

I needed to switch to `for ... in` instead of `Object.keys` because now we definitely need to walk the prototype chain to find associations. This was probably a separate latent bug, since users can put associations onto ancestor classes.

### DIFF
--- a/tests/integration/schema/reinitialize-associations-test.js
+++ b/tests/integration/schema/reinitialize-associations-test.js
@@ -1,0 +1,38 @@
+import Mirage from 'ember-cli-mirage';
+import Model from 'ember-cli-mirage/orm/model';
+import Schema from 'ember-cli-mirage/orm/schema';
+import Db from 'ember-cli-mirage/db';
+import {module, test} from 'qunit';
+
+// Model classes are defined statically, just like in a typical app
+var User = Model.extend({
+  address: Mirage.hasMany()
+});
+var Address = Model.extend();
+
+module('Integration | Schema | reinitialize associations', {
+  beforeEach: function() {
+    this.db = new Db({
+      users: [],
+      addresses: []
+    });
+    this.schema = new Schema(this.db);
+
+    this.schema.registerModels({
+      address: Address,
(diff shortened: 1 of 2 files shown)


tests/integration/schema/reinitialize-associations-test.js
```diff
@@ -0,0 +1,38 @@
...
+    this.db = new Db({
+      users: [],
+      addresses: []
+    });
+    this.schema = new Schema(this.db);
+
+    this.schema.registerModels({
+      address: Address,
+      user: User
+    });
```
samselikoff: By the way in some tests I've started just instantiating `Server`, since the setup is a bit easier (ie test needs to know less about how to wire up schema/db). Wanted to share with you for future reference:

```js
this.server = new Server({
  environment: 'development',
  models: {
    author: Model.extend({
      posts: Mirage.hasMany(),
    }),
    post: Model
  }
});
this.server.timing = 0;
this.server.logging = false;
```

You can then work with schema via `this.server.schema`.

tests/integration/schema/reinitialize-associations-test.js
```diff
@@ -0,0 +1,38 @@
...
+
+    this.schema.user.create({ id: 1, name: 'Link' });
+    this.schema.address.create({ id: 1, country: 'Hyrule', userId: 1 });
```
samselikoff: `id`s are auto assigned by the db. Also you might have been testing something else here, but the "standard" way to create related models using schema will be either

```js
let user = this.schema.user.create({ name: 'Link' });
this.schema.address.create({ country: 'Hyrule', user: user });
```

```js
let user = this.schema.user.create({ name: 'Link' });
user.createAddress({ country: 'Hyrule' });
```

samselikoff:
:+1: thank you!

PR #378, opened 2015-11-06
Expose `id` as sequence value

seanpdoyle:
Expose `id` as sequence value for records with dynamic attributes
 that depend on the `id` -- like the `links` hash:

```js
export default Mirage.Factory.extend({
  links(id) {
    return {
      posts: `/api/users/${id}/posts`,
    };
  }
});
```

Previously, the `sequence` value was an auto-incrementing value starting
from `0`. ID generation behaved exactly the same, but started from `1`.

This commit changes the sequence to generated exactly the same as the
`id`.

Additionally, in the case of the consumer overriding the `id`, the
sequence value defer to the overriding value:

```js
const author = server.create('user', { id: 100 });

assert.equal(author.links.posts, '/api/users/100/posts');
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Changes:
   - [BREAKING CHANGE] missing routes will now throw an Error instead of logging
     to the Logger's `error` channel.
   - [BREAKING CHANGE] Move /app/mirage to /mirage
+  - [FEATURE] The `sequence` value in a dynamic attribute's callback is the same
+    value as the record's `id`. [#378]
+
+[#378]: https://github.com/samselikoff/ember-cli-mirage/pull/378
 
 ## 0.1.9
 
--- a/addon/factory.js
+++ b/addon/factory.js
@@ -2,7 +2,7 @@ import _assign from 'lodash/object/assign';
 import _keys from 'lodash/object/keys';
 
 var Factory = function() {
-  this.build = function(sequence) {
+  this.build = function(id) {
     var object = {};
     var attrs = this.attrs || {};
 
@@ -10,7 +10,7 @@ var Factory = function() {
       var type = typeof attrs[key];
 
       if (type === 'function') {
-        object[key] = attrs[key].call(attrs, sequence);
+        object[key] = attrs[key].call(attrs, id);
       } else {
         object[key] = attrs[key];
       }
--- a/addon/server.js
+++ b/addon/server.js
@@ -126,9 +126,11 @@ export default class Server {
   }
 
   create(type, overrides) {
+    overrides = overrides || {};
     var collection = this.schema ? pluralize(camelize(type)) : pluralize(type);
     var currentRecords = this.db[collection];
     var sequence = currentRecords ? currentRecords.length: 0;
+    var id = overrides.id || sequence + 1;
     if (!this._factoryMap || !this._factoryMap[type]) {
       throw "You're trying to create a " + type + ", but no factory for this type was found";
     }
@@ -136,7 +138,7 @@ export default class Server {
     var Factory = OriginalFactory.extend(overrides);
     var factory = new Factory();
 
-    var attrs = factory.build(sequence);
+    var attrs = factory.build(id);
     return this.db[collection].insert(attrs);
   }
 
--- a/addon/server.js
+++ b/addon/server.js
@@ -138,7 +138,7 @@ export default class Server {
     var Factory = OriginalFactory.extend(overrides);
     var factory = new Factory();
 
-    var attrs = factory.build(id);
+    var attrs = factory.build(typeof id === 'string' ? sequence + 1: id);
     return this.db[collection].insert(attrs);
   }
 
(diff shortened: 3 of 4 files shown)


samselikoff:
Interesting... there are enhancements planned for the factory layer that will allow you to access attrs inside a function via `this`

```js
export default Mirage.Factory.extend({
  links() {
    return {
      posts: `/api/users/${this.id}/posts`
    }
  }
});
```

or perhaps more explicitly, where you specify the dependent attrs you need:

```js
export default Mirage.Factory.extend({
  links: Factory.lazy('id', function(id) {
    return {
      posts: `/api/users/${id}/posts`
    }
  }
});
```

So the only reason I'm hesitant to merge this is because it would be two api changes in relatively short time. Also, some people explicitly use string ids for their DB, and they may still want a simple sequence number in their dynamic factory attributes.

seanpdoyle:
@samselikoff I would prefer your first approach, making `this.id` accessible (and non-`null`/`undefined`).

However, `id` attributes aren't assigned until DB-insert-time, so it's impossible to use `this.id` unless your factory explicitly defined `{ id: 1 }`, which is workable, but tedious.

In this PR, `id` and `sequence` are separate variables. I could modify the code to only ever pass in the number version (whether that's the auto-incrementing `id`, or the `sequence` when a String `id` is specified).

Would that change make this mergible?

geoffharcourt:
This change would definitely remove a lot of boilerplate setup in our current tests.

samselikoff:
Good point, can't access `id` until the object is created. factory_girl gets around this with an `after_create`. We could do something similar:

```js
  afterCreate(obj) {
    obj.links = {posts: `/api/users/${obj.id}/posts`};
    obj.save();
    // or use object.update('link', ...) to do in one line
  }
});
```

Another thought, there's been discussion about storing ids as strings, so that might eventually interfere with this.

I can look at this more a bit later, I definitely want to make your lives easier and since `sequence` is arbitrary this shouldn't make much of a difference for now.

seanpdoyle:
@samselikoff I'm pretty opposed to adding an `afterCreate` hook. 

I'd prefer a simpler solution where `this` exposes everything you'd need, whatever that ends up being.

samselikoff:
Interesting, but then `this` couldn't expose `id`, as you pointed out. Why opposed to `afterCreate`? Do you also oppose this in factory_girl? (genuinely curious, I haven't done *tons* with factory_girl).

seanpdoyle:
@samselikoff I'm a big fan of mirage's simplicity.

Push hash in, get hash out. Introducing hooks would complicate an otherwise simple and minimal interface

samselikoff:
I see. I'm not sure if you've been following recent development but I just wanted to mention that there are some "opinions" coming down the line. The db will remain agnostic about its hashes, but models will hold opinions on (and take care of storing) foreign keys.

You'll be able to continue using mirage with just the unopinionated `db` if you so choose, but you'll end up rewriting a lot of the serialization logic we hope to include (which is totally fine).

Anyways, I only mention this bc I don't know if the simplicity argument will hold forever. But I appreciate your input. I'll check this soon.

seanpdoyle:
Sounds good. Thanks for the hard work :facepunch:

samselikoff:
Thank you :)

So, my concern here is that we're anticipating what the `id` will be, instead of waiting to find out what it actually is. It's true that *right now* the algorithm used to generate `ids` and `sequences` is the same. But that just happens to be the case. Sequence is still just a proxy for id, not the actual thing, so calling it `id` worries me a bit.

We have an app where `id < 1000` is meaningful for the domain object (not the best data modeling, but bear with me). So, developers in this app have a reason to create models with overridden ids. But then, they don't get a sequence number anymore. I think sequence is useful. If they create users with ids of `2000`, `2001`, then `50`, they should still be able to have a `position` attribute with something like `1`, `2` and `3`.

Also, I know the mirage server is tied to a single test right now, but if down the road there's ever a situation where there's multiple reads/writes to the db, you can easily imagine `sequence` no longer tracking `id`.

I just don't see a way around using `afterCreate` to get at the `id`. You have to wait until the db assigns the `id`. I do understand that this would be convenient for y'all, but it just feels too slippery.

Btw I generally think `afterCreate` is a great escape hatch since it's general purpose + people can use it to create related models or what have you. We already have it on the roadmap and since it solves your problem here, I'm inclined to stick with it.

Curious, could I see some of the boilerplate code you mentioned in your current factories?

seanpdoyle:
@samselikoff I've given this more thought, and I'm ok with an `afterCreate` so long as it plays nice with both the factory definition and test `server.create` call.

Maybe something like:


```js
// app/mirage/factories/author.js
export default Mirage.Factory.create({
  afterCreate() {
    this.links = {
      posts: `/api/authors/${this.id}/posts`,
    };
  }
});

// tests/acceptance/some-test.js

const author = server.create('author', {
  name: 'Ernest Hemmingway',
  afterCreate(db) {
    db.posts.insert({ author_id: this.id });
  }
});
```

Does that sound unreasonable or difficult to implement?

seanpdoyle:
Closing in favor of https://github.com/samselikoff/ember-cli-mirage/pull/385

PR #379, opened 2015-11-06
DB Supports Querying with a Filter Function

### LINKED ISSUE #365: Update `schema.where` (and `db.where`) to take a filtering function
Right now `schema.user.where(...)` takes an object query, we should also enable it to take a filtering function.

`schema.user.where(u => u.activityCount > 100)`

jherdman:
Querying can now be done with a function in case you want to do super
spicy things with Mirage! Peep this example:

    let legalDrinkersInCanada = db.users.where(function(user) {
      return user.age > 19;
    });

Boom! Resolves #365

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -207,11 +207,17 @@ class DbCollection {
   _findRecordsWhere(query) {
     let records = this._records;
 
-    for (let queryKey in query) {
-      records = records.filter( r => String(r[queryKey]) === String(query[queryKey]) );
+    function defaultQueryFunction (record) {
+      let keys = Object.keys(query);
+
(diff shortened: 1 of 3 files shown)


samselikoff:
:boom:

this is awesome--thank you!

xamoom-pali:
Amazing!

OakBehringer:
NOTE: This also works for model.where - docs don't state this; might not be obvious to new users.

tumentumurchudur:
This made my day. Thank you.

KeranChen:
This is great!

gbl3:
Excelent solution, thanks!

PR #384, opened 2015-11-09
Throw Errors instead of `console.error`

seanpdoyle:
In the same spirit as [#329].

> While TDDing, test failure feedback is very important. We prefer loud
> failures of quiet ones.
>
> Unfortunately, `console.log` won't cause a test failure while running
> `ember test --serve`. If a Mirage route is undefined, tests tend to
> hang.

> Upon further investigation in a UI-backed browser like Chrome or
> Firefox, the `console.log` message is visible.

> This feedback is still useful, but we'd prefer to see it in the
>  console, without having to resort to a visual browser.

[#329]: https://github.com/samselikoff/ember-cli-mirage/pull/329

### DIFF
--- a/addon/route-handler.js
+++ b/addon/route-handler.js
@@ -60,7 +60,7 @@ export default class RouteHandler {
     try {
       response = handler.handle(request);
     } catch(error) {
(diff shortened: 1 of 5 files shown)


samselikoff:
:heart:

PR #385, opened 2015-11-11
Introduce `afterCreate` hook for Factories

seanpdoyle:
The `afterCreate` hook will be invoked after `server.create` inserts the
record into the DB.

There are two ways to define an `afterCreate` hook:

* In the Factory definition. This will be invoked on each
  `server.create` call:

```js
export default Mirage.Factory.create({
  afterCreate(db) {
    db.posts.update(this.id, {
      links: {
        posts: `/api/authors/${this.id}/posts`,
      },
    });
  },
});
```

* In the overrides passed to the `server.create` call:

```js
server.create('post', {
  afterCreate(db) {
    db.author.insert({ name: 'Tobey Flenderson', post_ids: [this.id] });
  }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changes:
   - [BREAKING CHANGE] missing routes will now throw an Error instead of logging
     to the Logger's `error` channel.
   - [BREAKING CHANGE] Move /app/mirage to /mirage
+  - [FEATURE] Introduce `afterCreate` hooks for Factories.
 
 ## 0.1.9
 
--- a/addon/factory.js
+++ b/addon/factory.js
@@ -9,7 +9,9 @@ var Factory = function() {
     _keys(attrs).forEach(function(key) {
       var type = typeof attrs[key];
 
-      if (type === 'function') {
+      if (key === 'afterCreate') {
+        // no op
+      } else if (type === 'function') {
         object[key] = attrs[key].call(attrs, sequence);
       } else {
         object[key] = attrs[key];
--- a/addon/server.js
+++ b/addon/server.js
@@ -11,15 +11,6 @@ import _isArray from 'lodash/lang/isArray';
 import _keys from 'lodash/object/keys';
 import _pick from 'lodash/object/pick';
 
-function extractAfterCreate(attrs) {
-  attrs = attrs || {};
-
-  var hook = attrs.afterCreate || Ember.K;
-  delete attrs.afterCreate;
-
-  return hook;
-}
-
 export default class Server {
 
   constructor(options = {}) {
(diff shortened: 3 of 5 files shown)


seanpdoyle:
@samselikoff let me know if this was the API you envisioned.

The implementation is a little wacky, so if you have issues with my approach I'll gladly refactor to something more idiomatic to this project.

samselikoff:
Thanks for working on this! I was traveling last week.

The factory definition API looks perfect. Is there a need for supporting it in the overrides to server.create?

seanpdoyle:
@samselikoff I was under the assumption that it'd be useful for setting up ad-hoc relationships (as opposed to having them defined within the factory)

samselikoff:
I see. I'm planning on handling relationships through a new api on Factory, so I'd like to leave this off for now. I know it would make it easier in the short-term, but it's a new API that would only be temporary, so I'd rather not add it in + explain it.

Factory relationships will look something like

```js
// factories/author.js
  name(i) { return `Author ${i}`; },
  include: ['posts']   // this
  posts: hasMany(), // or this
  relationships: { // or something else
    posts: ...
  }
});
```

Need to think through the API for the details, but basically yah I want to hold off on this.

seanpdoyle:
:ok_hand:

samselikoff:
I'm still in favor of this but I'd like to see the `afterCreate` in `server.create` removed (i.e. I'd like users to just define `afterCreate` within a factory). If you find time, could you remove that feature?

Also, I think it would be ideal if the logic invoking the `afterCreate` hook was in `factory.js` rather than `server.js`.

seanpdoyle:
@samselikoff sure thing, but if we remove the call from `server.create`, how will consumers setup relationships within their test files if the hook is only exposed in the factory?

samselikoff:
Ah, I see... my answer is, traits and factory relationships, which is kind of a crappy answer because they don't exist yet. But they are top priority, after 0.2.0 is released.

I guess my question is, what's the benefit of 

```js
server.create('post', {
  afterCreate(db) {
    db.author.insert({ name: 'Tobey Flenderson', post_ids: [this.id] });
  }
});
```

in a test, versus

```js
let post = server.create('post');
db.author.insert({ name: 'Tobey Flenderson', post_ids: [post.id] });
```

? Again I'm mostly hesitant about introducing a new API if we don't have to. `afterCreate` in a Factory definition is totally valid to me, but after factory relationships/traits land, is there any valid use case for `afterCreate` in `server.create`?

samselikoff:
2634942811fc44153ca4525514a08a991dea965f closes

Thanks for kicking this off!

PR #392, opened 2015-11-16
fixes issue with quotes when body is string

bekzod:
```javascript
  this.get('/historydata/', function() {
    return new Mirage.Response(200, {'Content-Type': 'text/csv'},"firstname,lastname\nbob,dylon");
  });
```
was being send as `""firstname,lastname\nbob,dylon""` with extra quotes around them, this fixes it

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -30,7 +30,11 @@ export default class Server {
 
     this.pretender = this.interceptor = new Pretender(function() {
       this.prepareBody = function(body) {
-        return body ? JSON.stringify(body) : '{"error": "not found"}';
+        if (body) {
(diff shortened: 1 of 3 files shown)


samselikoff:
It looks like there are some failing tests. Also, could you add a test case for the issue you've raised?

bekzod:
Ok will do

bekzod:
rebased from master, it turns out it is better to check for it body is string and not serialize it, I didn't know were to put new test let me know if it is in wrong place thanks

samselikoff:
This looks good, could you please squash these 2 commits + rebase?

bekzod:
@samselikoff it is done :)

samselikoff:
Thank you!

PR #399, opened 2015-11-19
Check if relationship exists before trying to persist

tim-evans:
When saving a model with a `null` relationship, mirage throws an error. This allows nullable relationships using `Mirage.Model`. I'm not sure how to test this change.


samselikoff:
thank you!

PR #281, opened 2015-09-20
Shutdown the pretender server on destroyApp.

blimmer:
Ember-CLI 1.13.9 and beyond will have a destroy-app helper that we can shutdown the server in.
This patch automatically inserts this behavior, or shows a message if the user doesn't have a destroy-app helper explaining how to fix the problem.
Fixes #226.

PR #402, opened 2015-11-23
Do not reinitialize mirage config during each test

kamalaknn:
During each acceptance test, the initializer creates a new pretender server and adds all the config, even if a `window.server` exists.

Actually, I faced an issue in my work app, where each acceptance test would add a new server with listeners and eventually hang. I am not able to isolate it into a reproduction.

I believe it is okay to skip doing all the work in subsequent tests, after the initializer has booted the mirage server once.

### DIFF
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -13,7 +13,7 @@ export default {
     }
     let environment = ENV.environment;
 
-    if (_shouldUseMirage(environment, ENV['ember-cli-mirage'])) {
+    if (_shouldUseMirage(environment, ENV['ember-cli-mirage']) && !_isMirageServerRunning(environment)) {
       let modules = readModules(ENV.modulePrefix);
       let options = _assign(modules, {environment, baseConfig, testConfig})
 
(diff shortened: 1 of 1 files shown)


samselikoff:
Im on vacation and can't look at this now, as a stop gap you can add
server.shutdown() to each tests afterEach

kamalaknn:
Thanks @samselikoff , I did try your stop gap solution first. Then wondered if this non-intrusive solution would work.

No problem, I'll wait till you have time to look at this. Enjoy your vacation :smile:

samselikoff:
Thanks for the PR! The problem is, the Mirage db is expected to be fresh at the beginning of each test, so we can't just keep it around, I think.

I believe https://github.com/samselikoff/ember-cli-mirage/pull/281 will solve this problem, but I'll leave this PR open until we get that one merged in.

kamalaknn:
> The problem is, the Mirage db is expected to be fresh at the beginning of each test

Ah, I get it.

> I believe #281 will solve this problem, but I'll leave this PR open until we get that one merged in.

Thanks :+1:

samselikoff:
I think #281 closes this now, in a way consistent with other addons. Let me know if you still think there's a problem! Thank you for pointing this out.

kamalaknn:
I think this will do it. Thank you for the update.

PR #407, opened 2015-11-28
Allow nested factory objects

### LINKED ISSUE #206: Calling functions inside nested objects
Hello,

current Factory implementation execute functions  only if the defined on the top-level of the object https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/factory.js#L6

so, something like this won't ever work 

```javascript 


export default Mirage.Factory.extend({
  first_name: faker.name.firstName,
  last_name: faker.name.lastName,
  company: faker.company.companyName,
  birthday: '11-30-1980',
  photo: '/img/avatar.png',
  emails: [
    {
      id: i => `${ parseInt(Math.random() * 100) }`,
      address: faker.internet.email(),
      type: 'personal'
    }
  ],
  addresses: [
    {
      id: i => `${ parseInt(Math.random() * 100) }`,
      street_address: faker.address.streetAddress(),
      buzzer: faker.address.secondaryAddress(),
      city: faker.address.city(),
      state: faker.address.stateAbbr(),
      zip: faker.address.zipCode(),
      country: faker.address.country(),
      type: 'home'
    }
  ],
  phones: []
});

```

john-kurkowski:
Recursively copies nested properties and calls nested functions in the factory definition. Not always as robust as factory relationships. But sometimes your API response format is trivially nested, and this gets ember-cli-mirage to match it quickly. Fixes #206.

### DIFF
--- a/addon/factory.js
+++ b/addon/factory.js
@@ -1,22 +1,32 @@
 import _assign from 'lodash/object/assign';
-import _keys from 'lodash/object/keys';
+import _isArray from 'lodash/lang/isArray';
+import _isFunction from 'lodash/lang/isFunction';
+import _isPlainObject from 'lodash/lang/isPlainObject';
+import _mapValues from 'lodash/object/mapValues';
 
 var Factory = function() {
   this.build = function(sequence) {
-    var object = {};
-    var attrs = this.attrs || {};
+    const topLevelAttrs = this.attrs || {};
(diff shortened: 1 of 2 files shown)


samselikoff:
Looks like some of this may be solved with embedded serializers, but, I think this won't step on any future APIs. I could be wrong, but I think it looks good.

Thank you!

PR #417, opened 2015-12-07
Improve ID Auto-Incrementation

### LINKED ISSUE #415: Improve DB's autoincrement
Currently the db assigns an `id` based on the `length` of the array. This can cause problems if you delete some models then create a new one. The `id` can clobber an existing model.

jherdman:
Resolves #415

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -20,6 +20,7 @@ class DbCollection {
   constructor(name, initialData) {
     this.name = name;
     this._records = [];
+    this.identityManager = new IdentityManagaer();
 
     if (initialData) {
       this.insert(initialData);
@@ -34,33 +35,24 @@ class DbCollection {
   }
 
   insert(data) {
-    let copy = data ? duplicate(data) : {};
-    let records = this._records;
-    let returnData;
-
-    if (!_isArray(copy)) {
-      let attrs = copy;
-      if (attrs.id === undefined || attrs.id === null) {
-        attrs.id = records.length + 1;
-      }
-
-      records.push(attrs);
-      returnData = duplicate(attrs);
-
+    if (!_isArray(data)) {
+      return this._insertRecord(data);
     } else {
-      returnData = [];
-      copy.forEach(data => {
-        if (data.id === undefined || data.id === null) {
-          data.id = records.length + 1;
-        }
-
-        records.push(data);
-        returnData.push(data);
-        returnData = returnData.map(duplicate);
-      });
+      // Need to sort in order to ensure IDs inserted in the correct order
+      return data
+        .sort(function(a, b) {
+          // typeof ... is to hack around Chrome versus Phantom behaviour
+          if (typeof a.id === 'string' || a.id < b.id) {
+            return -1;
+          } else if (a.id === b.id) {
+            return 0;
+          }
+          else {
+            return 1;
+          }
+        })
+        .map(this._insertRecord.bind(this));
     }
-
-    return returnData;
   }
 
   find(ids) {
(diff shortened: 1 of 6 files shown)


addon/db-collection.js
```diff
@@ -14,6 +17,7 @@ class DbCollection {
   constructor(name, initialData) {
     this.name = name;
     this._records = [];
+    this._collectionId = new IntegerIdCounter();
```
jherdman: In case we ever decide to support UUIDs out of the box.

addon/db-collection.js
```diff
@@ -29,32 +33,25 @@ class DbCollection {
...
-          data.id = records.length + 1;
-        }
-
-        records.push(data);
-        returnData.push(data);
-        returnData = returnData.map( r => stringify(r) );
-      });
+      // Need to sort in order to ensure IDs inserted in the correct order
+      return copy
+        .sort(function(a, b) {
```
jherdman: This kind of makes me feel that the implementation is dancing around a weird edge case. Maybe the implementation should tracked used IDs instead of just being a counter?

jherdman: In a perfect world we wouldn't allow ID assignment any more :)

jherdman:
@samselikoff I think I'm at the point where this is ready for a review. I'm not entirely happy with the solution, but I think it's a decent first kick at the cat.

samselikoff:
Awesome work! :heart:

I'm thinking, 0.2.x is the next release + it will have breaking changes, why don't we just go ahead and throw an error when inserting with multiple IDs?

jherdman:
> why don't we just go ahead and throw an error when inserting with multiple IDs?

As in any noted deprecations should just throw, or the following use case throws?

```javascript
db.vacuumCleaners.insert({ id: 1, ..}, { id: 2 ...});
```

samselikoff:
I think the two use cases you added the `deprecate` messages to (inserting a duplicate `id`, updating the `id` of a record`. I think inserting a record with a specific `id` is still valid - e.g, if they're using clientside generated ids. Do you agree?

jherdman:
Yup! I'll whip up these changes.

jherdman:
So, hey, I'm wondering if we should take this PR one step further and enforce that the ID key _always_ be unique, regardless of what's being used (e.g. UUID, integer counter, strings). If we do this, we could change the underlying data store for records from an Array into a hash-map. This would allow for the find-by-id use case performance to go from O(N) to O(1). Lodash's [`_.values`](https://lodash.com/docs#values) could be used for other uses cases such as filtering, or returning a whole collection.

samselikoff:
Hm, very interesting. I'm wondering, in the spirit of keeping things small, should we deal with that in a separate issue? Would it greatly affect the solution here to have it separate?

jherdman:
> Hm, very interesting. I'm wondering, in the spirit of keeping things small, should we deal with that in a separate issue?

Yeah, I'm inclined to agree with you. I think we're good to go on this pull request! Apologies for the delay, pre-holiday work was crazy busy.

samselikoff:
Ok great - no probs at all on the delay of your free work!

Could we get a quick rebase? then I'll merge

jherdman:
Good to go!

jherdman:
Ugh. Something happened in my rebase, I need a little itme.

samselikoff:
No prob, I'm going to work on some smaller things so just make sure to rebase again when you have time.

jherdman:
@samselikoff apologies for the delay yet again. We're definitely good to go. Since you last looked I've implemented ID tracking regardless of the type. This clears the way to, if it's desirable, or even interesting, make the record store a hash map. I'll open an RFC for discussion sometime soon.

samselikoff:
:+1: thank you for your excellent work on this!

PR #426, opened 2015-12-11
[Docs] Fixes variable width on white panel

gcollazo:
screenshot

### DIFF
--- a/_sass/mixins/_white-panel.scss
+++ b/_sass/mixins/_white-panel.scss
@@ -1,7 +1,7 @@
 @mixin white-panel() {
   background-color: white;
   padding: 20px;
-  max-width: none;
+  max-width: 720px;
   @include above($breakpoint) {
     border-radius: 4px;
   }


samselikoff:
That's bizarre, I'm unable to reproduce. What version of Safari?

`.Container` should have a max-width of 1000px, which should take care of constraining the panel.

gcollazo:
Actually Google Chrome Version 48.0.2564.41 beta (64-bit) on [this page](http://www.ember-cli-mirage.com/docs/v0.1.x/simulating-cookie-responses/).

samselikoff:
Can you inspect and see why `.Container` is not constraining the width?

samselikoff:
I'm seeing this in Canary, it has to do with the `<pre>` tag. We shouldn't just stick a max-width on the white panel, I think. If you'd like to figure out with the flexbox stuff isn't working I'd be happy to merge.

gcollazo:
Tried a few things but couldn't figure out how to fix it without the the `max-width` on the white panel. Feel free to close it.

samselikoff:
Ok. Thanks for your help, hopefully this is a bug in canary since we're not seeing it anywhere else.

PR #429, opened 2015-12-13
use startMirage function consistently

daveworth:
Ensure that in both the component test and the helper we use the same `startMirage` function name.


samselikoff:
Thank you!

PR #430, opened 2015-12-14
Add space in Factory string

daveworth:
This addition means that there is the stated space between `User` and the sequence-generated "id" in the `name` function.

I know this is a tiny nit-pick but it caught my eye!  I hope this is a reasonable (tiny tiny _tiny_) contribution to Mirage as I learn the system.


samselikoff:
:heart: every little bit helps, thank you!

PR #431, opened 2015-12-14
Make link to issue via HTML anchor tag.

daveworth:
It appears that Markdown-style links do not work within HTML tags.  Use HTML
instead!

Before!

After!


samselikoff:
Thanks!

PR #432, opened 2015-12-14
Fixup some small capitalization mistakes

daveworth:
This fixes what appears to be missed capitalization when splitting a long sentence into more tractable ones.


samselikoff:
Thank you for your help David!

PR #433, opened 2015-12-14
Polymorphic relationships

Exelord:
What are you thinking about support of polymorphic relationships?


samselikoff:
Yes, we need that. Thanks for kicking this off - if you want to keep working with it, this PR will need tests around the new behavior.

Exelord:
Moved to https://github.com/samselikoff/ember-cli-mirage/pull/451

PR #435, opened 2015-12-14
Port #429, #430, and #431 to the v0.2.x docs

daveworth:
This is a series of changes made to the v0.1.x docs that should probably be merged into the upcoming release docs.

I did not _back_ update to prior versions but could do the same if that's appropriate?


samselikoff:
I think this is fine for now. Thank you very much, this is very helpful!

PR #438, opened 2015-12-15
fixes bug when `delete` default handler was returning 200

### LINKED ISSUE #408: Delete should return 204 response
I'm trying out 0.2.0-beta.1 and the delete shorthand is not working with EmberData JSON API.  It seems to return a 200 response which causes EmberData to attempt to parse the empty body and throw an error.  If you create your own implementation of delete and return 204 for the response, EmberData is happy again.  For example:

```javascript
  this.delete('/registrations/:id', (schema, request) => {
    schema.registration.find(request.params.id).destroy();
    return {};
  }, 204);
```

bekzod:
fixes #408

### DIFF
--- a/addon/route-handler.js
+++ b/addon/route-handler.js
@@ -118,8 +118,7 @@ export default class RouteHandler {
     if (this.customizedCode) {
       code = this.customizedCode;
     } else {
-      code = this._defaultCodeFor[this.verb];
-
+      code = this._defaultCodeFor(this.verb);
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you - how bizarre, that function was never being invoked? The things hidden in codebases...

:beers:

bekzod:
hehe yep, didn't notice that they were square brackets at first :stuck_out_tongue_closed_eyes: easy to miss, would have left unnoticed if not the bug

PR #441, opened 2015-12-16
Fixes performance regression for createList

### LINKED ISSUE #391: O(N^2) performance when using createList due to get for this.db[collection].
In server.js, `createList` calls `create` for each list item. In `create`, `this.db[collection]` is referenced twice. Because `defineProperty` is used to create the ` get` method, it is used for this access. This copies the entire database to JSON and then parses it back.

So if mirage creates a list of 1000 records, 2,000 calls JSON encode/decodes are made, that last one with an array of 1000 elements.

With 100 records, the delay is imperceptible. With 1000 records, the delay is about 15 seconds.

Neither of the two calls to `get` actually uses the copied database. One just checks the length and the other calls it's bound `insert` method, which access the real database through an instance variable.

Or at least I think that's what's happening by stepping through the code. :)

alvinvogelzang:
fixes #391

`server.create()` is triggering `this.db[collection]` for every record and creates a copy of the collection. This causes a huge performance regression when creating a lot of factories. In this PR we trigger `this.db[collection]` also in `server.createList()` and pass it as an argument in `server.create()`.

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -125,26 +125,36 @@ export default class Server {
     });
   }
 
-  create(type, overrides) {
-    var collection = this.schema ? pluralize(camelize(type)) : pluralize(type);
-    var currentRecords = this.db[collection];
-    var sequence = currentRecords ? currentRecords.length: 0;
+  create(type, overrides, collectionFromCreateList) {
+
+    // Store sequence for factory type as instance variable
+    this.factorySequences = this.factorySequences || {};
+    this.factorySequences[type] = this.factorySequences[type] + 1 || 0;
+
     if (!this._factoryMap || !this._factoryMap[type]) {
       throw "You're trying to create a " + type + ", but no factory for this type was found";
     }
-    var OriginalFactory = this._factoryMap[type];
-    var Factory = OriginalFactory.extend(overrides);
-    var factory = new Factory();
 
-    var attrs = factory.build(sequence);
-    return this.db[collection].insert(attrs);
+    const OriginalFactory = this._factoryMap[type];
+    const Factory = OriginalFactory.extend(overrides);
+    const factory = new Factory();
+
+    const sequence = this.factorySequences[type];
+    const attrs = factory.build(sequence);
+
+    const collectionName = this.schema ? pluralize(camelize(type)) : pluralize(type);
+    const collection = collectionFromCreateList || this.db[collectionName];
+
+    return collection.insert(attrs);
   }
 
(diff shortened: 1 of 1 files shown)


samselikoff:
Thank you for opening this PR. One issue with this is, the sequence number should increment each time either `create` or `createList` is invoked. So using the local variable `sequenceFromCreateList` means, `createList('user', 3)` followed by another `createList('user', 2)`, the latter two users would have sequence numbers of 1 and 2 (when they should be 4 and 5).

I think a better solution to this problem is, not to use `collection.length` at all. If that's the only reason we're using `db.collection`, which is causing problems, why don't we just store the sequence in a hash using the factory's type? I think this is a better solution.

Feel free to run with that here, or I'll get to this as I work.

alvinvogelzang:
@samselikoff ohh yeah of course, didn't think about that but. I'm not sure what you mean by storing the sequence in a hash. Can you explain?

samselikoff:
Meaning, what if we just stored the sequence number on an instance variable on `server`?

```js
create(type, overrides) {
  this.factorySequences = this.factorySequences || {};
  this.factorySequences[type] = this.factorySequences[type] || 0;
  this.factorySequences[type]++;
  let sequence = this.factorySequences[type];
  ...
}
```

alvinvogelzang:
Ready for review

addon/server.js
```diff
@@ -125,26 +125,37 @@ export default class Server {
...
 
-  create(type, overrides) {
-    var collection = this.schema ? pluralize(camelize(type)) : pluralize(type);
-    var currentRecords = this.db[collection];
-    var sequence = currentRecords ? currentRecords.length: 0;
+  create(type, overrides, collectionFromCreateList) {
+
+    // Store sequence for factory type as instance variable
+    this.factorySequences = this.factorySequences || {};
+    this.factorySequences[type] = this.factorySequences[type] || 0;
```
samselikoff: Forgot current implementation was 0-based. Let's just make this line

```js
this.factorySequences[type] = this.factorySequences[type] + 1 || 0;
```

then we can drop line 133, and drop the `- 1` from line 143. I think that's a bit easier to follow.

samselikoff:
@alvinvogelzang thank you, awesome work! Could you squash these? Then I'll get them merged in!

alvinvogelzang:
@samselikoff :+1: thanks for your help as well ;-)

samselikoff:
:clap: well done!

PR #444, opened 2015-12-19
Improve Collection Duplication

jherdman:
Using lodash's `assign` method, we can improve the performance of
duplicating objects. A naive benchmark shows that lodash's `assign`
is considerably faster than the previous approach:

https://jsperf.com/object-dup/2


samselikoff:
:+1:

PR #448, opened 2015-12-22
Upgrade Ember CLI to 1.13.13

jherdman:
Now I'm not harassed about having an incompatible version of Watchman :)


samselikoff:
shweet! need to remove https://github.com/samselikoff/ember-cli-mirage/blob/master/.travis.yml#L14 this line

.jshintrc
```diff
@@ -30,5 +29,5 @@
   "white": false,
   "eqnull": true,
   "esnext": true,
-  "unused": "vars"
+  "unused": true
```
samselikoff: I like keeping this "vars", I find it nice to be able to reference expected params even if they're not used. Am I an outsider on this one? O.o

jherdman: TBH I'm not sure, nor do I understand the motivation for the setting as it is in the first place. That said, you're at the helm, revert for this is on the way!

jherdman:
This fail is a mystery to me...

jherdman:
Boom, we're green! I wonder if ember-beta shouldn't be allowed to fail going forward?

samselikoff:
Yah...I'm trying to remember why we added that. I think it was when CLI was going through more churn, but I'm not sure. I'll make the change!

samselikoff:
(btw, thanks for this!)

PR #451, opened 2015-12-23
Polymorphic relations and different model types

Exelord:
Hey,
I just created new branch for better developing the feature.

So, what i implemented here:
## You can create polymorphic relations

Here is an example of model structure:

``` js
    var Member = Model.extend({
      resource: Mirage.belongsTo('resource', {polymorphic: true}),
    });

    var Resource = Model.extend({
      members: Mirage.hasMany(),
    });

    var Post = Resource.extend();
```

I have to still write a better tests for that because there is a lot of edge cases. But It seams works if you care :)!
## You can change the type of `belongsTo` models if you wish

And here is an example of structure:

``` js
    var User = Model.extend({
      posts: Mirage.hasMany('post', {inverse: 'writer'}),
    });

    var Post = Model.extend({
      writer: Mirage.belongsTo('user'),
      editor: Mirage.belongsTo('user'),
    });

```

> Warning! For now it only works for `belongsTo`
### The problems I have unresolved :(

If i have 

``` js
    var User = Model.extend({
      account: Mirage.belongsTo(),
    });

    var Account = Model.extend({
      users: Mirage.hasMany(),
    });


    this.schema.create('user', {name: 'Exelord'});
    this.schema.create('user', {name: 'Sam'});
```

and I create an associated `Account`:

```
this.schema.create('account', {name: 'Apple', userIds: [1,2]});
```

It drops into infinity loop :/ Probably on model find() methods :/

Have you any ideas? It's also a bug on master branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ jekyll-tmp
 # IDE
 /.idea/
 /*.iml
+.DS_Store
--- a/addon/index.js
+++ b/addon/index.js
@@ -6,11 +6,12 @@ import Serializer from './serializer';
 import HasMany from './orm/associations/has-many';
 import BelongsTo from './orm/associations/belongs-to';
 
-function hasMany(type) {
-  return new HasMany(type);
+function hasMany(type, options) {
+  return new HasMany(type, options);
 }
-function belongsTo(type) {
-  return new BelongsTo(type);
+
+function belongsTo(type, options) {
+  return new BelongsTo(type, options);
 }
 
 export {
@@ -27,5 +28,5 @@ export default {
   Factory,
   Response,
   hasMany,
-  belongsTo
+  belongsTo,
 };
--- a/addon/orm/associations/association.js
+++ b/addon/orm/associations/association.js
@@ -1,6 +1,6 @@
 class Association {
 
-  constructor(type) {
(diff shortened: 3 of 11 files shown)


ajcobo:
About the unresolved problem.
Debugging this type of relationship I noticed two things:

1. The fk of the schemas are being duplicated on creation, so you have 2 belongsTo fks on account. Therefore you have additional loops when reading them.

2. When you create the related models, the keys are set and the associated model updated. This triggers updates on related models recursively.

See:

Update [hasMany] (https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/associations/has-many.js#L123)

which triggers [saveAssociations and update on the children](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L50)

samselikoff:
Just an FYI that this is behind many other issues currently. Unless it starts becoming an issue for many others I'll likely want to tackle it after the 0.2 release.

dodeja:
Would love to have this supported.

samselikoff:
Thank you so much for your work on this. It became clear I needed to rework much of the ORM and it is nearing completion. We'll be adding proper polymorphic support to the new code.

You can follow along in #965.

PR #453, opened 2015-12-24
fixes issue with trailing slash

### LINKED ISSUE #418: Trailing slashes in model URL's
I'm running into a weird issue with trailing slashes. When my app requests `/somecollection/1/`, the `getTypeFromUrl` function doesn't seem to handle this correctly. It returns `1` as the type, which means mirage tries to look for the `1s` collection.

Has anyone else experienced this?

bekzod:
fixes #418


samselikoff:
:heart: Thank you! PRs like this are *so* helpful!

PR #461, opened 2015-12-29
Fixes a Subtle ID Tracking Bug

jherdman:
We were tracking the _next_ value, not the currently used value. Ugh... I should have caught this :/

PR #462, opened 2015-12-30
IDs as Strings

### LINKED ISSUE #67: db's collection.find should treat ids as strings not numbers

jherdman:
Resolves #67. Please see commits for details. Note also that this requires that #461 be merged first (this pull request exposed the bug), and is presently rebased on top of it.

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -1,6 +1,7 @@
 import _assign from 'lodash/object/assign';
 import _isArray from 'lodash/lang/isArray';
 import _isEqual from 'lodash/lang/isEqual';
+import _sortBy from 'lodash/collection/sortBy';
 
 function duplicate(data) {
   if (_isArray(data)) {
@@ -39,18 +40,7 @@ class DbCollection {
       return this._insertRecord(data);
     } else {
       // Need to sort in order to ensure IDs inserted in the correct order
-      return data
-        .sort(function(a, b) {
-          if (a.id < b.id) {
-            return -1;
-          } else if (a.id === b.id) {
-            return 0;
-          }
-          else {
-            return 1;
-          }
-        })
-        .map(this._insertRecord.bind(this));
+      return _sortBy(data, 'id').map(this._insertRecord.bind(this));
     }
   }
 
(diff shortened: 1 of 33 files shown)


jherdman:
Blech. Lots of code churn in this one, it's unavoidable though. It'd be nice to get this in sooner than later as it's likely to cause numerous merge conflicts.

samselikoff:
You're doing the lord's work brother. This looks great. I made a note on the other PR, once we get that in I'm happy to merge this.

Btw, I think because that other commit is part of this PR we could just do this one? It doesn't make a huge difference either way.

jherdman:
> Btw, I think because that other commit is part of this PR we could just do this one? It doesn't make a huge difference either way.

Totally. I'm cool with rolling it into one PR, I figured I'd split it into two in the event this was total garbage :)

samselikoff:
@lukemelia I'd like to get a quick weigh-in from you, if possible. I think this is the right move - it's basically an implementation of [your comment](https://github.com/samselikoff/ember-cli-mirage/issues/67#issuecomment-102394201) from #67.

Don't worry about looking through the entire code diff though, just a quick question. So, I understand that, by ensuring IDs will be stored as strings in Mirage's DB, we'll be able to coerce Mirage's API methods to Always Work: `db.find(1)`, for example, will work without the caller having to know that `1` is actually a string. Same for all the DB methods, and all the Model/Schema methods (`schema.blogPost.find`, etc.).

The one concern I have is when users are calling plain array methods, like filter/sort. Now they need to be aware of the type:

```js
return db.users.filter(user => [1, 3, 4].indexOf(user.id) > -1);
```

Of course, this is not the most realistic example, but it's just to illustrate the point. Although, maybe that is the point, that most of the query methods against IDs will be accounted for by Mirage's `find` APIs. Also, users need to be aware of types when using `:id` from the request itself (e.g., coercing to an INT before matching against the DB), so maybe this is just a fact of life.

What do you think, am I getting the tradeoffs right? I just want to make sure I fully understand so I can communicate the changes to users correctly.

lukemelia:
@samselikoff Yes, I think that captures the tradeoffs well.

samselikoff:
Ok great. Thanks!

PR #467, opened 2015-12-31
serializer accept modelName instead of model

bekzod:
since `_serializerFor` using just `model name` to find serializer, I think it would be better to pass just modelName to it. Also since `isModel` and `isCollection` not related to class I think they should be outside of class


samselikoff:
Yep, conforms with Ember Data's API. Wonderful, thank you!

PR #468, opened 2015-12-31
Adds a mirage-serializer blueprint

xcambar:
Because why not :wink: 

FYI, the blueprint defaults to `JsonApiSerializer`.


samselikoff:
thank you sir!

PR #469, opened 2015-12-31
Add docs to migrate from 0.1.x to 0.2.x

xcambar:
When migrating from `0.1.x` to `0.2.x`, the default folder for Mirage moves from `/app/mirage` to `/mirage`.

Having files in both folders __do__ conflict, I've spent quite some time figuring out why `/mirage/config.js` was not interpreted.
The reason is `/app/mirage/config.js` takes precedence.
The fix is to move everything from `/app/mirage` to `/mirage`.

Maybe a code-based solution should be planned, but a documentation quickfix might help potential users meanwhile.

### DIFF
--- a/docs/v0.2.x/installing-and-updating.md
+++ b/docs/v0.2.x/installing-and-updating.md
@@ -45,3 +45,12 @@ To use beta, run
 ```sh
 ember install ember-cli-mirage@v0.2.0-beta.1
 ```
+
+### Updating to beta
+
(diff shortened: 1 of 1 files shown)


samselikoff:
Oy, yes I am so sorry about this. I have a whole "Upgrading from 1.x to 2.x" guide planned, but I've neglecting it in favor of working on issues. Sorry about that, and thanks for the PR!

PR #470, opened 2016-01-02
adds exports for ActiveModel and JSONAPI serializer to index

### LINKED ISSUE #465: Probably export JSONAPISerializer + ActiveModelSerializer as named exports from index.js
since we're doing it with every other class

makepanic:
fixes #465 

Do we need tests that check for existence of these named exports?

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -3,6 +3,8 @@ import Response from './response';
 import faker from './faker';
 import Model from './orm/model';
 import Serializer from './serializer';
+import ActiveModelSerializer from './serializers/active-model-serializer';
+import JSONAPISerializer from './serializers/json-api-serializer';
 import HasMany from './orm/associations/has-many';
 import BelongsTo from './orm/associations/belongs-to';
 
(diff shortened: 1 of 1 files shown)


samselikoff:
Thank you! Umm, that would probably be a good idea :)

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/issues/471

addon/index.js
```diff
@@ -3,6 +3,8 @@ import Response from './response';
 import faker from './faker';
 import Model from './orm/model';
 import Serializer from './serializer';
+import ActiveModelSerializer from './serializers/active-model-serializer';
+import JSONAPISerializer from './serializers/json-api-serializer';
```
cs3b: shouldn't it be `JsonApiSerializer` to keep consistency (alternative, we should change that also in other places?)

samselikoff: Let's stick with `JSONAPISerializer` to match Ember Data, and if it's different anywhere else, let's change it.

PR #472, opened 2016-01-03
removed references to `this`

bekzod:


samselikoff:
Thanks! As we go forward we could even remove most of the calls where we pass `this` as the last argument by using fat arrows.

Issue #450, opened 2015-12-23
Updates to `Collection`


samselikoff:
- don't extend from array
- make an ES6 class with a private `_models` property that has the array
- implement all relevant array methods. take the normal args, apply them to the underlying `_models` property, and return the collection (so type information is passed)

bekzod:
I guess this could be closed :)

samselikoff:
yessir! :)

PR #477, opened 2016-01-04
new collection 

bekzod:
replaced old collection with new one as mentioned in #450

### DIFF
--- a/addon/orm/collection.js
+++ b/addon/orm/collection.js
@@ -4,52 +4,68 @@ import _isArray from 'lodash/lang/isArray';
   An array of models, returned from one of the schema query
   methods (all, find, where). Knows how to update and destroy its models.
 */
-var Collection = function(modelName, ...args) {
+
+export default class Collection {
+ constructor(modelName, ...args) {
   if (!modelName || typeof modelName !== 'string') {
     throw 'You must pass a `modelName` into a Collection';
   }
+
   this.modelName = modelName;
 
   if (_isArray(args[0])) {
     args = args[0];
   }
-  this.push.apply(this, args);
+  this._models = args || [];
+ }
 
-  this.update = function(key, val) {
-    this.forEach((model) => {
-      model.update(key, val);
-    });
-  };
-
-  this.destroy = function() {
-    this.forEach((model) => {
-      model.destroy();
-    });
-  };
-
-  this.save = function() {
-    this.forEach((model) => {
-      model.save();
-    });
-  };
-
-  this.reload = function() {
-    this.forEach((model) => {
-      model.reload();
-    });
-  };
-
-  this.mergeCollection = function(collection) {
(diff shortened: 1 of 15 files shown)


addon/orm/collection.js
```diff
@@ -4,52 +4,76 @@ import _isArray from 'lodash/lang/isArray';
...
 
-Collection.prototype = Object.create(Array.prototype);
+  filter() {
+    var models = this._models.filter.apply(this._models, arguments);
+    return new Collection(this.modelName, models);
+  }
 
-export default Collection;
+  reload(){
+    this._models.forEach((model) => model.save());
```
samselikoff: `reload`, not `save`

samselikoff:
This is so awesome - thank you for helping this along!

- could you add more unit tests to verify each array method returns an `instanceof Collection`, and also verify that the `collection.modelName` is carried along correctly? 
- I believe the answer to this is no, but could you investigate if it's possible to to get the `collection.[0]` ability? Perhaps by adding Array to Collection's prototype chain? We basically tried to do this before, but now that we're overriding all of the array methods, I wonder if we could grab that `[]` functionality while still carrying along our type information?

bekzod:
* added test, is there any other method that returns new Collection other than `filter` ?
* I am looking at it

bekzod:
didn't know that it was possible, implemented accessing models now like native array

samselikoff:
Woah, this is awesome, I also didn't know this was possible! I think this is the best api we can ask for, given the difficulties involved with actually extending `Array`.

Great work, thanks for this!

bekzod:
awesome :+1:

ballPointPenguin:
This is really great. I'm using it already.

One question:
How might I turn an array of models into a collection?
For example in a route handler I am doing something like:
```
this.get('/projects', (schema, request) => {
  let projectMembers = schema.projectMember.where({ userId: 12345 });
  let projects = projectMembers.map((member) => member.project);
  return projects;
});
```
and if I return an array of models, they throw a wrench in my serializer.

ballPointPenguin:
OK. I can do this:

ballPointPenguin:
Now I am wondering if I should put some logic in the serializer that would check to see if you have an array of instances of a model and do this automatically...

samselikoff:
@ballPointPenguin interesting, you've "escaped the Collection API" because we hadn't thought about mapping to another array of models... I'll need to think more about this. I'm really glad you brought this up.

One thought would be do to something like

```js
this.get('/projects', (schema, request) => {
  return schema.project.where(p => p.projectMember.userId == '12345');
});
```

but I don't this works today, I think `where` delegates the query to the db collection, but it really should operate on the model layer. I'll need to investigate.

As far as you instantiating `Collection`, that's certainly a stop-gap for now, but we need to solve this for everyone. I'll open a new issue.

PR #478, opened 2016-01-04
server class cleanup

bekzod:
it would be easier to read/manage code if creating pretender logic is outside class

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -22,11 +22,11 @@ function createPretender(shouldLog) {
     this.unhandledRequest = function(verb, path) {
       path = decodeURI(path);
       throw new MirageError(
-                      "Mirage: Your Ember app tried to " + verb + " '" + path +
-                      "', but there was no route defined to handle this " +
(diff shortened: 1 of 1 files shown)


addon/server.js
```diff
@@ -9,6 +10,27 @@ import _isArray from 'lodash/lang/isArray';
 import _keys from 'lodash/object/keys';
 import _pick from 'lodash/object/pick';
 
+function createPretender(shouldLog) {
+  return new Pretender(function() {
+    this.handledRequest = function(verb, path, request) {
+      if (shouldLog) {
+        console.log('Successful request: ' + verb.toUpperCase() + ' ' + request.url);
```
jherdman: Would it be a better approach to use a template string here, and on L24's error message?

bekzod: fixed that

samselikoff:
Looks great! Would love for the Pretender create stuff to be in it's own class eventually, I think that's easier to read and maintain than ad-hoc functions.

Good cleanup though! Really appreciate all your help recently :)

PR #484, opened 2016-01-05
moved `extractArguments` to server class

bekzod:
I think `extractArguments` should be inside user-facing API, because only user will be passing variable length arguments  and in `route-handler` arguments become more explicit, need your opinion


samselikoff:
I like that. Maybe eventually we could add an object, `RouterHandlerDefinitionReader` which would be a better home for this function (and testable!).

bekzod:
cool thanks :+1:

Issue #333, opened 2015-10-28
Ensure serializers work with dasherized names


samselikoff:

ballPointPenguin:
This one just bit me. I'll look into it.

PR #485, opened 2016-01-06
ensure serializers work with dasherized names

ballPointPenguin:
This addresses https://github.com/samselikoff/ember-cli-mirage/issues/333.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -225,7 +225,7 @@ export default class SerializerRegistry {
 
   _serializerFor(modelName) {
     let type = modelName ? camelize(modelName) : null;
-    let ModelSerializer = this._serializerMap && (this._serializerMap[type] || this._serializerMap['application']);
+    let ModelSerializer = this._serializerMap && (this._serializerMap[modelName] || this._serializerMap[type] || this._serializerMap['application']);
 
     /*
       TODO: This check should exist within the Serializer class, when the logic is moved from the registry to the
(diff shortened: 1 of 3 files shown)


samselikoff:
Thanks for this!

I think we need to go ahead and tackle this at the `read-modules` level. Basically, I think this `camelize/dasherize` code is getting spread over a lot of places already. If we just ensure that the modules map passed into `Server` is all camelCased, we can at least solve some of this.

Basically, if you imagine instantiating a `Server` manually, you'd want to do

```js
new Server({
  serializers: {
    wordSmith: Serializer.extend({...}),
    blogPost: ...
  }
});
```

instead of dasherized names there. So let's just change it [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/utils/read-modules.js#L43), let's camelize `moduleKey` right there. Then, we'd need to change your test to actually test the `read-modules` bit...maybe the easiest way would be to write a new acceptance test for a multi-word model name, and make sure the response is what you expect.

I know that's a lot! Don't feel like you need to do it :)

ballPointPenguin:
I think that is a sound approach @samselikoff. I actually considered messing with read-modules first before changing my mind (don't recall why). I'll take a stab at it.

ballPointPenguin:
just a note: this also affects fixture file names

samselikoff:
f95aa53262164953f70c970812d8462aa13feab2 closes for serializers, 92d1cc8bd3fb6a812dbcf0d166016581f6278e80 for fixtures.

Thank you for your work!

ballPointPenguin:
awesome!

PR #487, opened 2016-01-06
shorthands cleanup

bekzod:
* moved selecting request handler to constructor
* made more explicit selecting handler
* changed order and removed unnecessary arguments for handler methods

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -1,4 +1,4 @@
-import { singularize, capitalize, camelize, dasherize } from 'ember-cli-mirage/utils/inflector';
+import { singularize, camelize, dasherize } from 'ember-cli-mirage/utils/inflector';
 import _isArray from 'lodash/lang/isArray';
 import assert from 'ember-cli-mirage/assert';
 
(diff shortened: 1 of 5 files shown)


addon/route-handler.js
```diff
@@ -25,10 +24,8 @@ function createHandler({ verb, schema, serializerOrRegistry, rawHandler, options
   let handler;
   let args = [ schema, serializerOrRegistry, rawHandler, options ];
   let type = typeOf(rawHandler);
-  if (type === 'function') {
-    handler = new FunctionHandler(...args);
-  } else if (type === 'object') {
-    handler = new ObjectHandler(...args);
+  if (type === 'function' || type === 'object') {
+    handler = new BaseHandler(...args);
```
samselikoff: I'm wary of using this base class here because, it's a *shorthand* base route handler, but the function/object route handlers aren't shorthands. Shorthand means something specific, it means you either passed a string, array, or nothing (undefined) after the path parameter. So I'm wary of using it for function/object because there's a lot of logic there that doesn't apply.

samselikoff:
I like the idea of setting `this.schema` on the handlers but I don't see a ton of clarity gained from this refactoring.

addon/route-handler.js
bekzod: ok reverted this part

bekzod:
main point of refactoring was to initialize handlers in `constructor` rather than having it created in `handle` method, and since `this.schema` is part of class I think it is better access it directly rather than passing it as argument to each method, and such structor allows further simplification

samselikoff:
Ah, right, no need to switch on `type` at request-time, since we know at construction what type of handler it will be. I like that.

Also `this.schema` makes sense. I remember when I originally wrote these, I wanted every handler to be a pure function. But now many pieces of the system are stateful objects, so I think it makes sense to continue that pattern here (as well as elsewhere in the system).

:+1: lgtm!

samselikoff:
Looks like we need a rebase now

bekzod:
rebased it, :+1:

samselikoff:
woot!

PR #495, opened 2016-01-08
factory `build` cleanup

bekzod:


addon/factory.js
```diff
@@ -4,29 +4,21 @@ import _isFunction from 'lodash/lang/isFunction';
...
+    return value.map((val) => buildValue(val, value, sequence));
+  } else if (_isPlainObject(value)) {
+    return _mapValues(value, (val) => buildValue(val, value, sequence));
+  } else if (_isFunction(value)) {
+    return value.call(parentAttrs, sequence);
+  } else {
+    return value;
+  }
+}
+
```
samselikoff: I'm finding this code difficult to read. In the future I'd like to see

 - `Factory` become an ES6 class, so we can easily add new methods on it
 - fewer named functions divorced from the class

bekzod: I will have look what can be done tmrw, for now you can close the PR if you want

PR #496, opened 2016-01-08
added `passthroug request` logging

bekzod:
since `unhandled` and `handled` requests being logged I thought it would be logical to add logging for `passthrough requests` also


samselikoff:
Ah cool, I think this is a newer API of Pretender I didn't know about. Thanks!

PR #500, opened 2016-01-11
Shorthand refactor

bekzod:
* identifying `modelName` from path not from request url
* moved identifying `modelClass` to constructor

### DIFF
--- a/addon/route-handler.js
+++ b/addon/route-handler.js
@@ -21,9 +21,9 @@ function isNotBlankResponse(response) {
 
 const DEFAULT_CODES = { get: 200, put: 204, post: 201, 'delete': 204 };
 
-function createHandler({ verb, schema, serializerOrRegistry, rawHandler, options }) {
+function createHandler({ verb, schema, serializerOrRegistry, path, rawHandler, options }) {
   let handler;
-  let args = [ schema, serializerOrRegistry, rawHandler, options ];
+  let args = [ schema, serializerOrRegistry, rawHandler, path, options ];
   let type = typeOf(rawHandler);
   if (type === 'function') {
     handler = new FunctionHandler(...args);
@@ -43,11 +43,11 @@ function createHandler({ verb, schema, serializerOrRegistry, rawHandler, options
 
 export default class RouteHandler {
 
-  constructor({ schema, verb, rawHandler, customizedCode, options, serializerOrRegistry }) {
+  constructor({ schema, verb, rawHandler, customizedCode, options, path, serializerOrRegistry }) {
     this.verb = verb;
     this.customizedCode = customizedCode;
     this.serializerOrRegistry = serializerOrRegistry;
-    this.handler = createHandler({ verb, schema, serializerOrRegistry, rawHandler, options });
+    this.handler = createHandler({ verb, schema, path, serializerOrRegistry, rawHandler, options });
   }
 
   handle(request) {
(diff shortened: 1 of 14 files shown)


bekzod:
also `shorthand class` always expects shorthand parameter passed

bekzod:
now since `modelClass` are identified in initialization time not in runtime, we can throw errors in initialization time about missing `modelClass`es  for that path

addon/route-handler.js
```diff
@@ -41,12 +42,26 @@ function createHandler({ verb, schema, serializerOrRegistry, rawHandler, options
...
+}
+
 export default class RouteHandler {
 
-  constructor({ schema, verb, rawHandler, customizedCode, options, serializerOrRegistry }) {
+  constructor({ schema, verb, rawHandler, customizedCode, options, path, serializerOrRegistry }) {
     this.verb = verb;
     this.customizedCode = customizedCode;
     this.serializerOrRegistry = serializerOrRegistry;
+    rawHandler = rawHandler || getModelName(path);
```
samselikoff: The problem with putting this here is that `getModelName` is a special "ability" of the shorthand route handlers. So let's keep it within the shorthands. It shouldn't be here, because function and object handlers have no notion of a modelName for a path.

So instead, let's pass along `path` through `createHandler`, give it to the shorthand route handlers, and then they can do their special magic probably in base shorthand. It will still be at construction time (you could even set it as state on the route handler, `this.modelClass = this.getModelClassFromPath(path)`.

Also I prefer these functions to be methods on the class, I find it easier to notice when a new object needs to be made if there are a lot of related methods on the class. The free-standing functions makes that a bit harder, I think.

bekzod: Cool made those changes, I think all of them valid

tests/integration/route-handlers/post-shorthand-test.js
```diff
@@ -33,17 +33,6 @@ module('Integration | Route Handlers | POST shorthand', {
...
-  let request = {requestBody: JSON.stringify(this.body), url: '/authors'};
-  let handler = new PostShorthandRouteHandler(this.schema, this.serializer);
-
-  let model = handler.handle(request);
-
-  assert.equal(this.schema.db.authors.length, 1);
-  assert.ok(model instanceof Model);
-  assert.equal(model.modelName, 'author');
-  assert.equal(model.firstName, 'Ganon');
-});
```
samselikoff: Why remove this test?

addon/route-handlers/shorthands/base.js
```diff
@@ -50,48 +53,30 @@ export default class BaseShorthandRouteHandler {
...
-    let [ urlSplit ] = url.split('?');
-    let path = urlSplit.split('/');
-    path = path[path.length - 1] === '' ? path.slice(0, path.length - 1) : path; // when trailing slash
-    let typePath = hasId ? path[path.length - 2] : path[path.length - 1];
-    let modelName = dasherize(camelize(singularize(typePath)));
-
-    return modelName;
-  }
-
   _getJsonBodyForRequest(request) {
```
samselikoff: If we're moving the call to `normalize` here this method name isn't really accurate anymore, can we change it to `getJsonApiDocForRequest`?

samselikoff:
Just a few more questions, then if you could squash + rebase we can get this merged in!

bekzod:
returned that test and renamed method

samselikoff:
Awesome, looks great. One more rebase and I'll get this in (sorry - I must have merged another PR in that popped up first).

bekzod:
rebased

samselikoff:
:+1: thanks!

Issue #504, opened 2016-01-17
Can't define inverse relationships


thomasdashney:
First of all, awesome project :)

I'm hoping to have an inverse relationship (as supported in Ember Data). For example:

```javascript
// in /mirage/models/User.js
export default Model.extend({
  projects: hasMany()
});
```

```javascript
// in /mirage/models/Project.js
export default Model.extend({
  user: belongsTo()
});
```

When I try to find a record through the schema, I get a maximum call stack exceeded. Is this something this add-on plans to support? Happy to contribute a PR if so! (created a failing test for this case)

samselikoff:
Thank you! :smile:

This absolutely should work today, a PR with a failing test would be ideal. We could also start with a snippet of the code that's giving the error.

samselikoff:
Ahhh... I believe it's the one-to-one that's causing the problem. (The code in your comment above, with one hasMany and one belongsTo should work).

I need to add an option to let the user specify which model owns the foreign key.

Thank you so much for bringing this up!

samselikoff:
Do you have a 1-1 model in your server you're mocking out?

thomasdashney:
It fails with a one-to-many relationship as well–I just pushed another failing test case for that into my branch.

I'm not developing this against a server.. I just started prototyping an app a few days ago and am probably not going to actually develop a server for a bit. Mostly using ember-cli-mirage for some text fixtures to get up and running quickly! Working like a charm so far

samselikoff:
In your db setup, you only need one foreign key. Mirage assumes the `belongsTo` owns the foreign key.

In your second test, could you remove the `projectIds` from the user and see if that fixes it? `user.projects` should still work, since it looks for all projects with a matching `userId`.

Also, in your first test, remove the `addressId` from the user.

thomasdashney:
Ok, just made the test changes and one-to-many works now. Still having a problem with the one-to-one (`address.user` should exist right?)

samselikoff:
Yah I believe one-to-one won't work right now, I need to allow user to say "this model owns the fk of this one-to-one relationship". For now I'd say stick to a one-to-many, if you can.

thomasdashney:
Cool, will do. Thanks for your help :smile:

samselikoff:
Let me know if you have any more questions

ivanbanov:
Any other workaround for a while for one-to-one relationship? I'm having the same problem here

samselikoff:
[...]

PR #505, opened 2016-01-17
WIP: Inverse relationships

thomasdashney:
See #504


tests/integration/schema/inverse-relationships-test.js
```diff
@@ -0,0 +1,35 @@
...
+    user: Mirage.belongsTo()
+  });
+
+  var schema = new Schema(db);
+  schema.registerModels({
+    user: User,
+    address: Address
+  });
+
+  var address = schema.address.find(1);
```
thomasdashney: It seems to fail at the `find()` call

PR #510, opened 2016-01-19
patching pretender so it does not store requests

bekzod:
attempt to solve #510

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -47,6 +47,14 @@ function createPretender(server) {
   });
 }
 
+function patchPretender(pretender) { // according to samselikoff/ember-cli-mirage/issues/509
+  let noopArray = { push: function(){} };
+  pretender.handledRequests = noopArray;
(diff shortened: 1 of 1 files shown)


Leooo:
@samselikoff we are fighting for months because we have memory leaks in our tests "somewhere". If this PR helps in any way, can it be merged?

samselikoff:
I don't want to monkey patch Pretender here, and others might be relying on those APIs from the pretender instance.

Could you help by moving the [upstream PR](https://github.com/pretenderjs/pretender/pull/141) along? Once we get the option there it will be simple to expose it on Mirage.

Also if you need something immediately, you can run the above code in an initializer or your moduleForAcceptance.

Leooo:
@samselikoff I've no Idea how to modify tests in the pretender repo unfortunately, see https://github.com/pretenderjs/pretender/pull/141.

samselikoff:
Going to wait for Pretender to add this and set a default from within Mirage

PR #512, opened 2016-01-19
made passthrough not overridable

### LINKED ISSUE #400: route handler precedence and passthrough
Hi, I would like to use the passthrough api. However it only works when I remove an existing handler for the same route. (I would like to passthrough some routes in a test environment to gradually workout the api while the rest is still being mocked). The documentation has a comment 

```
Put all passthrough config at the bottom of your config.js file, to give your route handlers precedence.
```

which suggests that putting the passthrough at the top of the handler definitions prevents the handler from being called. This does not seem to be the case. So for example:

```
this.passthrough('/posts/:id');
// this.get('/posts/:id', (db, req) => ({ data: db.posts.find(req.params.id) }));
```

works fine, but the following does not.

```
this.passthrough('/posts/:id');
this.get('/posts/:id', (db, req) => ({ data: db.posts.find(req.params.id) }));
```

Is this the expected behaviour? So for my use-case I would need the following?

```
import ENV from 'app/config/environment';
ENV.environment === 'test' 
? this.passthrough('/posts/:id') 
: this.get('/posts/:id', (db, req) => ({ data: db.posts.find(req.params.id) }));
```

bekzod:
should fix #400

### DIFF
--- a/tests/integration/passthrough-test.js
+++ b/tests/integration/passthrough-test.js
@@ -202,3 +202,45 @@ test('it can passthrough other-origin hosts', function(assert) {
   });
 });
 
+
+test('passthrough requests follow normal precedence rules 1', function(assert) {
+  assert.expect(1);
+  var done1 = assert.async();
+  var server = this.server;
+
+  server.loadConfig(function() {
+    this.urlPrefix = '/vet/123';
+    this.passthrough('/pets');
+    this.get('/pets', function() { return 123; });
+  });
+
+  $.ajax({
+    method: "GET",
+    url: "/vet/123/pets",
+    error: function() {
+      assert.ok(true);
+      done1();
+    }
+  });
+});
+
+test('passthrough requests follow normal precedence rules 2', function(assert) {
+  assert.expect(1);
+  var done1 = assert.async();
+  var server = this.server;
+
+  server.loadConfig(function() {
+    this.urlPrefix = '/vet/123';
+    this.get('/pets', function() { return 123; });
+    this.passthrough('/pets');
+  });
+
+  $.ajax({
+    method: "GET",
+    url: "/vet/123/pets",
+    error: function() {
+      assert.ok(true);
+      done1();
+    }
+  });
+});
--- a/addon/server.js
+++ b/addon/server.js
@@ -47,6 +47,12 @@ function createPretender(server) {
   });
 }
 
+function hasPassthrough(pretender, path, verb) {
+  let recognized = pretender.hosts.forURL()[verb.toUpperCase()].recognize(path);
+  let match = recognized && recognized[0];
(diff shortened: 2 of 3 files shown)


tests/integration/passthrough-test.js
```diff
@@ -202,3 +202,25 @@ test('it can passthrough other-origin hosts', function(assert) {
...
 });
 
+
+test('passthrough is not overridable', function(assert) {
+  assert.expect(1);
+  var done1 = assert.async();
+  var server = this.server;
+
+  server.loadConfig(function() {
+    this.urlPrefix = '/vet/123';
```
samselikoff: why do you specify the `urlPrefix` here? As it is written this test will actually also pass on `master`, because of this. I think we can remove it.

tests/integration/passthrough-test.js
```diff
@@ -202,3 +202,25 @@ test('it can passthrough other-origin hosts', function(assert) {
   });
 });
 
+
+test('passthrough is not overridable', function(assert) {
```
samselikoff: could we rename the test to `passthrough requests follow normal precedence rules`

samselikoff:
With the changes made this will be good to go!

Again, thank you so much for all your work lately on Mirage, it's really helping me keep up momentum.

tests/integration/passthrough-test.js
bekzod: I think wanted to test if it works with `fullPath` but actually forgot to put `/vet/123/pets` in ajax url

bekzod:
Cool, whenever got free time at hand trying to contribute

samselikoff:
Looks great. Just a thought, can we add a second test for the opposite case, i.e. make sure if `this.get` is defined first that it wins?

bekzod:
added another test

tests/integration/passthrough-test.js
```diff
@@ -202,3 +202,45 @@ test('it can passthrough other-origin hosts', function(assert) {
...
+
+test('passthrough requests follow normal precedence rules 2', function(assert) {
+  assert.expect(1);
+  var done1 = assert.async();
+  var server = this.server;
+
+  server.loadConfig(function() {
+    this.urlPrefix = '/vet/123';
+    this.get('/pets', function() { return 123; });
+    this.passthrough('/pets');
```
samselikoff: In this case, I'd expect the first handler (the mirage handler `.get('/pets')`) to "win", since it's defined first. I think you may need to update your strategy here to accommodate this.

I don't want you to have to re-invent Pretender's route handler registry... perhaps this is something we should fix upstream in Pretender itself?

samselikoff:
Closing due to inactivity

PR #515, opened 2016-01-19
always pass Serializer.serialize the request as a second argument

2468ben:
http://www.ember-cli-mirage.com/docs/v0.2.0-beta.3/serializers/#serializeresponse-request

Sometimes while on the way to the serializer's serialize function, the request argument gets left behind. These changes should make sure it gets there.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -78,9 +78,9 @@ export default class SerializerRegistry {
         let serializer = this._serializerFor(collection.modelName);
 
         if (serializer.embed) {
-          json[pluralize(collection.modelName)] = this._serializeModelOrCollection(collection);
+          json[pluralize(collection.modelName)] = this._serializeModelOrCollection(collection, request);
         } else {
-          json = _assign(json, this._serializeSideloadedModelOrCollection(collection));
+          json = _assign(json, this._serializeSideloadedModelOrCollection(collection, request));
         }
 
         return json;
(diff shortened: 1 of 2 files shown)


samselikoff:
Good catch... how did you discover this? Was there a particular bug? Ideally we'd think of a failing test along with these updates.

If users need to override `#serialize` they should *always* get the request... I think it got left behind in a few of these methods, simply because they are implementation details of the default serialize handler, and they don't rely on data from the `request` object (as opposed to, for example, the json-api-serializer).

There's a refactoring that needs to happen, moving most of these methods into the `serializer` class (they are really the job of the serializer, not the registry). In any case, I'm interested to hear how this came up.

2468ben:
This came up when I was trying to access request.queryParams an AMS serialize. I'll add a test to cover it.
Here's my current use case:
```
export default ActiveModelSerializer.extend({
  serialize(response, request) {
    if (response instanceof Model) {
      const _attrs = this._attrsForModel(response);
      if(request.queryParams.include_stats) {
        _attrs.stats = /* some stats building function */
      }
      return _attrs;
    }
    return response;
  }
});
```
I have to use the private serializer._attrsForModel to save redoing all of that logic myself, so ideally it would be cool to have that as a protected function instead. and I agree like in #516 that something like the inline serializer API would work great for this.

ballPointPenguin:
I had a very similar experience, trying to access query params in an AMS serialize method.

2468ben:
test added!

samselikoff:
Looks great - thank you!

PR #516, opened 2016-01-19
Includes takes an optional function

2468ben:
This relies on the "fix" in #515.

Based on the very last code snippet you added here: https://github.com/samselikoff/ember-cli-mirage/pull/271#issuecomment-138127734, I added an optional function at `Serializer.includesForRequest(request)` that, if it's defined, gets preference over the static `Serializer.include` prop.

I just arbitrarily picked a name and made it an optional function (so it's a non-breaking change), but tell me whatever you'd ideally like and I'll tweak the PR.

I'll also re-check the old #266 and #270 issues now that I'm using 0.2.0-beta.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -7,6 +7,7 @@ import assert from './assert';
 
 import _assign from 'lodash/object/assign';
 import _isArray from 'lodash/lang/isArray';
+import _isFunction from 'lodash/lang/isFunction';
 
 function isModel(object) {
   return object instanceof Model;
@@ -124,7 +125,7 @@ export default class SerializerRegistry {
     }
 
     // Traverse this model's relationships
-    serializer.include
+    this._valueForInclude(serializer, request)
       .map(key => model[camelize(key)])
       .filter(Boolean)
       .forEach(relationship => {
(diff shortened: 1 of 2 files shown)


samselikoff:
Interesting. I typically like to come up with more general purpose APIs initially, then drop into more specific "sugar" for common cases once they show up with some use. An inline serializer (as discussed in the issue) would allow users to essentially accomplish a dynamic `includes` array, among other things. I'm wondering if we should land that API first, then add sugar on top of it via something like `includesForRequest`.

Curious, how did this need arise for you? What backend are you using - ActiveModelSerializers? JSON:API? Something else?

2468ben:
I agree, the inline serializer API would be a cool way to support a bunch of 'custom' cases.
I'm using Mirage's ActiveModelSerializer, and my use case is always something like this:
```
//mirage/models/topic.js
export default Model.extend({
  responses: hasMany('comment')
});
```
```
//mirage/serializers/topic.js
export default ActiveModelSerializer.extend({
  includesForRequest: function(request) {
    return request.queryParams.include_responses ? ['responses'] : [];
  }
});
````
I also have one case where I need to check request.url instead of request.queryParams, but in any case the inline serializer API you described would cover all the bases.

**Comment by samselikoff** [comment]
@2468ben How about we drop `includesForRequest ` and just let users define either a static array or a function for `include`? That way they can use your feature via

```js
  include(request) {
    // return an array
  }
});
```

and we don't have a new method name as part of the API. If you make that change + add a test I'm happy to merge.

2468ben:
Cool, did just like you said, so it'll act the same way that factory props do. This is the only "feature" I needed to get our whole backend API mocked in mirage, so thanks for supporting it.

Issue #506, opened 2016-01-18
AMS#normalize is not handled correctly


martndemus:
The `PUT`/`POST` shorthand methods don't handle normalizing the request payload very well.
The payload is first normalized by the `SerializerRegistry` :
https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializer-registry.js#L26-L29
```js
normalize(payload) {
  let model = payload[Object.keys(payload)[0]];
  return this._serializerFor(model.modelName).normalize(payload);
}
```

This correctly picks of the first key from the payload, but then trys to get `modelName` from the AMS payload, which if it is correct AMS, will never be there, so the AMS#normalize method will never be called.

samselikoff:
You're absolutely correct. I think the reason I missed this is, `model.modelName` will always return `undefined`, so `serializerFor` will default to returning the `ApplicationSerializer`. If that's defined (under `/serializers/application`), the normalization will generally work. But this is definitely broken.

samselikoff:
I'm curious, how did you come across this bug? Did you have an application serializer defined?

martndemus:
The `ApplicationSerializer` is a `JSONApiSerializer`, I need the AMS serializer for a few models on older endpoints.

samselikoff:
Perfect, makes total sense. Should be fixed in master!

PR #520, opened 2016-01-20
attempt at #506, passing modelName to registry.normalize for AMS compatibility

2468ben:
In #506, when AMS serializers used registry.normalize, they didn't have a model.modelName property. Now instead of discovering modelName from the payload props, we can pass it down from the shorthands.

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -78,10 +78,10 @@ export default class BaseShorthandRouteHandler {
     return body;
   }
 
-  _getAttrsForRequest(request) {
+  _getAttrsForRequest(request, modelName) {
     let id = this._getIdForRequest(request);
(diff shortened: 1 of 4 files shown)


samselikoff:
Thank you!

I tracked down this code path for a bit here locally and have a failing test (it was a bit tricky to figure out which level we should be testing at), but typically I won't merge PRs without associated tests. I'm just trying to move things along because there's a big backlog of issues I need to tackle.

Thanks for your help!

2468ben:
No problem, and I absolutely owe you tests for my remaining PRs (you're the reason I can write half of my acceptance tests!). I was adding PRs as I migrated a whole existing backend api to Mirage's ORM, and I expected to add coverage this weekend before some baby-troubles. I'll cover the rest asap.

samselikoff:
No worries at all, this PR was still saved me time! Thanks again.

PR #521, opened 2016-01-21
Camelize type of schema._collectionForType

dopin:
This fixes dasherlized collection finding error like `schema.find('item-option', 1)`
in route handlers.
I'm not sure if this is a bug or a feature. But I guess it would be nice that you can
write with dasherized `schema.find('foo-bar', id)` same as `server.createList('foo-bar', 10)`.

### DIFF
--- a/addon/orm/schema.js
+++ b/addon/orm/schema.js
@@ -112,7 +112,7 @@ export default function(db) {
     Private methods
   */
   this._collectionForType = function(type) {
-    var collection = pluralize(type);
+    var collection = pluralize(camelize(type));
     if (!this.db[collection]) {
       throw 'Mirage: You\'re trying to find model(s) of type ' + type + ' but this collection doesn\'t exist in the database.';
     }


samselikoff:
This is interesting... I'm very curious how you found about this api. `schema.find` is actually private. When models are registered they get a `modelClass` attached to schema. The idea was to take after Rails' API. So if you have a `models/item-option.js`, you'd access it in route handlers like

```js
schema.itemOption.find(1);
```

This works nicely with ES6 destructuring:

```js
this.get('/item-options', ({ itemOption }) => {
  return itemOption.all();
});
```

So you see `schema.find` was never intended to be called directly. Did you just happen to stumble across it in your console or something?

dopin:
>schema.find is actually private
:flushed: !! I'm sorry. I didn't know that.

>Did you just happen to stumble across it in your console or something?
I don't actually remember but yes I think I found it in console. Maybe I was confusing `schema.find`, `model.find(id)` and `store.find` of Ember DATA.

samselikoff:
Makes sense. I'll go ahead and close this for now, then.

PR #522, opened 2016-01-21
missing camelize

2468ben:
just a small one found when setting a hasMany with the ORM.

### DIFF
--- a/addon/orm/associations/has-many.js
+++ b/addon/orm/associations/has-many.js
@@ -121,7 +121,7 @@ class HasMany extends Association {
             .forEach(model => model.save());
 
           // Associate the new children to this model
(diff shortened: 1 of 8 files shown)


2468ben:
I know the test changes are a little overwhelming just to transition to using a camelized model name.

I thought I could make a separate PR later this week for the association tests, that would abstract the model calls to something like ```schema['childrenKey']``` instead of ```schema.homeAddresses``` so we can run all of those tests against a list of different schemas. Then you can check more unusual stuff like ```homeAddresses: hasMany('address')``` or other future options.

samselikoff:
This is really good, I've been wanting to upgrade all the tests to use multi-word model names for this very reason. And, like you're saying, we need to test naming the key something other than the model type.

Great work, thank you!

PR #523, opened 2016-01-21
filtering for existence

2468ben:
Just this one traversal function wasn't checking for null relatedModels.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -126,6 +126,7 @@ export default class SerializerRegistry {
     // Traverse this model's relationships
     serializer.include
       .map(key => model[camelize(key)])
+      .filter(Boolean)
       .forEach(relationship => {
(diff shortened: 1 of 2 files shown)


addon/serializer-registry.js
```diff
@@ -126,6 +126,7 @@ export default class SerializerRegistry {
     // Traverse this model's relationships
     serializer.include
       .map(key => model[camelize(key)])
+      .filter(i => i)
```
bekzod: I think instead of creating function you just do `.filter(Boolean)`

2468ben: Much better, thanks! I just updated to use that.

samselikoff:
Looks great! Could you squash commits and then I'll merge these in.

2468ben:
Sure, all set... I'll do it for my other PRs from now on.

samselikoff:
Awesome, thank you sir, and thanks for all your help lately! Great stuff.

PR #526, opened 2016-01-26
Add iteration argument to faker.list.cycle

alecho:
`faker.list.cycle` returns a function that expects an iteration to be passed: https://github.com/samselikoff/ember-cli-mirage/blob/dfd7f620277e3f9bddf6d817a19fd5802e5aaf33/addon/faker.js#L13


samselikoff:
Yes - thank you very much!

PR #536, opened 2016-01-29
Rename `type` to `camelizedModelName`

### LINKED ISSUE #531: Rename `type` to `camelizedModelName`
Often throughout the code, especially when dealing with `schema` and the ORM, we need to use a `camelized` version of a model's name. `model.modelName` is dasherized - something like `home-address` - and when we need to camelize it we'll often call it `type`:

```js
let type = camelize(modelName)`;
```

Let's rename all these local variables to `camelizedModelName`, to be more explicit.

```js
let camelizedModelName = camelize(modelName);
```

No need to introduce a new concept when we can be clearer in exchange for a few more characters.

alecho:
* closes #531

### DIFF
--- a/addon/orm/schema.js
+++ b/addon/orm/schema.js
@@ -23,14 +23,14 @@ export default class Schema {
   }
 
   registerModel(type, ModelClass) {
-    type = camelize(type);
+    let camelizedModelName = camelize(type);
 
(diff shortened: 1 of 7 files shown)


samselikoff:
Lovely! Thank you very much!

PR #541, opened 2016-01-31
Update to ember-cli 2.3.0-beta.1

cibernox:
Noticeably faster rebuild times :)


samselikoff:
Siiick. Thanks dude!

I'm sorry but I have a strong preference for `unused: "vars"`, think we could leave it in? (Am I going against the community here? I think it's extremely valuable to be able to hint to future me/other devs about params that are likely to be needed).

cibernox:
The most widely used config is `unused: true`. I guess that is a marginally safer.
When I want to document arguments that I don't use I just leave them commented:
```js
function(schema /*, request, options*/) {
  // stuff
}
```

That said, I don't mind change it back to vars, I just changed it because it's what `ember init` defaults to.

samselikoff:
Yah, if you could. Documenting args with comments is ok but I prefer to just have them there.

Once you do that I'm happy to merge in. Thanks Miguel!

cibernox:
Done

samselikoff:
Thank you sir!

PR #547, opened 2016-02-03
Add missing arrows

cafedomancer:
Fixed some expressions apparently missing arrows.
I wish we could check the syntax of codes embedded in documents...


samselikoff:
Thank you v much!

PR #549, opened 2016-02-04
Fixes #548: Use json api id when available.

### LINKED ISSUE #548: Mirage creating new id rather than using client-generated id
I'm using generateIdForRecord in a DS.Adapter to pass a client-side generated id when creating.  Mirage is generating it's own id value when it stores the record causing Ember to return the error "An adapter cannot assign a new id to a record that already has an id." when the payload returned uses a different id than the id already specified on the DS model.  I've verified that the "id" is passed within the payload as expected.

lkhaas:
This allows using an id value that is passed as part of the submitted json api resource when a value is not specified in the request params.  It also removes integer parsing of ids per #462.  I've tried to add a test as well although I'm having a hard time running the tests locally to verify :(

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -2,8 +2,6 @@ import _isArray from 'lodash/lang/isArray';
 import assert from 'ember-cli-mirage/assert';
 import { camelize, singularize, dasherize } from 'ember-cli-mirage/utils/inflector';
 
-const allDigitsRegex = /^\d+$/;
-
 export default class BaseShorthandRouteHandler {
 
   constructor(schema, serializerOrRegistry, shorthand, path, options={}) {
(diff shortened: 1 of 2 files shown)


samselikoff:
The changes look great - thanks for opening this PR!

Let's split up the test into two separate ones, something like

```js
test('it can read the id from the url', function(assert) {
  ...
});

test('it can read the id from the request body', function(assert) {
  ...
});
```

Also, we can test `_getIdForRequest` using free variables, without the need for `this.request = ` and `this.jsonApiDoc = `. Something like

```js
// first test
let request = {params: {id: "test-id"}};
assert.equal(this.handler._getIdForRequest(request), 'test-id');

// second test
let request = {...};
let jsonApiDoc = {...};
assert.equal(this.handler._getIdForRequest(request, jsonApiDoc), 'test-id');
```

With these changes I'm happy to merge in this PR! Lastly - what are you seeing when trying to run tests locally?

lkhaas:
Thanks for the feedback.  I've adjusted the PR with the test changes.  The test failures appear to be due to using phantomjs 1.9.  Once I upgraded everything worked fine :)

samselikoff:
Awesome - thanks for this!

PR #543, opened 2016-02-01
Fix addon integration

cibernox:
Closes #524 

It depends on / contains #541 (you can merge only this one and close the other if you want)

Honestly, I have no idea how to write a test for this.

Issue #524, opened 2016-01-22
Mirage is not working with an addon


BryanHunt:
I'm trying to use Mirage 0.2.0-beta.3 inside an add-on with the dummy app.  The first problem I ran into is:

```
Attempting to watch missing directory: /Users/bhunt/Documents/Code/OpenSource/ember-d3/mirage
Error: Attempting to watch missing directory: /Users/bhunt/Documents/Code/OpenSource/ember-d3/mirage
    at EventEmitter.Watcher_addWatchDir [as addWatchDir] (/Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/broccoli-sane-watcher/index.js:90:11)
    at /Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/broccoli/lib/builder.js:94:35
    at lib$rsvp$$internal$$tryCatch (/Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:493:16)
    at lib$rsvp$$internal$$invokeCallback (/Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:505:17)
    at /Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1001:13
    at lib$rsvp$asap$$flush (/Users/bhunt/Documents/Code/OpenSource/ember-d3/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1198:9)
    at doNTCallback0 (node.js:419:9)
    at process._tickCallback (node.js:348:13)
```

If you work around this problem my creating an empty mirage directory at the project root, you then get:

`Uncaught Error: Could not find module dummy/mirage/config imported from dummy/initializers/ember-cli-mirage`

samselikoff:
Thanks for reporting, we'll get this situated before the 0.2.0 release.

cibernox:
I also found this issue today after update from 0.1.X.

The stack trace is very very similar:

```
Attempting to watch missing directory: /Users/miguelcamba/code/own/ember-power-select/mirage
Error: Attempting to watch missing directory: /Users/miguelcamba/code/own/ember-power-select/mirage
[...]

PR #554, opened 2016-02-10
Fix the fix on addon integration.

abuiles:
The check in #543 is correct, but if a project has "in-repo-addons" then
it will include `ember-addon` in `package.json` without it being an
addon.

The following shows how in-repo-addons are discovered https://github.com/ember-cli/ember-cli/blob/282641ba662292afb40cb88cdf31557a3e4cc6b7/lib/models/addon-discovery.js#L160

Without this #524 still happens if you have in-repo-addons..

cc @cibernox

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ module.exports = {
     this.addonBuildConfig = this.app.options['ember-cli-mirage'] || {};
     if (this.addonBuildConfig['directory']) {
       this.mirageDirectory = this.addonBuildConfig['directory'];
-    } else if (app.project.pkg['ember-addon']) {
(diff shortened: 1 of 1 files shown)


cibernox:
Good catch, I never thought about in-repo addons.

topaxi:
I just stumbled upon this. A new beta release with this would be nice :+1:

samselikoff:
Thanks yall! Suuper helpful to have all of your eyes on this, as I haven't had need for Mirage in an addon yet.

I'll cut a new release soon.

topaxi:
Awesome thanks!

PR #555, opened 2016-02-10
Allow Mirage to be disabled for non-production environments

dnachev:
I only need Mirage for acceptance test, but it is included in 'development' environment as well. This PR changes the logic for determining when to include Mirage. It preserves the previous behavior where Mirage is included by default in all non-production environments and it can be enabled for 'production' environment. It adds ability to explicitly disable Mirage in other environments. It probably doesn't make any sense for disabling in 'testing', but it make sense to disable it for 'development' if the API server is fully implemented and there is no need for mocking.

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,12 @@ module.exports = {
 
   _shouldIncludeFiles: function() {
     var enabledInProd = this.app.env === 'production' && this.addonConfig.enabled;
-
-    return enabledInProd || (this.app.env !== 'production');
+    var explicitExcludeFiles = this.addonConfig.excludeFilesFromBuild;
+    if (enabledInProd && explicitExcludeFiles) {
+      throw new Error('Mirage was explicitly enabled in production, but its files were excluded ' +
+                      'from the build. Please, use only ENV[\'ember-cli-mirage\'].enabled in ' +
+                      'production environment.');
+    }
+    return enabledInProd || (this.app.env !== 'production' && explicitExcludeFiles !== true);
   }
 };


samselikoff:
There's already an enabled option available in the config:
http://www.ember-cli-mirage.com/docs/v0.1.x/server-configuration/#enabled.
Does that address your need?

dnachev:
Even when I added `enabled: false` for `development` environment, I still got Pretender library in the output (which I'm actually trying to exclude). I've looked at `shouldIncludeFiles` function and it seems it is applied in limited cases and not in the one I need. If `enabled: false` is considered whether the Pretender server should start or stop, this won't be enough for my case.

samselikoff:
Gotcha. The reason this behavior exists is because users may want to

```
ENV['ember-cli-mirage'].enabled = false
```

in development, but still be able to visit `localhost:4200/tests` to run their tests. In this case, the Ember app at `/tests` has a build-time environment of `development`, but a run-time environment of `test`. 

As long as people are using `/tests` (which I believe they still are), we shouldn't break this behavior. How about a new option, `excludeFilesFromBuild`, which when `true` will give you the behavior you want?

Out of curiosity, why are you trying to exclude Mirage/Pretender files from the development build in the first place?

dnachev:
Oh, I see. This is an interesting case :)

I'm rendering the Ember application server-side and Pretender doesn't play well with node environment. Therefore, I need Mirage/Pretender to be excluded from the actual application environment and be available only when running the tests. I'm fine with any other alternative you suggest for solving this case.

samselikoff:
I see. Totally valid. Let's change the option to `excludeFilesFromBuild` and add a check for that variable being `true` in the `shouldIncludeFiles` piece.

dnachev:
I didn't have the chance to check how it works yet. I will report again once I've tried it out.

samselikoff:
It looks great - the API is correct so I'll go ahead and merge, and if we need to change anything we can. 

Thank you!

PR #557, opened 2016-02-10
Updating ember-cli and related deps

ef4:
Starting with ember-cli 2.3.0-beta.2 these two dependencies move from bower to npm.


samselikoff:
Awesome, thanks Edward!

Issue #560, opened 2016-02-10
[Doc] seeding database when using Models


jelhan:
I'm running 0.2.0-beta.3. I've defined models since I need support for relationships. Seems to work fine. But I haven't any clue how to seed the database. If I'm running `server.create('my-model');` I'm getting an error that I must define a factory. But can't find any information in docs how models and factories play together.

jelhan:
Should have ready [quickstart guide](http://www.ember-cli-mirage.com/docs/v0.2.0-beta.3/quickstart/)... Sorry for the noise.

samselikoff:
Glad you found it. We will be changing this so if you don't have a factory
defined, server.create will fall back to the model.

jelhan:
I was to fast by closing. I thought it's the same as described for [schemas](http://www.ember-cli-mirage.com/docs/v0.2.0-beta.3/quickstart/#associations-and-serializers) but object returned by `server.create('factory-name');` hasn't a create function for relationships.

I thought it's like this:
```
// mirage/models/author.js
import { Model, hasMany } from 'ember-cli-mirage';

export default Model.extend({
  blogPosts: hasMany()
});

// mirage/models/blog-post.js
import { Model } from 'ember-cli-mirage';

export default Model;

// mirage/factory/author.js
port { Factory } from 'ember-cli-mirage';

export default Factory.extend({
  name(i) {
    return `author ${i}`;
  }
});

// mirage/factory/blog-post.js
port { Factory } from 'ember-cli-mirage';

export default Factory.extend({
  title(i) {
    return `title ${i}`;
  }
});

// acceptance test
const author = server.create('author');
autor.createBlogPost();
```

samselikoff:
Yes this is also as an open issue, server.create doesn't currently return
model instances, it only returns attrs. This will change soon.

However, you'll generally want to user server.create, since
author.createBlogPost won't leverage any factory attributes you've defined.

jelhan:
Thanks for your fast response. I still didn't understand how to associate the relationship when using `server.create()`. Could you please give an example?

jelhan:
Is it like this:
```
const blogPost = server.create('blog-post');
const author = server.create('author', {
  blogPosts: [
    blogPost.id
  ]
});
```

And could it be that setting a belongsTo is currently not supported?

```
// mirage/models/blog-post.js
import { Model, belongsTo } from 'ember-cli-mirage';

export default Model.extend({
  author: belongsTo()
});

[...]

PR #568, opened 2016-02-11
Fix: newModel.isNew is not a functioon when associating a belongsTo by plain object

jelhan:
Fix: Your handler for the url ... threw an error: newModel.isNew is not a function

Error was thrown while using a plain object returned by `server.create()`:
```
const author = server.create('author');
server.create('blog-post', {author: author})
```

Also adds support for associating by passing id of model:
```
const author = server.create('author');
server.create('blog-post', {author: author.id});
```

Bug was mentioned in #560.

### DIFF
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -2,6 +2,7 @@ import Association from './association';
 import _assign from 'lodash/object/assign';
 import { capitalize, camelize } from 'ember-cli-mirage/utils/inflector';
 import assert from 'ember-cli-mirage/assert';
+import Ember from 'ember';
 
 class BelongsTo extends Association {
 
@@ -44,7 +45,7 @@ class BelongsTo extends Association {
       set(id) {
         assert(
           !id || schema[camelize(association.modelName)].find(id),
-          `Couldn\'t find ${association.modelName} with id = ' + id`
+          `Couldn't find ${association.modelName} with id = ${id}`
         );
 
         this.attrs[foreignKey] = id;
@@ -75,13 +76,20 @@ class BelongsTo extends Association {
           - sets the associated parent (via model)
       */
       set(newModel) {
-        if (newModel && newModel.isNew()) {
+        if (!Ember.isEmpty(newModel) && typeof newModel === 'object' && typeof newModel.isNew === 'function' && newModel.isNew()) {
+          // newModel is a model
           this[foreignKey] = null;
           association._tempParent = newModel;
-        } else if (newModel) {
+        } else if (!Ember.isEmpty(newModel) && typeof newModel === 'object' && !Ember.isEmpty(newModel.id)) {
+          // newModel is a simple object with id property
           association._tempParent = null;
           this[foreignKey] = newModel.id;
+        } else if (!Ember.isEmpty(newModel)) {
+          // newModel is an id
+          association._tempParent = null;
+          this[foreignKey] = newModel;
         } else {
+          // newModel is not present
           association._tempParent = null;
           this[foreignKey] = null;
         }


samselikoff:
So, we want the behavior to match Rails. If you create a new post with `author: author`, the `author` var should be a model instance. You can also do `authorId: author.id`.

We're going to update `server.create` to return the model instance created instead of just plain JSON, so that would solve your problem here. For now you should be able to use `authorId`. 

Does that help?

jelhan:
I'm not a rails guy so didn't expected this api. I expected an ember-data like api.

Also the code gave me another feeling. If `newModel` must be a model instance or null/undefined [these `elseif` condition](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/associations/belongs-to.js#L81-L83) seems to be useless. Should it be removed and an `Ember.assert` be added to throw a more helpfull error?

Also please notice that passing object hash seems to work fine just with my little changes.

samselikoff:
That `elseif` is hit if the model exists but is not new.

I think it'd be okay to add the behavior you're talking about. We'll need a test [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/schema/belongs-to/set-association-test.js), something like

```
test(`a ${state} can update its association using a parent id`, function(assert) {
```

You can use the `var savedUser = this.helper.savedParent();` line to get a saved parent.

Also, we should have a test that throws an error if you try to set an id that references a parent that doesn't exist.

Leooo:
@samselikoff what we need is a backend serializer there, maybe by default `_getAttrsForRequest` should look for `relationshipName` attributes which should be transformed into `relationshipNameId(s)` in the request?

samselikoff:
I think now that `server.create` returns models I think this can be closed. Doing this now works:

```js
let author = server.create('author');
server.create('blog-post', { author })
```

Leooo:
@samselikoff not it doesn't, on a route POST / PUT request where relationships are as ids in the requests this fails

samselikoff:
@Leooo but isn't this issue just about `server.create`? Which is just about seeding the Mirage db (doesn't have to do with routes/POST/PUT etc)

Leooo:
Best is to create a test for it, but on my side I think I saw multiple such errors when using POST routes:

`````
  this.post('/', function ({serviceOrders}) {
    let serviceOrderJSON = this.normalizedRequestAttrs();
    let serviceOrder = serviceOrders.create(serviceOrderJSON);
    serviceOrder.hasNo('appointmentSlot');
    let serialized = this.serialize(serviceOrder);
    return {
      data: serialized

    };
  });
`````

samselikoff:
We do have a [normalize](http://www.ember-cli-mirage.com/docs/v0.2.x/serializers/#normalizejson) method on the Serializer, so maybe it's a bug in `normalizedRequestAttrs()`?

If it's a bug we should discuss in a separate issue

Leooo:
@samselikoff you can close as at this stage this seems to be a bug very specific to our applicationSerializer

Leooo:
@samselikoff here is what I meant by 'adding a normalize method': in our case we send to a PUT request this payload:
`````
"meterReads":[{"reading":"3456","register":"1"}
`````

Where `register` is actually the id of the model `register`, so should be converted to`registerId`

I'm able to use the `normalizedRequestAttrs()` without the PR proposed in this thread only by overriding the `normalize` method in the application serializer with the very ugly one below:

`````
  normalize(payload) {
    let type = Object.keys(payload)[0];
    let attrs = payload[type];

    let jsonApiPayload = {
      data: {
        type: type.pluralize(),
        attributes: {}
      }
    };
    if (attrs.id) {
      jsonApiPayload.data.id = attrs.id;
    }
    if (attrs.length) {// is an array
      attrs = attrs[0];
    }
    Object.keys(attrs).forEach(key => {
      if (key !== 'id') {
        let normalizedKey = key;
        // server sends payload with keys as "register" instead of "registerId"
        if (Object.keys(server.schema).contains(key.pluralize())) {
          normalizedKey = `${key}Id`;
        }
        jsonApiPayload.data.attributes[normalizedKey.dasherize()] = attrs[key];
      }
    });

    return jsonApiPayload;
  }
`````

PR #570, opened 2016-02-11
Add default passthroughs; fix #552

### LINKED ISSUE #552: [PROPOSED] Add default passthrough to prevent Chrome for iOS "unverified URL" errors
When configured for production, Mirage (via Pretender) intercepts Chrome for iOS' default URL verification, as noted [in this blog post](
http://blog.isleofcode.com/configuring-ember-cli-mirage-and-pretender-for-production/).

In discussion via Ember-Slack, @samselikoff agreed this solution would be nice to add as a default-but-configurable option in Mirage.

#### ToDo
- [x] add `defaultPassthroughs` array as config option to [config/environment.js#L22](https://github.com/samselikoff/ember-cli-mirage/blob/master/config/environment.js#L22) (defaults to `true`)
- [x] add `this.configureDefaultPassthroughs` function to [server.js#L97-L132](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/server.js#L97-L132)
    - [x] add file-level `const defaultPassthroughUrls = [...]`
    - [x] for each `defaultPassthroughUrl`, `server.passthrough(foo)`
- [x] add tests to validate default passthroughs are configured per boolean state

#### Questions
- Should we restrict this config setting to production environments only?

> No; users can refine as necessary

- Is it ok to import Ember's `isBlank` fn, or should we default to a more manual `typeof` / `instanceof` approach?

> Use `lodash/lang/isEmpty` per @samselikoff's suggestion below

anulman:
This PR adds a `defaultPassthroughs` array to `server.js`, and automatically configures passthroughs for all URLs in the array when `ENV['ember-cli-mirage'].useDefaultPassthroughs` is true (defaults to true).

For more, see #552 & [the related blog post](http://blog.isleofcode.com/configuring-ember-cli-mirage-and-pretender-for-production/)

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -52,6 +52,11 @@ const defaultRouteOptions = {
   timing: undefined
 };
 
+const defaultPassthroughs = [
+  'http://localhost:0/chromecheckurl'
+];
+export { defaultPassthroughs };
+
 function isOption(option) {
   if (!option || typeof option !== 'object') { return false; }
   const allOptions = Object.keys(defaultRouteOptions);
@@ -128,6 +133,10 @@ export default class Server {
       options.scenarios.default(this);
     } else {
       this.loadFixtures();
+    }
+
+    if (options.useDefaultPassthroughs) {
+      this._configureDefaultPassthroughs();
     }
   }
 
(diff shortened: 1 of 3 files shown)


samselikoff:
Looks perfect! Could you squash + I'll merge?

anulman:
#squashed

samselikoff:
woot, thanks!

anulman:
:tada:

PR #582, opened 2016-02-15
Make server.create and server.createList return Models

### LINKED ISSUE #427: server.create should return model instance, if defined
If models are being used, `server.create('post')` should return an instance of the `Post` model that was created (instead of the attrs hash).

ef4:
Closes #427.


samselikoff:
:heart::heart:

thanks man!

samselikoff:
We should probably make `build` do this too

PR #586, opened 2016-02-18
Fix check for belongsTo relationship and use null data.

abuiles:
typeof can produce inconsistent behavior here since typeof `[]` and
`null` is 'object'.

We still want to pass `null` as the id since that is valid when for
example removing the link to the other object.


samselikoff:
:+1: thank you sir

PR #590, opened 2016-02-19
fix js object typos

vikram7:
This PR just fixes up some of JS Object syntax typos.

For example, `{key: value}` should be `{ key: value }`.


samselikoff:
Thanks!

PR #593, opened 2016-02-20
Update serializers documentation for "include" rather than "relationships"

lkhaas:
The example is updated but the surrounding text still says "relationships" rather than "include"


samselikoff:
Thank you!

PR #603, opened 2016-02-24
Inverse Foreign Keys

ef4:
This adds an optional `inverseOf` option on `hasMany` associations so that you can correctly map relationships in which the child model has a non-default association key.

I also expanded test scenario coverage of all the tests that use the HasManyHelper so that they also include every permutation of default and non-default keys for the hasMany association and its inverse.

Note that I fixed what I would consider to be a bug, which is that the `newFoo` and `createFoo` methods were being named based on the destination model name and not the local key. That makes multiple associations to the same model type impossible to use unambiguously.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -8,11 +8,11 @@ import JSONAPISerializer from './serializers/json-api-serializer';
 import HasMany from './orm/associations/has-many';
 import BelongsTo from './orm/associations/belongs-to';
 
-function hasMany(modelName) {
-  return new HasMany(modelName);
+function hasMany(...args) {
+  return new HasMany(...args);
 }
-function belongsTo(modelName) {
-  return new BelongsTo(modelName);
+function belongsTo(...args) {
+  return new BelongsTo(...args);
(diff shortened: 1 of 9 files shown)


ef4:
Example usage:

```js
let User = Model.extend({
  photos: hasMany({ inverseOf: 'owner' }),
  appearsIn: hasMany('photo', { inverseOf: 'subject' })
});

let Photo = Model.extend({
  owner: belongsTo('user'),
  subject: belongsTo('user')
});
```

samselikoff:
Amazing PR. Seriously, thanks so much for putting this together. And I know many others will be happy to see this land.

The tests are getting pretty abstract - they already were, but now that the `key` is dynamic, even more so. I prefer duplication to abstraction when it comes to writing tests, but with with the number of permutations we have here it does seem untenable. Maybe one day we could discuss how to keep them more explicit.

Everything looks great - thanks again!

cs3b:
great job @ef4 :-)

PR #610, opened 2016-02-29
Allow for nesting ember-cli-mirage addon in another addon

donovan-graham:
This fix allow nesting ember-cli-mirage as an addon dependency of another addon.  

As described on the ember-cli docs here:
http://ember-cli.com/extending/#addons-depending-on-other-addons

The implementation has been discussed at length here:
https://github.com/ember-cli/ember-cli/issues/3718

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,13 @@ module.exports = {
   name: 'ember-cli-mirage',
 
   included: function included(app) {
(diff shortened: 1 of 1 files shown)


samselikoff:
Ok great, thanks a lot!

PR #572, opened 2016-02-12
Accept JSON:API requests which contain no model attributes.

nbibler:
See #534.

PR #614, opened 2016-03-01
refactoring getAttrsForRequest to allow relationship updates

zachfedor:
Previously, this method assumed that the request body contained an `attributes` property which is fine when you are actually updating properties of a model (e.g. changing the name of a user or the content of a post).

My changes allow updating model relationships without modifying the existing models (e.g. changing a post's author to another user).

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -63,16 +63,19 @@ export default class BaseShorthandRouteHandler {
   _getAttrsForRequest(request, modelName) {
     let json = this._getJsonApiDocForRequest(request, modelName);
     let id = this._getIdForRequest(request, json);
+    let attrs = {};
 
     assert(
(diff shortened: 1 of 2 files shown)


samselikoff:
Looks great! Could we get a test for this, one that fails on without your changes?

zachfedor:
Good call. I copied the previous unit test for `_getAttrsForRequest`, removed the attributes, and simplified it. It passes with my change and fails without it.

Hope this helps!

samselikoff:
Yes, perfect! Could you squash commits? Then this is ready to go

zachfedor:
Commits are squashed. I believe my PR will also do away with the need for [this one](https://github.com/samselikoff/ember-cli-mirage/pull/572)

samselikoff:
Nice! And yep, I agree. Thank you!

PR #617, opened 2016-03-02
Include intermediate dot-path relationships

RSSchermer:
Recently started using ember-cli-mirage. Was very happy when I found out it already supported the `include` query param for the JSON::API adapter!

However, per the JSON::API spec, when a dot-path relationship is included, all intermediate relationships should be included as well (see the second text box of the [Inclusion of Related Resources](http://jsonapi.org/format/#fetching-includes) section). This PR should make mirage comply with that.


samselikoff:
You're absolutely right - great catch! Thank you for this!

Issue #671, opened 2016-03-31
Latest Travis build did not run all tests


lependu:
I have no experience with Travis, so maybe the problem is in my side, but I am figthing with this issue for a while as mentioned [here](https://github.com/samselikoff/ember-cli-mirage/issues/657#issuecomment-200601524), and now it seems to me Travis does not run all tests.

Update: Maybe this issue is the reason: DockYard/ember-suave#94, kellyselden/broccoli-jscs#53.

Reproducing:
`git clone git@github.com:samselikoff/ember-cli-mirage.git`
`cd ember-cli-mirage`
`npm cache clean`
`bower cache clean`
`npm i`
`bower i`
`npm test`
>ember-cli-mirage@0.2.0-beta.8 test /home/ror/ember/ember-cli-mirage
>ember try:testall && mocha tests/**/*-test-node.js

>Cannot find module 'lodash/lang/isFunction'
>Error: Cannot find module 'lodash/lang/isFunction'
>    at Function.Module._resolveFilename (module.js:339:15)
>    at Function.Module._load (module.js:290:25)
>    at Module.require (module.js:367:17)
>    at require (internal/module.js:16:19)
>    at Object.<anonymous> (/home/ror/ember/ember-cli-mirage/node_modules/babel-core/lib/api/node.js:21:29)
>    at Module._compile (module.js:413:34)
>    at Object.Module._extensions..js (module.js:422:10)
>    at Module.load (module.js:357:32)
>    at Function.Module._load (module.js:314:12)
>    at Module.require (module.js:367:17)
>npm ERR! Test failed.  See above for more details.

`rm -rf node_modules/babel-core`
`npm i`
`npm test`

test runs with the result:
>1..1073
>tests 1073
>pass  1070
>skip  0
>fail  3

Meanwhile the latest travis job [1052.1](https://travis-ci.org/samselikoff/ember-cli-mirage/jobs/119717233):
>1..1009
>tests 1009
>pass  1009
>skip  0
>fail  0

samselikoff:
Try rebasing off master

lependu:
Done.

Issue #657, opened 2016-03-22
add ember-suave and fix any/all javascript linting errors


bantic:
Mentioned by @samselikoff as a useful task that's good for new contributors that could be done to help improve the overall code cleanliness of mirage.

lependu:
Hi, There are 13 failing test after installed ember-suave. I am working on it.

lependu:
For the record, rebuilding the project from scratch brought back the failing tests. But I still don't have any idea why it passes in my local environment after a while.

Still some work to do, but there is some talking points:

- [ ]  Filtered out the the disallowEmptyBlocks rule due the fixtures/config/*.js files. As I see the other option is commenting out the empty blocks in those files.

- [ ] Filtered out requireSpread rule because I couldn't figure out how to pass 

> //tests/helpers/module-for-acceptance.js
> options.beforeEach.apply(this, arguments);

samselikoff:
Awesome, thanks for the PR!

Is it possible to ignore certain rules within the files instead of omitting all linting from these problem files completely?

lependu:
As far I know `// jscs:disable` disables entry file, `// jscs:disable ruleName` disables specific rule  in file.

amyrlam:
Hi @lependu! Not working on it, you can implement...had never seen that before. 

Since we fixed all the JSCS errors I guess it would report nothing rn, but may be useful for future error debugging?

samselikoff:
I think we can filter out the errors for the known issues (e.g. snake_case) but leave the rest

lependu:
Ok, then I give it a try.

lependu:
It seems adding filters through `.jscsrc` broke tests, so I removed them completely. Please see previous comment. The same is true of the 'errorFilter' implementation.
So, converted them into inline filters and added `/.jscsfilters` file with the list of them in case someone wants to review it.

Because @samselikoff added global `requireSpread` and `requireParenthesesAroundArrowParam` filters in #670 I applied those filters in the current fixes. My question is what we shall do with the fixes commited in #661 regarding those rules?

acorncom:
@samselikoff Looks like the latest PR was merged, what additional items are you after here?

PR #673, opened 2016-04-01
Adds ember-suave inline filters + Fixes more linting errors

lependu:
https://github.com/samselikoff/ember-cli-mirage/issues/657#issuecomment-204349715
I think this solves https://github.com/samselikoff/ember-cli-mirage/issues/671#issue-144915155 as well.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,26 +1,3 @@
 {
-  "preset": "ember-suave",
-  "requireSpread": null,
-  "requireParenthesesAroundArrowParam": false,
-  "excludeFiles": [
-    "helpers/module-for-acceptance.js",
-    "fixtures/config/**",
-    "integration/schema/create-test.js",
-    "integration/schema/delete-test.js",
-    "integration/schema/read-test.js",
-    "integration/schema/reinitialize-associations-test.js",
-    "integration/schema/update-test.js",
-    "integration/schema/belongs-to/instantiating-test.js",
-    "integration/schema/has-many/instantiating-test.js",
-    "integration/serializers/active-model-serializer-test.js",
-    "integration/serializers/base/full-request-test.js",
-    "integration/serializers/json-api-serializer/key-formatting-test.js",
-    "integration/serializers/json-api-serializer/associations/included-test.js",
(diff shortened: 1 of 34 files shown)


samselikoff:
Awesome, thanks!

What's the `.jscsfilters` file for, global filters? Could we keep it all file-specific? Feels like that's too easy to ignore

samselikoff:
I just saw your comment in the other thread. Is it possible to just use comments at the top of files when we need to override the default settings?

lependu:
the `.jscsfilter` is just a plain text file listing all the filters. I added it because thought  it is hard to track down all the filters if someone wants to review them. But of course I can remove it.

Also updated filters regarding to changes in https://github.com/samselikoff/ember-cli-mirage/commit/13405df8770e496529e19985e9c6c46aed4c82ba

Shall I rebase into 1 commit, or keep commits separeted?

lependu:
One more concern: my npm test results
```
1..1094
tests 1094
pass 1094
skip 0
fail 0
```
while Travis [build #1057](https://travis-ci.org/samselikoff/ember-cli-mirage/builds/120266012) results
```
1..1030
tests 1030
pass  1030
skip  0
fail  0
```
So it seems in Travis still does not run all the tests.

samselikoff:
Sorry if I wasn't clear. I want `requireParenthesesAroundArrowParam` to be `false` globally for the project, because I don't like that rule. But if we have individual files that need to omit a certain rule for some reason, I'd rather do it in the file than in a global file.

Are you saying you cannot have a global change and also apply file-specific overrides? That would surprise me.

lependu:
I am saying you can define global rule in the `.jscrc`, but it will break your tests because https://github.com/DockYard/ember-suave/issues/94. That is why I removed all global rules. And wee need to apply them on file basis as inline filter.

samselikoff:
Ok - so are you happy with this PR? I've been heads down on some other issues so I don't have time to trace through all the references here.

lependu:
Yes, I think this is the best we can do in light of the `ember-suave` issue.

samselikoff:
:+1: thank you

PR #674, opened 2016-04-01
Corrects the meaning for the definition of `isNew()`

greyhwndz:
The `isNew()` instance method for the `mirage-model` is supposed to return true only when the model hasn't been persisted yet to the mirage db.
Added the word `not` :)


samselikoff:
thanks!

Issue #675, opened 2016-04-01
JSON-API serializer include dot-paths leave out intermediary models (but works with queryParams)


nruth:
I think Mirage (master) is leaving out intermediary models included with a dot path, but only if they're specified by the serializer instead of query params. The tests only cover the query-params case (which looks like it works) so I've added some below.

If someone could check if the test cases I've linked to below look OK that'd be great.

### JSON-API include multi-part paths and compound documents
According to http://jsonapi.org/format/#fetching-includes

>Note: Because compound documents require full linkage (except when relationship linkage is excluded by sparse fieldsets), intermediate resources in a multi-part path must be returned along with the leaf nodes. For example, a response to a request for comments.author should include comments as well as the author of each of those comments.

### Mirage behaviour (and integration test cases)

In the mirage tests, this seems to work with query-params but not serializer specified includes.

i.e. for query params see https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/serializers/json-api-serializer/associations/included-test.js#L298

However for serializer set it doesn't seem to work. This was missing in the original test so I've forked and added 3 cases to the integration test.

These two show the current behaviour:
https://github.com/nruth/ember-cli-mirage/blob/4a123614fd2966151ba3acb762538431f206b2e1/tests/integration/serializers/json-api-serializer/associations/included-test.js#L436
and
https://github.com/nruth/ember-cli-mirage/blob/4a123614fd2966151ba3acb762538431f206b2e1/tests/integration/serializers/json-api-serializer/associations/included-test.js#L485

And this shows what I think should happen:
https://github.com/nruth/ember-cli-mirage/blob/4a123614fd2966151ba3acb762538431f206b2e1/tests/integration/serializers/json-api-serializer/associations/included-test.js#L622

I'm running the tests with 
`ember test --module 'Integration | Serializers | JSON API Serializer | Associations | Included'` after npm install && bower install (and having run npm test a couple of times previously).

samselikoff:
Thanks for this! Easiest way for me to review is just to just PR your branch against the project. The tests will run & we'll be able to see them fail. Would you mind doing that?

nruth:
Yep ok. Closing this so it doesn't duplicate the PR. I've written less in the PR but it links back here anyway.

PR #676, opened 2016-04-01
tests showing json-api serializer dot-path includes miss intermediate models

nruth:
This adds 3 integration test cases for the json-api serializer's include dot-paths. 2 are false-positives that show the current behaviour. One is a failing case that shows what I think should happen. More info in issue https://github.com/samselikoff/ember-cli-mirage/issues/675

I think the things to do are:
- [ ] verify if it's a bug
- [ ] fix the bug
- [ ] delete the two false-positive test cases

### DIFF
--- a/tests/integration/serializers/json-api-serializer/associations/included-test.js
+++ b/tests/integration/serializers/json-api-serializer/associations/included-test.js
@@ -432,3 +432,326 @@ test(`dot-paths in include query params include query param`, function(assert) {
     ]
   });
 });
+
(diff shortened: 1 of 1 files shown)


nruth:
https://travis-ci.org/samselikoff/ember-cli-mirage/jobs/120205051#L1129 is the failing test. The error isn't very helpful so I added the other 2 test cases to show what it does instead. It might give more information using assertEqual, but I just went with using the same assert that the other test cases had used.

samselikoff:
Closing due to inactivity

PR #678, opened 2016-04-05
Index Route Passthroughs

noslouch:
Here's a failing test for passthroughs catching the index route.

I tried removing the leading `/`, but it didn't work. 

Maybe including the naked `/` in the list of paths when the `path.length === 0`?

### DIFF
--- a/tests/integration/passthrough-test.js
+++ b/tests/integration/passthrough-test.js
@@ -176,8 +176,8 @@ test('passthrough without args allows all paths on the current domain to passthr
   $.ajax({
     method: "GET",
     url: "/addresses",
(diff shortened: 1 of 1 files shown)


samselikoff:
Did you say you were able to do something to get this working? I actually can't find any configuration that tells Pretender to passthrough the root path of a domain. We may need to patch upstream.

samselikoff:
Oh woops, I think the test is wrong. If `/` is passed through, the request will be to `/` which is the test runner, which will return the HTML for the test page. I think I have it working.

samselikoff:
#683 closes, thank you!

PR #680, opened 2016-04-06
Fix typo in handledRequest string interp.

blimmer:
On master was seeing this in the output:
```
Successful request: GET $[request.url}
```
Subtle typo causing this problem.


samselikoff:
Ty sir!

Issue #658, opened 2016-03-22
add jsdoc style inline documentation to methods


bantic:
Mentioned by @samselikoff as a task that would be good for new contributors. Adding [JSDoc style](http://usejsdoc.org/howto-es2015-classes.html) comments for methods would help improve the overall quality of documentation.

samselikoff:
Some more details:

[YUIDoc](http://yui.github.io/yuidoc/syntax/index.html) is a syntax for adding comments to code. You add a comment to an object, method or property in a structured way, and then later you can run a tool to parse the comments from your code. The parsed comments can be used for all sorts of things, like autogenerating a documentation site. This is how [Ember's API docs](http://emberjs.com/api/) are generated - the descriptions for each object and method actually come from the code. This is great, because that means they stay up to date.

Here's an example of a YUIDoc for the `Ember.Route#serialize` method: https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js#L341-L347.

I'd love to get this into Mirage! We could eventually autogenerate the API docs from the code, which would be amazing.

A great place to start in Mirage would be the new `orm/Model` class.

- [Here's the class](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js)
- And [here's the docs from the site](http://www.ember-cli-mirage.com/docs/v0.2.0-beta.7/models/)
    - Note the methods in the "Class methods" section are on `schema.js`, so you can ignore those as a first pass.

The idea would be, basically copy the info from the docs, and write it as structured YUIDoc in the code, adding the important things like `@method`, `@param`, and so on, and marking methods `@private` or `@public`.

I would probably start with a single file and then PR it, and we can discuss from there.

samselikoff:
For anyone else interested in jumping in, check [Liz's PR](https://github.com/samselikoff/ember-cli-mirage/pull/682) for a great example on a single class.

lizzerdrix:
Trying to figure out which file to tackle next - any suggestions? I couldn't find any more that were quite as straightforward (in terms of their being similar enough to the docs that I could easily figure it out) as the ones I've already done. I can easily just pop in the basics (making note of the function name, params, and whether public/private) on any file, if that's helpful.

samselikoff:
[...]

PR #686, opened 2016-04-08
adds YUIDoc comments to addon/db-collection

lizzerdrix:
For issue #658 

I noticed that for `firstOrCreate` the documentation lists the second param as `attributesForCreate` while the code uses `attributesForNew`. Do you have a preference as to which should be referenced in the YUIDoc comments?

Also wasn't sure if I should document the `IdentityManager` methods at all. I didn't on this pass, but if I should, just let me know!

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -102,17 +102,17 @@ class DbCollection {
     `query` and optional `attributesForCreate`.
     @method firstOrCreate
     @param query
-    @param attributesForNew
+    @param attributesForCreate
     @public
   */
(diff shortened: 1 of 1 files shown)


samselikoff:
I think it should all be `attributesForCreate` since the record is immediately created (persisted). Would you mind changing the name in the code too?

Good to pass on IDManager for now, eventually we can pull it out into a separate file + document it.

lizzerdrix:
Cool! Yeah I thought about changing the code but wanted to ask first 🙂

Will do!

samselikoff:
Nice :+1:

PR #693, opened 2016-04-12
adds YUIDoc comments to addon/serializer

lizzerdrix:
Continuing on with #658 !

I left the examples that were in the previous comments, let me know if you would prefer they be removed (or if I should add the examples from the documentation to these comments). 

Also, feel free to let me know if you would prefer I add these to a bunch of files at once instead of one at a time!

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -19,6 +19,7 @@ function isNumber(n) {
   A collection of db records i.e. a database table.
   @class DbCollection
   @constructor
+  @public
 */
 class DbCollection {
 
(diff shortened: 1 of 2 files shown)


lizzerdrix:
Looks like ember-try is failing on beta. Not sure why, since I didn't change any actual code, but possibly because (as per `ember try:testall`): `Versions do not match: Expected: components/ember#beta but saw 2.6.0-beta.1 This might be ok, depending on the scenario`

Couldn't tell where this was configured, but maybe that's the issue? Not sure why the build would suddenly be failing on beta with no actual code changes.

Issue #669, opened 2016-03-30
Describe Mirage as a testing "fake"


mitchlloyd:
The term "mock" has been widely used in the Ember community to describe test double "fakes". It would be nice if Mirage could use the term "fake" to describe the pattern it enables or at least avoid using the term "mock".

Some of these terms are contentious, but the most common definitions I see have been summarized in this article:

http://martinfowler.com/articles/mocksArentStubs.html

<img width="654" alt="screen shot 2016-03-30 at 11 12 58 am" src="https://cloud.githubusercontent.com/assets/15169/14152502/7e3bd9da-f668-11e5-9018-fc91bd774612.png">

Sinon, for instance, describes this pattern accurately as a "fake server" (see the "Fake Server" section [here](http://sinonjs.org/docs/)). The book Growing Object Oriented Software Guided by Tests, walks through an implementation of an auction client that uses a "fake auction server" to stand in for their 3rd party API and describe it as "a minimal implementation just to support testing".

You can find implementations of mocks in many places (Ruby's [Mocha](http://gofreerange.com/mocha/docs/#Mock_Objects), [Sinon](http://sinonjs.org/docs/#mocks), Java's [Mockito](http://site.mockito.org/mockito/docs/current/org/mockito/Mockito.html#1). However, I've never seen this pattern in an Ember app.

Unfortunately, the term "mock server" may have taken on a new meaning for Ember developers at this point because Pretender describes itself as a "mock server". The Ember CLI docs [contain a bizarre section that juxtaposes "Mocks" and "Fixtures"](http://ember-cli.com/user-guide/#mocks-and-fixtures), but ultimately use the term mock to describe creating an Express server.

If you're amenable to this change I would be happy to submit a PR. I think the best approach would be to take the word "mock" out of the tagline and description, and include the term "fake" in the detailed documentation.

danilovaz:
I agree with you. The term "Fake" is the best approach on the current moment.

samselikoff:
Glad to have someone thinking about this. I'm not as familiar with the technical definitions but your references seem correct. Let's change it to "Fake server."

Also, I'm happy to use this more technically correct term throughout the docs, but perhaps we shouldn't even use any of this language on the homepage tagline. I want to make sure users understand what problem it solves for them, and if "fake" or "mock" gets them thinking in the weeds we might be able to do better.

Happy to change for now + to keep iterating on the messaging.

dlo83:
[...]

PR #695, opened 2016-04-13
Remove the term mock on the master branch

mitchlloyd:
To address #669 

* Changed the description in package.json
* Update the README

In the readme I think "fixtures" is a little better. I often seen people
turn to stubbing with fixtures first when trying to test against an API.
Using "mocks" outside of a testing context doesn't make sense.


samselikoff:
Thank you sir

PR #696, opened 2016-04-13
Remove "mock" references in documentation

mitchlloyd:
To address #669 on the gh-pages branch.

I audited the gh-pages brach for use of the word "mock". This is not a simple find/replace and needs some review before merging.

For the docs, I only changed the v0.2.0-beta.7 version.

Beyond my own pedantic hangups about terminology, I think this is a nice improvement. The work "mock" was used generally to mean:

* noun - "a mock" - a route handler
* noun - "a mock" - data returned from a route handler or stub data
* verb - "mock" or "mock out" - simulate a real API or implement a route handler
* adjective - "mock server" - fake server

Instead I replaced these uses with "fake", "simulate", "data", "stub", "route handlers", "Mirage" or anything else that seemed to convey the original intent.

### DIFF
--- a/docs/v0.2.0-beta.7/quickstart.md
+++ b/docs/v0.2.0-beta.7/quickstart.md
@@ -3,7 +3,7 @@ title: Quickstart
 version: v0.2.0-beta.7
 ---
 
-Mirage is all about faking your API server. You define *route handlers* to respond to your Ember app's AJAX requests.
+Mirage is all about simulating your API server. You define *route handlers* to respond to your Ember app's AJAX requests.
 
 Here's a simple example of a handler:
 
@@ -28,7 +28,7 @@ Now whenever your Ember app makes a GET request to `/api/authors`, Mirage will r
 
 ## Dynamic data
 
-This works, and is a common way to isolate tests from a real HTTP server - but hard-coded responses like this have a few problems:
+This works, and is a common way to simulate HTTP responses - but hard-coded responses like this have a few problems:
 
    - *They're inflexible*. What if you want to change this route's response data in your tests?
    - *They contain formatting logic*. Logic that formats the shape of your JSON payload (e.g., including the root `authors` key) is now duplicated across all your route handlers.
@@ -51,7 +51,7 @@ import { Model } from 'ember-cli-mirage';
 export default Model;
 ```
 
-The model will create an `authors` table in Mirage's *in-memory database*. The database makes our route handlers dynamic&mdash;we can change the returned data without rewriting the handler. In this way, we can share a single implementation for a route in both development and testing, while still having control over the response data.
+The model will create an `authors` table in Mirage's *in-memory database*. The database makes our route handlers dynamic&mdash;we can change the returned data without rewriting the handler. In this way, we can share the same route definitions in both development and testing, while still having control over their response data.
 
 Let's update our route handler to be dynamic:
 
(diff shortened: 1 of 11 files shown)


docs/v0.2.0-beta.7/quickstart.md
```diff
@@ -3,7 +3,7 @@ title: Quickstart
 version: v0.2.0-beta.7
 ---
 
-Mirage is all about mocking out your API server. You define *route handlers* to respond to your Ember app's AJAX requests.
+Mirage is all about faking your API server. You define *route handlers* to respond to your Ember app's AJAX requests.
```
samselikoff: I think I like "simulating" here

docs/v0.2.0-beta.7/quickstart.md
```diff
@@ -26,15 +26,15 @@ export default function() {
 
 Now whenever your Ember app makes a GET request to `/api/authors`, Mirage will respond with this data.
 
-## Dynamic mocks
+## Dynamic data
 
-This works, and is traditionally how HTTP mocking is done - but hard-coded responses like this have a few problems:
+This works, and is a common way to isolate tests from a real HTTP server - but hard-coded responses like this have a few problems:
```
samselikoff: "...and is a common way to simulate HTTP responses - but hard-coded..."

Since we're not talking about tests here yet I'd rather leave the concept of tests, isolation etc. out, and just focus on the differences between hand-rolling a fake server and using Mirage

docs/v0.2.0-beta.7/quickstart.md
```diff
@@ -51,11 +51,11 @@ import { Model } from 'ember-cli-mirage';
 export default Model;
 ```
 
-The model will create an `authors` table in Mirage's *in-memory database*. The database enables our mocks to be dynamic, and lets us change the return data without rewriting the entire mock. In this way, we can share a single set of mock routes in both development and testing, while still having control over the response data.
+The model will create an `authors` table in Mirage's *in-memory database*. The database makes our route handlers dynamic&mdash;we can change the returned data without rewriting the handler. In this way, we can share a single implementation for a route in both development and testing, while still having control over the response data.
```
samselikoff: how about

"In this way, we can share the same route definitions in both development and testing, while still having control over their response data."

docs/v0.2.0-beta.7/quickstart.md
```diff
@@ -253,16 +253,16 @@ Shorthands make writing your server definition concise, so you should use them w
 
 ## Passthrough
 
-Mirage is a great tool to use even if you're working on an existing app that doesn't have any mocks. By default, Mirage throws an error if your Ember app makes a request that doesn't have a corresponding mock defined. To avoid this, tell Mirage to let unhandled requests pass through:
+Mirage is a great tool to use even if you don't want to fake your entire API. By default, Mirage throws an error if your Ember app makes a request that doesn't have a corresponding route handler defined. To avoid this, tell Mirage to let unhandled requests pass through:
```
samselikoff: how about 

"... even if you're working on an existing app, or if you don't want to fake your entire API."

I've heard folks say before that they can't use Mirage since they're not starting a new app, so I just want to be explicit here

samselikoff:
This is awesome, I really appreciate your thinking + time on this.

mitchlloyd:
Normally I would squash, but I wanted you to be able to verify the changes and I'm excited to see if those fancy new Github features work. The maintainer should be able to squash the changes before merge.

samselikoff:
Fancy! Way better bc like you said I can see what's changed.

Thanks again man!

PR #698, opened 2016-04-14
adds YUIDoc comments to addon/serializer

lizzerdrix:
Continuing on with #658 ! Also, small addition to addon/db-collection to keep tests passing.

I left the examples that were in the previous comments, let me know if you would prefer they be removed (or if I should add the examples from the documentation to these comments).

Also, feel free to let me know if you would prefer I add these to a bunch of files at once instead of one at a time!

(closed the other PR because I have no idea why it kept failing, but this one shouldn't be)

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -19,6 +19,7 @@ function isNumber(n) {
   A collection of db records i.e. a database table.
   @class DbCollection
   @constructor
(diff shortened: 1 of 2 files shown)


samselikoff:
Thanks! Separate PRs is just fine.

PR #700, opened 2016-04-15
support aliases for --proxy flag (-pr, -pxy)

elbeezy:
The current `usingProxy` implementation only accounts for the long form of the `--proxy` flag. Added checks for `-pr` and `-pxy` aliases.


samselikoff:
Thanks!

PR #702, opened 2016-04-16
Slight elaboration of shorthand usage in config.js

BrianSipple:
I found myself adding this brief elaboration on the concept of shorthands to my `mirage/config.js` files. 

I know an effort was made to clean the [comments in this file a short while ago](https://github.com/samselikoff/ember-cli-mirage/commit/0fac9db036a58943568c2af686206ec521f870f4), but perhaps this would still be a handy -- while also clean & concise -- snippet for anyone new to Mirage?

**P.S: regarding the failing build**: this might be contingent upon #698.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,8 @@ Changes:
   - [FEATURE] #655 Add importable rest-serializer @rondale-sc
   - [FEATURE] #269 Dynamic factory attributes can reference other dynamic attributes @lazybensch
   - [FEATURE] #603 Support inverse foreign keys @ef4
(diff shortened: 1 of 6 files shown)


samselikoff:
We used to have more examples in the config but I think I prefer sending folks to the docs. This is how Rails' routes.rb file evolved as well.

Instead of this change how about we post a link to the shorthands reference?

http://www.ember-cli-mirage.com/docs/v0.2.0-beta.7/shorthands/

BrianSipple:
Perfectly okay with me. And in fact, leveraging the docs is, indeed, so much better now that I think about it.

There's the slight issue of keeping the versioning up to date, but that shouldn't be any trouble once we get to a stable 0.2 (and beyond).

PR #705, opened 2016-04-17
Update collections

samselikoff:
- [x] Add Collection.models plain array
- [x] Pluralize schema classes (`schema.user` => `schema.users`)
- [x] add this.serialize to route handlers


samselikoff:
woot - last breaking changes for 0.2! It's about time we release.

PR #709, opened 2016-04-19
Serializers can incude related resources using dot-path notation

### LINKED ISSUE #708: Include second degree objects from the serializer
The JSON-API spec, in the section about including related resources (http://jsonapi.org/format/#fetching-includes) allows to specify related resources of Nth degree using a dot-separated name.

Per example, when requesting a `Quiz` model, you can now include the related questions doing `/quizzes/1?include=questions`. If you also want to get the answers associated to those questions your can perform this query: `/quizzes/1?include=questions,questions.answer`. 

**This already works in mirage** when passing a `include` query param in the request, but no when using the `include` property in the serializer.

For consistency, ember-cli-mirage should make a serializer like this behave the same as if the `include` property was passed in the request.

```js
import BaseSerializer from './application';
export default BaseSerializer.extend({
  include: ["questions.answer"]
});
```

I could work on this if nobody is tacking it or any related work.

cibernox:
For now this is just a failing test demonstrating the issue.

Closes #708

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -242,20 +242,23 @@ class JsonApiSerializer {
 
   _getRelationshipNames(request = {}) {
     let requestRelationships = _get(request, 'queryParams.include');
+    let relationships;
 
     if (_isString(requestRelationships)) {
-      if (requestRelationships.length) {
-        let relationships = requestRelationships
-          .split(',')
-          .map(_trim)
-          .map((r) => r.split('.').map((_, index, elements) => elements.slice(0, index + 1).join('.')));
-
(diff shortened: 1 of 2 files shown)


samselikoff:
@cibernox I just merged #705, feel free to move forward on this

cibernox:
@samselikoff Done. 

The logic I've decided to follow is:
1. If the request contains an `included` param, that wins over any configuration, even if that `included` params is an empty string (that would be like asking to *not* include related resources to an endpoint that does include them by default)
2. If the request doesn't include it, use the serializer's default config.

samselikoff:
Looks great! Sorry about the delay, and thanks again :+1:

PR #712, opened 2016-04-21
Add multi-attribute update documentation

backspace:
Hey, I _suspected_ today that this would be possible and it was, so this just makes it explicit in the documentation. I’m not sure what the procedure is for documentation-only updates, hopefully this is the correct thing to do!


samselikoff:
Perfect, thank you!

Issue #723, opened 2016-04-29
Add shorthand for HEAD request


timjcook:
In the same way that we have shorthands for `get`, `post` etc. it would be nice to add another shorthand for a `HEAD` request. I can't seem to define a route that will handle this request and thus it breaks my acceptance tests when that request is sent.

Tested on `ember-cli-mirage 0.1.13`.

raido:
Same with Beta.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/commit/222a4b0815161de038942954f78922388193d234 closes

PR #724, opened 2016-05-03
Add HEAD request route handler shorthand.

timjcook:
refers to #723
 
- [x] Added head request shorthand
 - returns 204 if resource was found
 - returns 404 if a resource was expected to be found but wasn't

### DIFF
--- a/addon/route-handler.js
+++ b/addon/route-handler.js
@@ -7,6 +7,7 @@ import GetShorthandHandler from './route-handlers/shorthands/get';
 import PostShorthandHandler from './route-handlers/shorthands/post';
 import PutShorthandHandler from './route-handlers/shorthands/put';
 import DeleteShorthandHandler from './route-handlers/shorthands/delete';
+import HeadShorthandHandler from './route-handlers/shorthands/head';
 
 import _keys from 'lodash/object/keys';
 import _isArray from 'lodash/lang/isArray';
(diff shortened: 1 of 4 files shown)


samselikoff:
Thanks for working on this!

Do the shorthands make sense for HEAD? Is HEAD ever used to return resources like GET is?

timjcook:
@samselikoff I don’t think so, as far as I can tell a `HEAD` request will never return any content in the message body, just the status code along with other meta data. I’m not quite sure what the correct status code to return is for this request. Some suggest it should mirror the `GET` request to the same route, but rails seems to always return a `204`, hence my initial commit, but happy to change if others think differently. My use case only required a successful response so the action could proceed.

samselikoff:
In this case in might make sense to not add shorthands at all, just the actual `this.head` method and let folks do something like

```js
this.head('/users', undefined, 204);
```

The shorthands are meant to interact with db data.

What do you think?

timjcook:
I think that would be a fair way to do it and would remove the overhead of an extra handler. However it would, I think, introduce the possibility that someone could stub out say a route like:
```javascript
this.head('/users/:id', undefined, 204);
```
and always return `204`'s, but then if they didn't set up a `User` in their scenario and they made a get request to that route, it would return a `404`, which would be inconsistent with a real API.
So I guess we would be relying on people setting up their test environment correctly.

samselikoff:
Ohhh I see - I totally didn't understand your original point that the shorthand just responds with the correct status code, but an empty body. This looks great to me.

I'd say, let's just delete the array shorthand & tests. Doesn't really make sense here. (I don't really like the array shorthand anymore on any of the verbs tbh, now that we have orm & serializers)

timjcook:
Fantastic, I've removed the array shorthand and tests now so hopefully it is ready to go. Just let me know if you think it needs anything more.

samselikoff:
:+1: thank you!

PR #728, opened 2016-05-05
[DOCS] Added information about custom query ability

acorncom:
Thought we should probably document this functionality, as it'd be really useful to use more regularly ;-)


samselikoff:
Thank you!

PR #739, opened 2016-05-11
Issue 738 config directory

gthmb:
* Updated the index.js file to look for the `directory` setting in the config block
* Updated the index.js files in each respective blueprint to check for the directory setting
* Squashed the `mirage` folder in each blueprint to prevent creating file into a `mirage` folder inside of your defined directory


samselikoff:
thanks @gthmb!

gthmb:
My pleasure. Thanks for the great library!

PR #744, opened 2016-05-19
Fix the wrong path of ApplicationSerializer file.

kevinhyunilkim:
ApplicationSerializer's file path `application/serializer.js` should be `serializers/application.js`


samselikoff:
Thanks!

PR #748, opened 2016-05-23
update error url

koriroys:
links to a dead page.

should http://www.ember-cli-mirage.com/docs/v0.2.x/serializers/#normalizejson redirect to http://www.ember-cli-mirage.com/docs/v0.2.0-beta.9/serializers/#normalizejson instead of hardcoding it?


samselikoff:
Probably we'll change once 0.2.x ships

PR #757, opened 2016-05-31
update links to v0.2.x docs

HusseinMorsy:
after 0.2.0 final dropped, I found some links in blueprint and in route-handler which links to 0.2.0.beta versions.


samselikoff:
Thank you!

Issue #754, opened 2016-05-31
Ad hoc includes for the serializer layer


samselikoff:
Currently, Mirage's ORM only supports basic hasMany and belongsTo relationships. This makes it hard to fake out responses with many to many relationships that do not include the join model.

Originally I was going to add a manyToMany helper to the ORM, but that involves a lot of work, as it touches the foreign keys + makes assumptions about the database. A better short-term solution, to help folks out today, is to update the Serializer layer.

Let's say we have a manyToMany between posts and tags. In the ORM, we still need to set up the join model. So we'd have

``` js
// mirage/models/post.js
import { Model, hasMany } from 'ember-cli-mirage';
export default Model.extend({
  postTags: hasMany()
});

// mirage/models/post-tag.js
import { Model, belongsTo } from 'ember-cli-mirage';
export default Model.extend({
  post: belongsTo(),
  tag: belongsTo(),
});

// mirage/models/tag.js
import { Model, hasMany } from 'ember-cli-mirage';
export default Model.extend({
  postTags: hasMany()
});
```

This simulates the backend setup with the join table. Create related data like so:

``` js
let post = server.create('post');
let economics = server.create('tag', { title: 'Economics' });
let programming = server.create('tag', { title: 'Programming' });

server.create('post-tag', { post, tag: economics });
server.create('post-tag', { post, tag: programming });
```

This sets up the database correct; the trick is, how to get the Serializer to return the m2m between the posts and the tags, without the join model. We basically want to simulate something like this:

```
GET /posts?include=tags

{
  data: [{
    type: 'posts',
    id: 1,
    attributes: {}
    relationships: {
      tags: {
        data: [
          { type: 'tags', id: '1' },
          { type: 'tags', id: '2' }
        ]
      }
    }
  }],
  included: [{
    type: 'tags',
    id: '1',
    attributes: {
      title: 'Economics'
    }
  }, {
    type: 'tags',
    id: '2',
    attributes: {
      title: 'Programming'
    }
  }]
}
```

I think the easiest way to do this is to get the Serializer to accept adhoc `includes`, and to check for locally defined methods, similar to how active_model_serializers does it.

This is what I want to see:

``` js
// serializers/post.js
export default ApplicationSerializer.extend({

  include: ['tags'],

  tags(post) {
    let tags = post.postTags.map(pt => pt.tag);
    return new Collection('tag', tags);
  }

});
```

[...]

PR #758, opened 2016-06-01
Add support for a manyToMany relationship through hasMany

begedin:
This should now be good to review, albeit possibly not good to merge yet.

It works, and I've added a few tests, but I'm worried I may have missed a thing or two.

The whole PR was done in 2 steps, for the most part:

1. Refactor `JsonApiSerializer` to extend from `Serializer` base
2. Add support for many to many relationships by providing a passthrough function for the relationship through the serializer, as described in #754 

Some notes:

* The behavior of the hasMany relationship makes it so that having a passthrough model will create a useless foreignKey on both ends of the many-to-many

```Javascript
// example

//user.js

export default Model.extend({
   organizations: hasMany(),
   organizationMemberships: hasMany()
});

// organization

export default Model.extend({
  users: hasMany(),
  organizationMemberships: hasMany()
});

// organization-membership

export default Model.extend({
  user: belongsTo(),
  organization: belongsTo()
});


// proper setup for user serializer

  include: ['organizations'],
  organizations(model) {
    let models = model.organizationMemberships.models.map(om => om.organization);
    return new Collection('organization', models);
  }
});

// proper setup for organization serializer

  include: ['users'],
  users(model) {
    let models = model.organizationMemberships.models.map(om => om.user);
    return new Collection('user', models);
  }
});
```
With the scenario above, when properly configured, the user will end up having the following associations and fks: `organization, organizationId, organizations, organizationIds, organizationMemberships, organizationMembershipIds`. The first two in the list are completely useless and are autogenerated due to having a `belongsTo` in the organization membership model. The equivalent of this will happen with the organization model.

* I believe we could push this a step further. The "proper setup" should always be the same, so we could simply add an option to the `hasMany` class:

```Javascript
//user.js

export default Model.extend({
   organizations: hasMany({through: 'organization-memberships'})
});

// organization

export default Model.extend({
  users: hasMany({through: 'organization-memberships'})
});

export default Model.extend({
  user: belongsTo(),
  organization: belongsTo()
});
```

Then, the hasMany class could check this option and auto-generate the required function on the model, as it does the other properties and functions.

We could also add an option to the link model to ensure the belongsTo does not generate a useless fk and property on the end models. We would  have to be careful there, though, since the link model might have additional properties and an actual use other than just being the link.


begedin:
@samselikoff I believe this is ready for an initial review. I'm expecting I'll need tweaks and fixes, but it seems to be working for now.

samselikoff:
@begedin thanks for kicking this off! /cc @JoshSmith

1. Let's drop the acceptance test. It's ok for Mirage to have a few of these, but they add a ton of code and have code split up all over the place. For this permutation, no need to have an acceptance test. So let's get rid of the changes to

  - tests/acceptance/many-to-many-test.js
  - tests/dummy/app/models/address.js
  - tests/dummy/app/models/contact.js
  - tests/dummy/app/templates/contact.hbs
  - tests/dummy/mirage/config.js
  - tests/dummy/mirage/factories/address.js
  - tests/dummy/mirage/models/address.js
  - tests/dummy/mirage/models/contact-address.js
  - tests/dummy/mirage/models/contact.js
  - tests/dummy/mirage/serializers/address.js
  - tests/dummy/mirage/serializers/contact.js

2. Your `tests/integration/serializers/base/associations/many-to-many-test.js` test is a good start. However it uses `schemaHelper` (understandably, since that's what the other tests use). I've started to prefer having all the setup in the test file, though, so it's less confusing. In this case let's follow the example in a test like [customized normalize method](https://github.com/begedin/ember-cli-mirage/blob/master/tests/integration/server/customized-normalize-method-test.js), and set up the entire server within the test. So instead of `schemaHelper.setup()` we'll just instantiate the entire server there, with all relevant models + serializers defined.

3. To your point about the useless foreign keys, you're absolutely right. The problem is that `hasMany` and `belongsTo` were built for, and thus only work for, standard one-to-many relationships. Because that's all they work for, we're building this serializer layer as a workaround. That means you would never set up the domain model as you have in `schema-helper`, i.e. you would never do

  ```js
  contact: Model.extend({
    addresses: hasMany(),
    contactAddresses: hasMany()
  })
  ```

  since a contact *does not* `hasMany` addresses - it has many *through* contact-adresses (as you pointed out). Because of this, you wouldn't put that on the model.

  Originally I was going to add the `hasMany({ through: '' })` option - but if you read [this post](http://www.ember-cli-mirage.com/blog/2016/01/03/0-2-0-beta.9-released/) you'll see why this needs some more thought. We may still end up adding this convenience method, but it's beyond the scope of this PR, and will happen after the fundamental storage piece of the ORM is reworked.

  That means that your model definitions will stay standard:

  - a contact hasMany contactAddresses
  - an address hasMany contactAddresses
  - a contactAddress belongsTo both a contact and an address

  Basically, in `/models` you'll still always have to write out the join model. If your server doesn't expose it, though (and presumably your Ember app doesn't have it defined), that's when you'll define the m2m directly on the serializer, and it will *appear* to the Ember app as if your mock server has these resources modeled as a m2m. Does that make sense?

  So, when you rewrite the tests setting up the models and serializers directly inside, you'll keep the modeling standard (as I described above), and then define the `addresses` method directly on the `contact` serializer. That way the response will appear as if these are modeled using m2m.

4. In your many-to-many test you assert on `data`, and the mock response includes `contact-addresses`. This is wrong, it should only include the relationships defined on the serializer. Since the contact serializer only has an `addresses` relationship defined on it, that's the only thing that should appear under the `data.relationships` key in the json.

5. To your point about needing additional data on join record, well in that case this would fall back to simple `one-to-many` and we wouldn't need the `m2m` anymore. If you have a join record with more data - for example, a Doctor hasMany Patients through Appointments, but the Appointments have a date and time - you would be modeling those in your Ember app anyway, and it would all be one-to-many:

  - Doctor [one-to-many Appointments]
  - Appointment [belongs-to Doctor, belongs-to Patient]
  - Patient [one-to-many Appointments]

begedin:
@samselikoff I'll try and make a to-do list based on what you wrote:

* [x] Remove acceptance test and associated file changes
* [x] Rewrite integration tests so they don't user schema helper. **I went a step further here** and rewrote them so they don't instantiate a server at all. Instead, they just instantiate a schema and a serializer registry. We don't need server mocks for these tests.
* [ ] Rewrite the behavior so that the resulting json payload doesn't include the link relationship (`contact-addresses` in our case).

I got the last one from 

>In your many-to-many test you assert on data, and the mock response includes contact-addresses. This is wrong, it should only include the relationships defined on the serializer. Since the contact serializer only has an addresses relationship defined on it, that's the only thing that should appear under the data.relationships key in the json.

However, I'm not really sure how to go about achieving this. Just to make sure I'm understanding this correctly, would this be our proper setup?:

```Javascript
// app/models/addresses

  contacts: DS.hasMany()
});

// app/models/contacts

  addresses: DS.hasMany()
});


// mirage/models/addresses

export default Mirage.Model.extend({
  addresses: Mirage.hasMany(),
  contactAddresses: Mirage.hasMany()
});

// mirage/models/contacts

export default Mirage.Model.extend({
  contacts: Mirage.hasMany(),
  contactAddresses: Mirage.hasMany()
});

// mirage/models/contact-addresses

export default Mirage.Model.extend({
  address: Mirage.belongsTo(),
  contact: Mirage.belongsTo()
});

//mirage/serializers/address

export default Serializer.extedn({
  include: ['contacts'],
  contacts(model) {
    let models = model.contactAddresses.models.map(ca => ca.contact);
    return new Collection('contact', models);
  }
});

// mirage/serializers/contact

export default Serializer.extedn({
  include: ['addresses'],
  addresses(model) {
    let models = model.contactAddresses.models.map(ca => ca.address);
    return new Collection('contact', models);
  }
});


````

If this is the correct setup, then I'm not seeing a way to auto-detecting that the `contact-relationships` key should not be under the relationship object.

Are we talking about serializers in general, or the `JSONAPISerializer` specifically. With the `JSONAPISerializer`, the `relationship` object is always populated with all relationships the server renders, while the `included` object is the one that has or does't have all the entire payload of individual relationship records.

If you take a look at `json-api-serializer/associations/included-test`, you'll notice that this behavior is present there as well.

I agree that it would make sense not to have `contact-addresses` in the resulting payload at all for this specific case, but I don't think we have enough information with our setup to achieve something like that.

**Comment by samselikoff** [comment]
I didn't fully understand until now, but you're right. I was thinking of how, in the Rails case, not all model relationships are automatically serialized. I'm thinking specifically of the `json-api-resources`, which serves as a layer of abstraction on top of the model layer. In that case, only the relationships specified on the resource objects are always populated, which means things like join tables aren't. Since we don't have a layer of abstraction on top of the ORM, what you're saying is absolutely correct. All relationships known to the server should always be included in the `relationships` key, meaning the `contact-addresses` join will be there, too.

We'll need to add an API for blacklisting attrs to be returned, so users can mock out a m2m without including the join model. I'll do that in a separate PR, so for now, your test assertions are correct.

I think your Mirage modeling is still off. You want

```js
// mirage/models/addresses

export default Mirage.Model.extend({
  contactAddresses: Mirage.hasMany()
});

// mirage/models/contacts

export default Mirage.Model.extend({
  contactAddresses: Mirage.hasMany()
});

// mirage/models/contact-addresses

export default Mirage.Model.extend({
  address: Mirage.belongsTo(),
  contact: Mirage.belongsTo()
});
```

i.e. you need to model using the join record. It's the Serializer layer that simulates the direct `hasMany`.

begedin:
@samselikoff If we do as you describe, then there will be no way for the serializer to know that there needs to be an `addresses` relationship to serialize on a `contact` and vice-versa.

With the base `Serializer` (and thus, `ActiveModelSerializer` and `RestSerializer`), it could work. We could simply have: 

```Javascript
//mirage/serializers/address

  include: ['contacts'],
  contacts(model) {
    let models = model.contactAddresses.models.map(ca => ca.contact);
    return new Collection('contact', models);
  }
});

```

However, with `JSONAPISerializer`, the following is possible:

```Javascript
//mirage/serializers/address

  // no include specified
  contacts(model) {
    let models = model.contactAddresses.models.map(ca => ca.contact);
    return new Collection('contact', models);
  }
});

```

Setting it up this way means that contact addresses and relationships will be part of the `relationships` object in the payload, but will not be present in the `included` array of the payload. With that sort of setup, and without a `hasMany` for for `contacts` specified in the `address` model, there is no way for the serializer to know there is any relationship to `addresses` at all.

To elaborate, here is what the `_serializeModel` function of the `JasonApiSerializer` looks like:

```Javascript

_serializeModel(model, request) {
    this._augmentAlreadySerialized(model);

    let response = {
      data: this._resourceObjectFor(model, request)
    };

    this._serializeRelationshipsFor(model, request);

    return response;
  }

```

The key here is `_resourceObjectFor`. This what crates the `data: { attributes: {...}, relationships: {...} }` object of the payload. To create the relationships, it uses the following code:

```Javascript
    model.associationKeys.forEach(camelizedType => {
      let relationship = this._getRelatedValue(model, camelizedType);
      let relationshipKey = this.keyForRelationship(camelizedType);

      // transform each relationship into { type: relationshipKey, id: relationship.id }
      // add link data
    });
```

On the other hand, the base `Serializer` only relies on the contents of the `include` property specified in the serializer.

The `JsonApiSerializer` **needs to rely on `model.associationKeys.forEach`** because of the "relationships vs included" behavior of the JSON API specification we described above.

If address doesn't have a `hasMany` for contacts, then the address model will not have a `contacts` association key, so `addresses` will not be listed in the `relationships` object of the resulting payload.

samselikoff:
But the ad-hoc extensions should be able to take care of this. Meaning, we should modify `JsonApiSerializer` to not only look for `model.associationKeys`, but also any locally defined ad-hoc methods that may return Models or Collections.

Does that make sense? Do you see any issues with that?

begedin:
The issue is that there is no way for the serializer to know the local methods are "ad-hoc associations". As I said, without a `contacts: hasMany()` in the address model, there is no way for the address serializer to know there is supposed to be an `contacts` collection on the address model. With the base serializer, and this, the `RestSerializer` and `ActiveModelSerializer`, any association will be mentioned in the `includes` property, since there is no other way for it to be part of the final payload. 

For `JsonApiSerializer`, however, an association doesn't have to me be mentioned in the `includes` property, but it still needs to be part of the final `relationships` object in the payload.

As a specific example:

```Javascript

// app/models/address.js

  contacts: DS.hasMany({ async: true }) // a man-to-many relationship
});

// mirage/models/address.js

export default Mirage.Model.extend({
  contactAddress: hasMany()
});

// mirage/serializers/address.js

export default JSONAPISerializer.extend({
  contacts(model) {
    let models = model.contactAddresses.models.map(ca => ca.contact);
    return new Collection('contact', models);  
  }
});

```

Would this be the correct configuration in your opinion?

In that case, `model.associationKeys` does not contain `contacts` and neither does the `include` property of the serializer. 

However, a valid JSON API response we would get from an API in an equivalent scenario (and what the actual Ember `DS.Model` is expecting would look something like:

```Javascript
{
  data: {
    attributes: {}
    relationships: {
      addresses: {
        data: [{ id: "x", type: "addresses" }, { id: "y", type: "addresses" } /* and so on */]
      }
    }
  }
}
```

There is no includes in the payload, but the addresses are still there in the relationships. Since the mirage serializer has no knowledge of addresses, it will not know to call the method we defined at any point, and the resulting payload will look more like:

```Javascript
{
  data: {
    attributes: {}
    relationships: {
      "contact-addreses": {
        data: [{ id: "x", type: "contact-addresses" }, { id: "y", type: "contact-addresses" } /* and so on */]
      }
    }
  }
}
```

What we wan't it to look like in this PR, I believe is a payload that contains both `addresses` and `contactAddresses` in the relationship object and without the knowledge of addresses being a relationship, I can't think of a practical way to do this.

samselikoff:
Just to jot this down here before Slack history pwns us, @begedin and I were discussing this in Slack and I think an API like this might work:

```js
// mirage/serializers/organization.js
import { toMany, toOne } from 'ember-cli-mirage';

  users: toMany({
    models(organization) {
      return organization.organizationMemberships.models.map(om => om.user);
    },
  }),
  
  primaryAddress: toOne({
    type: 'address',
    model(organization) {
      return organization.addresses.first();
    }
  })
});
```

Relationship name is based on key. Default `type` for JSONAPI resource object is based on key, but can be overwritten. `model` or `models` is used for the meat, based on whether it's `toOne` or `toMany`.

samselikoff:
@begedin are you going to have more time to work on this? If not, it might be best for me to pick it up this weekend

begedin:
@samselikoff First off, using the approach we discussed, we would even be able to do something along the lines off


```Javascript
// mirage/serializers/organization.js
import { toMany, toOne } from 'ember-cli-mirage';

  users: toMany({ throughKey: 'userOrganizations', sourceKey: 'organization' });
});

```

Basically allowing us to use a default method, instead of us providing it (while still being possible to provide it if necessary).

I tried getting this to work as discussed, but I don't think we should push this without a decent amount of refactoring first.

As it is right now, the base `Serializer` (so `ActiveModelSerializer` and `RestSerializer` as well) completely ignore `model.associationKeys` when figuring out which associations the model has. Instead, they rely solely on the value provided (or returned) by the `include` property/function of the serializer.

On the other hand, the `JsonApiSerializer` relies purely on `model.associationKeys` to figure out associations, and only uses the `include` property to figure out which records to sideload into the `included` part of the payload.

Firstly, the base serializer should differentiate between a list of `relationshipNames` and `namesOfRelationshipsToInclude`. I wen't in the wrong direction here initially, when I renamed the base `valueForInclude` into `_getRelationshipNames`, thinking it's the same thing as the method contained in the `JsonApiSerializer`.

I think the serializer should have `_getRelationshipKeys` and `_getIncludedKeys`. In the case of serializers other than the `JsonApiSerializer`, where these are the same, one method can simply delegate to the other, but there should still be a difference, to allow for easier inheritance.

Also, currently, the JsonApiSerializer managed the `included` part of the resulting payload by pushing into `this.included`directly, while the base serializer instead passes along an `allAttrs` object through the same chain of methods, making the behavior slightly different and making it hard to organize code that's pretty similar otherwise (mostly, I'm talking `_serialize[Included][Model/Collection]` here.

I'm making some progress here, but it's difficult to figure it out. I'll keep you updated as I go along.

samselikoff:
Thanks for your work.

For some background JSONAPISerializer was implemented separately and after the rails-style serializers. It was coded separately since it was so different. I was worried about cramming too much into a base class and making unnecessary abstractions.

If you think there's enough in common here, then I agree we should have a similar interface across all serializers. Alternatively, since the JSONAPISerializer actually has a spec behind it, it can be kept separate from the others, if you think that's easier and more maintainable. It does seem there's a lot of shared code between them, however.

begedin:
I believe it's worth consolidating, since yes, as you said, there's a good amount of common code. Seeing how `include` works on the base serializer, though, makes me wonder if this is actually correct.

According to the behavior, if we have a `user`, with a `blogs: hasMany()`, for example and with embedding set to false, it seems that the only two possible payloads we can get are

```
// if user serializer is configured to include blog

{
  user: { id: 1, /* other attributes */, blogIds: [1] },
  blogs: [{ id: 1, /* other attributes */, userId: 1 }]
}

// if user serializer is not configured to include blog

{
  user: { id: 1, /* other attributes */ }
}

```

This means that the payload, when not including the blog relationship in it looks as if there is no blog relationship to begin with. Shouldn't it actually look like

```Javascript
{
  user: { id: 1, /* other attributes */, blogIds: [1] }
}
```

Allowing blogs to be fetched later asynchronously. I'm not really an expert on formats other than JSON API, so I'm not sure, but it seems reasonable to me.

Basically, to describe in a simplified manner, the JsonApiSerializer behaves in the following way:

* Include ids of all relationships of the parent object
* If some relationships are listed in the `include` property of the serializer, then also include full information for those relationships.

On the other hand, the other three serializers (due to extending the base), behave in the following way:

* Include ids and full payloads of those relationships of the parent object listed in the `include` property of the serializer
* Ignore and do not provide any other relationship information whatsoever.

Is this the correct behavior, or is the base serializer misbehaving here?

samselikoff:
What you're saying is correct, and needs to be changed. The current implementation, as you state, is a two-way switch: 

 - if `embed` is true, both the `blogIds` will be included, and the `blogs` will be sideloaded
 - if `embed` is false, neither of these things will happen

There's a few ways we could approach this. Let me think for a bit and get back to you.

samselikoff:
Here's an option. Say you want to `GET /authors/:id` and have it sideload posts but not addresses. You want this:

```
GET /authors/1

{
  author: {
    id: 1,
    post_ids: [1, 2, 3],
    address_ids: [1, 4]
  },
  posts: [
    { id: 1, title: 'Lorem' },
    { id: 2, title: 'Lorem' },
    { id: 3, title: 'Lorem' },
  ]
}
```

Notice the `address_ids` are there but not included, for later async fetch. 

We could add a `serializeRelationshipIds` option to serializers. It can take three values:

 - `"when-included"` would mean, only serialize the relationship ids on the parent when the relationship is being included (sideloaded). This is the default and current behavior.
 - `"always"` would mean, always include ids on the parent, regardless if that relationship is being included or not
 - finally it could also be a function, which takes the key of the relationship and the relationship itself, and returns `true` or `false`, indicating whether you wanted that relationship's ids to be on the parent:

  ```js
  serializeRelationshipIds: 'when-included',

  serializeRelationshipIds: 'always',

  serializeRelationshipIds(key, relationship) {
    if (key === 'some-weird-join') {
      return false;
    } else {
      return true;
    }
  }
  ``` 

What do you think?

begedin:
@samselikoff That would work. I guess the real problem is, that there is no official spec for `RestSerializer` or `ActiveModelSerializer`? As in, what is the "offical" default behavior. `'always'` or `'when-included'`? 

If there is a spec that covers this, we could just make that behavior the default and let the user do their own customization if they want to. To me anyway, including relationship ids always makes more sense. Embed should only affect how "included" models are serialized (as children of the parent, or in a side collection).

In the case of JsonApiSerializer, where this is the correct behavior, we would have 3 required functions.

* `getRelationshipKeys/Names` - gets a concatenation of `model.associationKeys` and any `hasMany` keys defined on the model.
* `getIncludedRelationshipKeys/Names` - basically just the value of `include` or `query.include`. `hasMany` has no special behavior here. If it's to be included, it should be listed inside one of those two.
* `getRelationshipValue(model, relationshipKey)` - returns one of the following
  * if there is a `hasMany` with that key defined on a serializer, return `serializer[relationshipKey].getCollection().models
  * otherwise, get the value of `model[relationshipKey]` or `model[relationshipKey].models` depending on if `model[relationshipKey]` is a `Model` or a `Collection`

lolmaus:
Hey @begedin and @samselikoff! :wave: 

@Sweet-Bob has tried this branch to set up a very basic many-to-many relationship and ran into a weird problem. Looks like this implementation is broken and requires work!

He mentioned this issue in the [Russian-speaking Ember chatroom](https://gitter.im/dev-ua/ember). Being surprised by Mirage's lack of many-to-many support, I rushed to create a failing test. Here's a PR against @begedin's branch: https://github.com/begedin/ember-cli-mirage/pull/1

The failing test is pretty straightforward, please have a look: https://github.com/lolmaus/ember-cli-mirage/blob/10c1410f45e2dde53796da268408ece66c068bf5/tests/integration/orm/many-to-many-test.js

I bet a beer you won't be able to guess its results. :beer :trollface:

samselikoff:
@lolmaus thanks for the additional info, and sorry for the delayed response.

I am actually still working on this, as the direction in the current PR here was not quite right.

Many-to-many is more complicated, as it opens questions of whether we create an implicit join table, or come up with another method to store the relationship data. I am still thinking through this but think I have a solution. In the meantime, this PR will provide a way to define ad-hoc relationships at the serializer level, to make mocking m2m easier even before proper ORM support lands.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/781 is going to be the new PR

JoshSmith:
@samselikoff do we want to close this then?

samselikoff:
@JoshSmith yes we should - I've pulled in @begedin's commit into my PR so he will get recognition there.

PR #763, opened 2016-06-06
Document usage of `?include` query param

### LINKED ISSUE #659: Document `?include` query param support in JSONAPISerializer
It works. we just need to write about it!

alecho:
This will close #659.

Let me know if we should say anything about a comma separated list of relationships paths or dot-serperated relationships paths.

### DIFF
--- a/docs/v0.2.x/serializers.md
+++ b/docs/v0.2.x/serializers.md
@@ -167,8 +167,10 @@ GET /authors/1
 
 ## include query param
 
+*Note: This is only available when using the JSONAPISerializer.*
+
 The JSONAPISerializer supports the use of `include` query parameter to return compound documents.
-Prior to Ember Data 2.5, you will need to add `'ds-finder-include': true` to your app FEATURE object:
+Prior to Ember Data 2.5, you will need to add `'ds-finder-include': true` to your app FEATURES object:
 
 ```js
 // config/environment.js
@@ -179,6 +181,7 @@ var ENV = {
     }
   }
 };
+```
 
 To tell Mirage to sideload blogPosts when we find find all authors we can do something like the following:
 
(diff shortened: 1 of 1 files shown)


docs/v0.2.x/serializers.md
```diff
@@ -165,6 +165,102 @@ GET /authors/1
...
+```js
+// routes/authors.js
+export default Ember.Route.extend({
+  model() {
+    return this.store.findAll('author', {include: 'blogPosts'});
+  }
+}
+```
+
+The above will make a GET request to `/api/authors?include=blogPosts` which will return something similar to the following:
```
samselikoff: Let's clarify a bit here, something like:

The above will make a GET request to `/api/authors?include=blogPosts`, and then the appropriate Mirage route handler will be invoked. When it comes time to serialize the response, the JSONAPISerializer will inspect the query params of the request, see that the `blogPosts` relationship is present, and then proceed as if this relationship was specified directly in the `include: []` array on the serializer itself.

Note that, in accordance with the spec, Mirage gives precedence to an `?include` query param over a default `include: []` array that you might have specified directly on the serializer. Default includes will still be in effect, however, if a request does not have an `?include` query param.

samselikoff:
Awesome work! Few things:

1. Incorporate my in-line feedback
2. I think we should drop the json example. The docs are getting long as it is and it just adds a bit too much noise. I think we can do without the example at all, just keep the rest of the explanation as you have it.

alecho:
I agree with removing JSON example. I didn't squash the commits yet so that the diff of my changes is still on Github but I will squash before this is good to merge.

samselikoff:
I can do it now from GH 💃 

Thanks!

PR #764, opened 2016-06-06
Ensure `mirage` tree is linted.

### LINKED ISSUE #714: Lint not running on `mirage` directory in v0.2.0-beta8
## The problem
I am running v0.2.0-beta8 and noticed that JSHint and ember-cli-eslint do not run on the `mirage/` directory now that the `mirage/` directory is in the project root and not in the `app/` directory.

## To reproduce
You should be able to easily reproduce this by removing a necessary semicolon on a JS file within the `mirage/` directory. You will notice that this does not cause a JSHint error or ember-cli-eslint error (whichever you are using).

## My attempts to fix
Digging in a little further I noticed that there is a `lintTree` hook for addons. This hook is actually used to generate tests given an already determined lint tree that is passed into the hook. I believe this is the hook that JSHint and ember-cli-eshint use to generate the linting tests you see in your test output. According to [the documentation](http://ember-cli.com/api/classes/Addon.html#method_lintTree) hook only gets invoked for the `app`, `test`, and `addon` trees. So I believe this is where the problem lies, it is never called for the `mirage` tree and the `mirage` tree is not included in any of the other linted trees.

I notice the index.js in ember-cli-mirage merges the `mirage/` directory into the app tree here: [index.js#L54-L61](https://github.com/samselikoff/ember-cli-mirage/blob/master/index.js#L54-L61). Unfortunately this does not seem to be the same the same `app` tree used for linting. I also tried to add similar code to the `treeForTestingSupport` but that didn't work either.

Looking through the ember-cli addon hooks and trying to fix this I think I have run into a wall. I cannot find a hook which will allow me to merge the `mirage/` directory with one of the existing trees and I can also not find any hooks for me to invoke `lintTree` for an additional tree.

## Conclusion
Personally, I really like that the `mirage/` directory has been moved to the root and I have talked to others who feel the same. I believe this is probably actually an issue in ember-cli using a different app tree for linting but wanted to raise awareness and gather feedback here before opening an issue on ember-cli.

rwjblue:
After moving the `mirage` tree out of one of the already linted trees (i.e. `app/mirage`) it is no longer being linted.  This ensures that the tree is linted in the same way as `addon` or `app` are normally linted.

Fixes #714.

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,13 +54,43 @@ module.exports = {
     return this._super.treeFor.apply(this, arguments);
   },
 
-  treeForApp: function(name) {
-    var originalAppTree = unwatchedTree(path.resolve(__dirname, 'app'));
+  _lintMirageTree: function(mirageTree) {
+    var lintedMirageTrees;
+    // _eachProjectAddonInvoke was added in ember-cli@2.5.0
+    // this conditional can be removed when we no longer support
+    // versions older than 2.5.0
+    if (this._eachProjectAddonInvoke) {
+      lintedMirageTrees = this._eachProjectAddonInvoke('lintTree', ['mirage', mirageTree]);
+    } else {
+      lintedMirageTrees = this.project.addons.map(function(addon) {
+        if (addon.lintTree) {
+          return addon.lintTree('mirage', mirageTree);
+        }
+      }).filter(Boolean);
+    }
+
+    var lintedMirage = mergeTrees(lintedMirageTrees, {
+      overwrite: true,
+      annotation: 'TreeMerger (mirage-lint)'
+    });
+
+    return new Funnel(lintedMirage, {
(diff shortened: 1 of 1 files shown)


rwjblue:
HAHA! Test failures are from files in https://github.com/samselikoff/ember-cli-mirage/tree/master/tests/dummy/mirage not passing linting.  I'll try to clean those up so this is green...

jimmay5469:
😜

jimmay5469:
I tested this and it is working great for me, thanks!

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/commit/fb24028e4ba8e0ec125bcb21e2de5e10fff6d9db closes, thanks dude!! You're the man

PR #769, opened 2016-06-09
Fix serializing embedded collection keys

arnodirlam:
I was trying to serialize a model that has a multiple hasMany() relationships with the standard ActiveModelSerializer plus `include` of these relationships with `embed: true`.

Although the regular fields are serialized fine as `underscore`, the embedded relationship keys were still in camelcase.

As it turned out, `camelize` was still hard-coded instead of `keyForRelationship`, so I fixed it and added a test. All existing tests still pass.


samselikoff:
Oh - awesome. This is great, thanks so much!

arnodirlam:
No problem, glad I could help! :)

PR #775, opened 2016-06-12
Remove styles for highlighting syntax errors

alecho:
@samselikoff I think this is the best course of action. By adjusting the styles we don't need to open code fences in markdown with the `text` type and we still get to keep _some_ highlighting for those blocks. The downside is that if a code block does have real errors we rely on only a keen eye.


samselikoff:
awesome - thanks a lot for tackling this!

PR #776, opened 2016-06-12
Update quickstart.md

0000marcell:


samselikoff:
There's other text around this that references `/api/authors` - can we add a 

```js
this.namespace = 'api';
```

to the top?

0000marcell:
Yes, i think its a good idea it also highlights the usage, there're other things in the quickstart that could be different imo, some pieces of code won't work if your just copy/paste in to your tests, a warning about routes, templates and models would be nice too, like: "assuming you have a author's template in your app..."  i think most users just creates a new ember-cli project when they're following the quickstart.

samselikoff:
I hear ya. For now, just for this PR, could you add a `this.namespace = 'api'` line above, and make sure the other lines that reference `/api` be updated as well?

0000marcell:
I didn't have the fork that i've made the change anymore, so i had to make another PR

PR #786, opened 2016-06-28
Added ms dependancy to package.json

rfbarber:
When setuping up the development env for ember-cli-mirage I had to install ms using npm


samselikoff:
I don't think we us this package. Do you remember why you needed to add it? Did you get an error?

acorncom:
@rfbarber We're aiming to clean up older issues/PRs, any updates on this?

samselikoff:
Closing for now, can always reopen if we need to. I don't believe we have an explicit dependency on this lib.

PR #791, opened 2016-06-30
Fix example usage of route’s this.serialize

backspace:
The arrow function can’t be used when calling `this.serialize` because `this` won’t be the route.


samselikoff:
Correct, thank you very much!

Wondering if there's a better API here, this has tripped up a few people. Or maybe there's a way to throw an error.

PR #795, opened 2016-07-04
Add findBy(query) method to db / schema

promisedlandt:
Hey Sam,

I want to add a finder method that takes a multiple attribute query (like `where`), but always returns 0 or 1 results (like `find`).
This method does exactly that.

My particular use case was for finding by slug instead of id in route handlers, like so:

````
let slug = request.params.id;

return schema.contacts.findBy({ slug: slug });
````

Please let me know if this is acceptable, and I'll write the docs for the site.

### DIFF
--- a/addon/orm/schema.js
+++ b/addon/orm/schema.js
@@ -85,6 +85,7 @@ export default class Schema {
       create: (attrs) => this.create(camelizedModelName, attrs),
       all: (attrs) => this.all(camelizedModelName, attrs),
       find: (attrs) => this.find(camelizedModelName, attrs),
+      findBy: (attrs) => this.findBy(camelizedModelName, attrs),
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: addon/orm/schema.js [hunk 1]
```diff
@@ -144,6 +145,23 @@ export default class Schema {
...
 
   /**
+   * @method findBy
+   * @param type
+   * @param attributeName
+   * @public
+   */
+  findBy(type, query) {
+    let collection = this._collectionForType(type);
+    let records = collection.where(query);
```
samselikoff: should this use `collection.findBy`?

samselikoff:
Nice! One comment, otherwise looks good

promisedlandt:
Yeah, good catch.

samselikoff:
Thanks!

Issue #753, opened 2016-05-31
Form Data


jbailey4:
How should form data within the FakeRequest be handled? Does mirage currently parse form data in any way? If not, I feel like parsing form data and allowing easy access to it would be a good addition.

Currently form data in the FakeRequest looks like the following:

`requestBody: "foo=bar&baz=someValue"`

samselikoff:
No convention here yet, you'd need to deserialize that string yourself using some JS (or maybe a helper function from jQuery). Would be a great addition - maybe we could update the [#normalizedRequestAttrs](https://github.com/samselikoff/ember-cli-mirage/blob/a595279657be150dff1e7a9cb50ee2041660865f/tests/integration/server/custom-function-handler-test.js#L46) helper method to return a JS object?

jbailey4:
I'm currently doing the deserialization myself. I did post a similar issue on the pretender repo, thinking maybe Form Data should be supported there. I could look into supporting this in mirage though.

I agree this addition should just parse the form data key/values and provide POJO.

I see [this method](https://github.com/samselikoff/ember-cli-mirage/blob/a595279657be150dff1e7a9cb50ee2041660865f/addon/route-handlers/base.js#L41-L73), maybe that would be a good place to look for the use of Form Data and start the parsing.

samselikoff:
Yep, that's exactly the spot. If you want to take a crack at it I'm happy to advise! I think a simple deserialization should suffice (POJO of key/value pairs).

jbailey4:
Looks like [the method](https://github.com/samselikoff/ember-cli-mirage/blob/a595279657be150dff1e7a9cb50ee2041660865f/addon/route-handlers/base.js#L41) I linked to above is very specific to jsonapi documents, do you think a separate helper method should added to function handlers for normalizing form data?

samselikoff:
Oh, you're right.

Add a new method [here]. Note that `this.request` will be set by the time the function is available in the handler. Let's name it `normalizedFormData`.

jbailey4:
> Add a new method [here].

I didn't get the link, did you mean [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/route-handlers/function.js#L37) in route-handlers/function.js? 😄

samselikoff:
Hehe, yes exactly 😛

samselikoff:
#796 closes

PR #796, opened 2016-07-04
adds helper method to normalize form data requests

jbailey4:
Adds a helper method to parse a request sent using `formData`. Currently just supports simple key/value url encoded strings, does not support `multipart/form-data`.

Related to https://github.com/samselikoff/ember-cli-mirage/issues/753

### DIFF
--- a/tests/integration/server/custom-function-handler-test.js
+++ b/tests/integration/server/custom-function-handler-test.js
@@ -43,7 +43,7 @@ test(`a POJA of models defaults to responding with an array of each model's attr
   });
 });
 
-test(`#normalizedRequestAttrs returns the an object with the primary resource's attrs and belongsTo keys camelized`, function(assert) {
+test(`#normalizedRequestAttrs returns an object with the primary resource's attrs and belongsTo keys camelized`, function(assert) {
   assert.expect(1);
   let done = assert.async();
   let { server } = this;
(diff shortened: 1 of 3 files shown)


samselikoff:
Nice work!

I'm wondering, is there any way we could make `normalizedRequestAttrs` smarter so we don't need to add a new method to the external API? Maybe it could check if the data associated with the request is a JSON string or looks like form-data, and delegate to the appropriate method call?

jbailey4:
@samselikoff Yeah I think that can be done. That was an option I thought of at first, but thought having the consumer explicitly choose the method to normalize their request was a safer option. 

Would checking the `content-type` on the request be a safe option? If the content type is `application/x-www-form-urlencoded` use the `_getAttrsForFormRequest`, otherwise if `application/json` or `application/vnd.api+json` use the current `_getAttrsForRequest`?

samselikoff:
Yes, that sounds like a great strategy

jbailey4:
@samselikoff See the latest commit, added the check for the `content-type` in `normalizedRequestAttrs` to determine which normalize helper method to use.

tests/integration/server/custom-function-handler-test.js
```diff
@@ -74,3 +75,29 @@ test(`#normalizedRequestAttrs returns the an object with the primary resource's
...
+      name: 'Sam Selikoff',
+      company: 'TED',
+      email: 'sam.selikoff@gmail.com'
+    }, '#normalizedRequestAttrs successfully returned the parsed x-www-form-urlencoded request body');
+
+    return {};
+  });
+
+  $.ajax({
+    method: 'POST',
```
samselikoff: ah - yes it does

samselikoff:
nice - thanks!

PR #798, opened 2016-07-04
Documentation for findBy

promisedlandt:
Pending #795, assuming the feature would be in v0.2.0-beta10


samselikoff:
Thank you very much 👍

PR #799, opened 2016-07-05
Adds Delete to Identity Manager

ryanpatrickcook:
When a record is removed, the identity manager currently retains the id with a value of true. When trying to revive the record, mirage gives an error like: `Attempting to use the ID 1, but it's already been used`. I'd like the ability to reinstate the object that was deleted (via undo or some action) and use the same id.

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -207,12 +207,13 @@ class DbCollection {
       let record = this._findRecord(target);
       let index = this._records.indexOf(record);
       this._records.splice(index, 1);
-
+      this.identityManager.delete(record.id);
     } else if (_isArray(target)) {
       records = this._findRecords(target);
       records.forEach((record) =>  {
         let index = this._records.indexOf(record);
         this._records.splice(index, 1);
+        this.identityManager.delete(record.id);
       });
 
     } else if (typeof target === 'object') {
@@ -220,6 +221,7 @@ class DbCollection {
       records.forEach((record) =>  {
         let index = this._records.indexOf(record);
         this._records.splice(index, 1);
+        this.identityManager.delete(record.id);
       });
     }
   }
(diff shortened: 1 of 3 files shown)


cibernox:
@samselikoff any feedback on this?

samselikoff:
Sorry, got busy & have a bit of a backlog of Mirage issues. Thx for the ping.

Could I ask the use case? I'm just curious because this is modeled after how "actual" dbs work.

ryanpatrickcook:
A lot of times we soft-delete records. So, we will use `destroyRecord` to fire off a request to the server to delete. The record will be marked as deleted in the database. If the user happened to delete the record by accident, they can undo the change. By undoing, we remove the deleted flag in the db and send the record back to the front-end with the same ID.

cibernox:
Actually my use case was different. If you delete a user and then you attempt to create a new one, it exploded with the error above. 

I was under the impression that somehow mirage was keeping a count of the last id assigned to records of a collection, but that somewhere else when you create a records the id was something like `collectionName.length`, so deleting one record meant that the length was diminishing but the lastId wasn't, so the next creation failed.

Maybe I have an entirely different issue, I'll try to check it.

samselikoff:
If this is true this is definitely a bug

samselikoff:
@ryanpatrickcook I think we should make this opt-in, because otherwise if you `server.create()` then `delete` then `server.create()` again, it would be id `1` both times, which would be confusing and wouldn't mimic most use cases. Right?

I'm more than happy to merge this in if we can make it opt-in... let's make it a setting on the `db` that you can set in `config.js`, `this.db.allowIdReuse = true;`. What do you think?

samselikoff:
Closing due to inactivity

PR #801, opened 2016-07-06
Pass schema to serializer

davewasmer:
Currently, the docs state that a good place to add metadata to a response is when overriding the `Serializer.serialize()` method. One common piece of metadata is the total number of records when making a filtered or paginated request.

Except the recommended `.serialize()` method is only given the response and the request as arguments, so it can't get a reference to the schema/server/db objects to get a total count.

This is one approach to solving that problem - just pass the schema into the serialize method.

---

However, I'm not sure this is the best long term solution. Metadata can be tricky - if I make a request that is both paginated **and** filtered, perhaps the metadata's `total: 42` indicates that there are a total of 42 records **that match that filter**, rather than 42 records of that type overall. 

If that's the case, mocking that behavior in Mirage would require calculating metadata in the route handler (where the filtering is happening), which this PR doesn't solve. But I figured I'd at least offer some kind of solution to start the conversation :wink:. One idea: perhaps you could return a Mirage.Response object that allows arbitrary metadata payloads but still triggers the serializer layer? Then that Response object could be read out in the serializer to populate the appropriate JSON payload structure.

### DIFF
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -24,7 +24,7 @@ export default class SerializerRegistry {
     if (this._isModelOrCollection(response)) {
       let serializer = this.serializerFor(response.modelName);
 
-      return serializer.serialize(response, request);
+      return serializer.serialize(response, request, this.schema);
     } else if (_isArray(response) && response.filter(this._isCollection).length) {
       return response.reduce((json, collection) => {
         let serializer = this.serializerFor(collection.modelName);


samselikoff:
Right - tricky problem. There's been some discussion about introducing APIs to pass metadata from route handlers to serializers.

Btw in case you weren't aware, you can also call [`this.serialize`](http://www.ember-cli-mirage.com/docs/v0.2.x/route-handlers/#serialize) from the route handler to get the `json`, and then add some final metadata right there in the route handler. It's definitely a kludge, though.

Not sure yet the best way to handle this. Will probably end up needing something like a `meta` object you can attach arbitrary data to that gets passed along all the way through the serializer. Going to leave this open for now so we can mull it over.

davewasmer:
> Not sure yet the best way to handle this. Will probably end up needing something like a meta object you can attach arbitrary data to that gets passed along all the way through the serializer. Going to leave this open for now so we can mull it over.

I'd suggest trying to avoid introducing more "top-level" concepts, like a `meta` object. We have a `Response` class which is already used when you need a response that isn't totally cookie cutter (i.e. a custom status code, headers, etc), seems like it would be straightforward to add arbitrary data to that as well.

It wouldn't have to be limited to `meta` data either - it could act as an escape valve for more complex edge cases (i.e. pass something along from the route handler by attaching it to the Response object, rather than calling the serializer manually)

samselikoff:
It's true - and I definitely prefer escape valves to new APIs when there's so much uncertainty.

For now I'd say call `this.serialize` in route handler and add more stuff there, until someone has time to work on this.

samselikoff:
Going to close this for now, until we can tackle the meta question head-on

allthesignals:
Here is a snippet of what I ended up doing for this:

```javascript
  this.get('/projects', function(schema, request) {
    let { queryParams: { page: offsetParam = 1 } } = request;
    let offset = offsetParam - 1;
    let begin = 0 + (30 * offset);

    const json = this.serialize(
      schema.projects.all().slice(begin, begin + 30)
    );

    json.meta = {
      total: schema.projects.all().length,
      pageTotal: 30,
      tiles: ['http://localhost:4200/test-data/tiles/96.mvt'],
      bounds: [[-73.9916082509177, 40.6824244259472],[-73.8847869770557, 40.8933091086328]],
    };

    return json;
  });
```

PR #803, opened 2016-07-07
Doc updates

jetpackpony:
Hi! I've noticed a couple of issues with the docs while working though them:

- need a belongsTo relation in a child model, otherwise the factory won't create objects
- wrong object is used in serializer example


samselikoff:
:+1: thank you very much!

PR #809, opened 2016-07-15
handle hasMany relationships in _getAttrsForRequest

Azdaroth:
Currently `hasMany` relationship are not handled in `_getAttrsForRequest`, only `belongsTo` are.  This PR makes both relationship types work.

### DIFF
--- a/addon/route-handlers/base.js
+++ b/addon/route-handlers/base.js
@@ -59,7 +59,9 @@ export default class BaseRouteHandler {
       Object.keys(json.data.relationships).forEach((key) => {
         let relationship = json.data.relationships[key];
 
-        if (!_isArray(relationship.data)) {
+        if (_isArray(relationship.data)) {
+          attrs[`${camelize(singularize(key))}Ids`] = relationship.data.map(obj => obj.id);
+        } else {
           attrs[`${camelize(key)}Id`] = relationship.data && relationship.data.id;
         }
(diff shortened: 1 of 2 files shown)


Azdaroth:
Wondering if it's the right change... What's the proper way currently to handle creating records with hasMany relationship? My initial problem was that  hasMany relationships assignment didn't work in acceptance tests and thought that adding the proper payload to attributes will do the job, but apparently it doesn't and there are some other places in codebase that would require change. Is it by design or lack of feature?

samselikoff:
Busy at the moment, but quickly: `attrs` within the context of Mirage usually means the `db` attributes, which means plain attributes and foreign keys. A `belongsTo` side of a relationship typically is the "owner" of the foreign key, and fks are typically what's persisted to the db. `hasMany` side usually is built up from the fks of the owner (belongsTo side). That's why `getAttrs` returns just the "persistable" attrs.

We might be moving away from the single-owner foreign key concept, so this might change (see [this post](http://www.ember-cli-mirage.com/blog/2016/01/03/0-2-0-beta.9-released/)). In the mean time, how you "persist" many-to-many is typically via a join table, which only persists the fks. So you don't need the hasMany side in getAttrs.

Azdaroth:
Ok, thanks, I get the idea now, this PR can probably be closed then.

PR #812, opened 2016-07-19
hasMany association cache issue

designblooz:
Not sure why `_cachedChildren` is set on the `ModelClass.prototype` level, but this causes hasMany model factories that were created from the a test to be part of the schema of the model throughout all the test.

To make sure every data is wiped clean after each test, added a method in `server.js` to go through every `ModelClass` to remove `_cachedChildren`.

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -368,15 +368,19 @@ export default class Server {
 
   _resetHasManyAssociations() {
     Object.keys(this.schema._registry).forEach((modelKey)=> {
-      if (this.schema._registry[modelKey].class) {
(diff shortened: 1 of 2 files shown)


samselikoff:
Probably no good reason :) Thanks for digging into this - any chance you could throw together a quick test?

designblooz:
@samselikoff @acorncom added test

kategengler:
Ran into this problem tonight while converting emberobserver over to JSONAPI. Setting the hasMany relationship through `update` _seems_ to be a workaround.

samselikoff:
I recall this being fixed elsewhere

PR #813, opened 2016-07-19
Pull third-party dependencies from npm instead of Bower

dfreeman:
I'm on a quest to remove Bower dependencies from our apps and addons where possible in the ultimate hope that once [Ember itself moves to npm](https://github.com/ember-cli/ember-cli/issues/4546) we'll be able stop using Bower entirely.

This changes switches over Faker, Pretender and friends over to come from npm, and eliminates the installation of those two packages via Bower in the default blueprint.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,9 +8,19 @@ var unwatchedTree = require('broccoli-unwatched-tree');
 module.exports = {
   name: 'ember-cli-mirage',
 
-  included: function included(app) {
-    this._super.included.apply(this, arguments);
+  options: {
+    nodeAssets: {
+      'route-recognizer': npmAsset({
+        path: 'dist/route-recognizer.js',
+        sourceMap: 'dist/route-recognizer.js.map'
+      }),
+      'fake-xml-http-request': npmAsset('fake_xml_http_request.js'),
+      'pretender': npmAsset('pretender.js'),
+      'faker': npmAsset('build/build/faker.js')
+    }
+  },
 
+  included: function included(app) {
     // see: https://github.com/ember-cli/ember-cli/issues/3718
     if (typeof app.import !== 'function' && app.app) {
       app = app.app;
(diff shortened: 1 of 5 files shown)


samselikoff:
@locks any feedback here?

locks:
👌 this is aligned with the on-going goal of making Ember an addon and removing Bower as a default dependency for Ember projects.

samselikoff:
Thanks!

Dhaulagiri:
I'm curious what options there are after this change for folks who want to use a specific version of Faker?  Is there a way for consumers of ember-cli-mirage to peg it at a specific version?  There are a couple of breaking changes in Faker 3.1 that we have been avoiding but now we are also wanting to take advantage of new changes in mirage as well so I'm not sure what we can do to reconcile things.

samselikoff:
@Dhaulagiri I agree this should absolutely still be possible for users to customize.

@locks can users override Mirage's default Faker version by supplying their own in their app's `package.json`? Or, is there some other way?

index.js
```diff
@@ -8,9 +8,19 @@ var unwatchedTree = require('broccoli-unwatched-tree');
 module.exports = {
   name: 'ember-cli-mirage',
 
-  included: function included(app) {
-    this._super.included.apply(this, arguments);
+  options: {
+    nodeAssets: {
```
josemarluedke: Hey @dfreeman, I'm wondering where this option `nodeAssets  is been handled. Is it something that Ember CLI does for us? Could you point me in the right direction to learn more about it? Thanks.

dfreeman: @josemarluedke It's handled by https://emberobserver.com/addons/ember-cli-node-assets, which basically just provides some sugar for funneling files from `node_modules` into your build to either be `app.import`ed or land in `public`.

samselikoff:
@locks this is causing more issues for people, users will still need to be able to customize their versions of pretender and faker.

Is it possible to do this via package.json?

locks:
@samselikoff I thought that if you specify the dependency in `package.json` that it would be enough. Hm, maybe ask someone from Ember CLI to confirm/suggest a better solution.

samselikoff:
@locks yep now that we float Faker in mirage's package.json, host apps can override in their own.

Thanks!

PR #814, opened 2016-07-20
add :resource helper for handling generic CRUD

Azdaroth:
Shorthands are super useful, but for very generic CRUD it might be a bit tedious to define a route handler for every endpoint, so I thought that Rails-like `resource` helper would help with that.

### DIFF
--- a/tests/integration/http-verbs-test.js
+++ b/tests/integration/http-verbs-test.js
@@ -107,7 +107,7 @@ test('mirage responds to resource', function(assert) {
   assert.expect(0);
   let done = assert.async();
 
-  this.server.resource('/contacts');
+  this.server.resource('contacts');
 
   $.ajax({
     method: 'GET',
--- a/tests/integration/http-verbs-test.js
+++ b/tests/integration/http-verbs-test.js
@@ -1,10 +1,14 @@
 import {module, test} from 'qunit';
 import Server from 'ember-cli-mirage/server';
+import { Model } from 'ember-cli-mirage';
 
 module('Integration | HTTP Verbs', {
   beforeEach() {
     this.server = new Server({
-      environment: 'development'
+      environment: 'development',
+      models: {
+        contact: Model
+      }
     });
     this.server.timing = 0;
     this.server.logging = false;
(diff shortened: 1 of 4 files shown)


Azdaroth:
@samselikoff What do you think about this feature?

tests/integration/http-verbs-test.js
```diff
@@ -99,6 +103,20 @@ test('mirage responds to patch', function(assert) {
   });
 });
 
+test('mirage responds to resource', function(assert) {
+  assert.expect(0);
+  let done = assert.async();
+
+  this.server.resource('/contacts');
```
samselikoff: I think the api should be `this.resource('contacts')` - you're declaring a contact, not defining a url

samselikoff:
pretty neat!

In general I think your mirage server should be conservative it what it defines. My problem with this is that you'd use it and get 6 endpoints, and maybe you're actual server only defines 3.

Thinking through some different apis, we could borrow something from rails like

```js
this.resource('blog-posts'); // all operations
this.resource('contacts', { only: ['index', 'show', 'create', 'update', 'delete'] });
this.resource('comments', { except: ['delete'] });
```

I'd like to sit and hold off on this one for a bit, just because it adds more abstraction to mirage. I'll be thinking about it.

tests/integration/http-verbs-test.js
Azdaroth: I was actually thinking if it's better to be more expressive (`this.resource('contacts')`) or keep the symmetry with other shorthands (`this.post(/contacts)`). For me both would be fine here,  no problem to change the api if you prefer `this.resource('contacts')` syntax.

Azdaroth:
Sure thing, let me know what you final thoughts are, I would be happy to extend this API with `only` and `except` if needed.

Azdaroth:
@samselikoff I wanted  to give this `only` and `except` API a try, so I quickly prototyped something working which looks pretty much done. Let me know what you think.

samselikoff:
Nice, sorry about the delay. This is a fantastic PR! Excellent test coverage.

Could you look at the test failures? https://travis-ci.org/samselikoff/ember-cli-mirage/jobs/151434254#L1306

I'm happy to merge this once we get things passing.

Azdaroth:
Great to hear that! 

Sure thing, I'll take care of that tomorrow I guess.

Azdaroth:
@samselikoff Specs seem to be happy now ;)

samselikoff:
Incredible work. Thank you very much!

PR #816, opened 2016-07-21
Has many relationship should create *_ids property

alvincrespo:
### Problem

When a  belongsTo/hasMany relationship exists, if embed is false the resulting response should include the ids of the many relationship. However this is not the case:

### Example

**Given the following**

```
  let game = this.schema.games.create({ name: 'Legend of Zelda' });

  this.schema.players.create({ game });
  this.schema.players.create({ game });

  let games = this.schema.games.all();
  let result = this.registry.serialize(games);
```

**The expected JSON response should be**

```
{
  "games": [
    {
      "id": "1",
      "name": "Legend of Zelda",
      "player_ids": [
        1,
        2
      ]
    }
  ]
}
```

### DIFF
--- a/tests/integration/serializers/active-model-serializer-test.js
+++ b/tests/integration/serializers/active-model-serializer-test.js
@@ -23,6 +23,12 @@ module('Integration | Serializer | ActiveModelSerializer', {
       }),
       contactInfo: Model.extend({
         user: belongsTo()
+      }),
+      game: Model.extend({
+        players: hasMany()
+      }),
+      player: Model.extend({
+        game: belongsTo()
       })
     });
 
@@ -48,6 +54,9 @@ module('Integration | Serializer | ActiveModelSerializer', {
         attrs: ['id', 'name'],
         include: ['ContactInfos'],
         embed: true
+      }),
+      game: ActiveModelSerializer.extend({
+        attrs: ['id', 'name']
       })
     });
   },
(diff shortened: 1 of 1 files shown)


alvincrespo:
I haven't gotten to the solution yet, but I thought I'd start out by including the failing test. I'm still going through the source code trying to figure out where the issue might be.

samselikoff:
I'm working on a refactor of the Serializer layer over in #781 that I think will take care of this. We'll add a new option:

     Serializer.prototype.serializeIds = 'included'; // can be 'included', 'always', or 'never'

acorncom:
@alvincrespo #781 has been merged, can we close this PR at this point?

samselikoff:
I think my rework didn't catch this, I ended up overriding too much of the
base to get this for free. Won't be hard to add.

samselikoff:
However this did make it in for Serializer/ActiveModelSerializer/RestSerializer, so I think we can close this for now.

Here's the test: https://github.com/samselikoff/ember-cli-mirage/blob/e92b9391a209b11a7c032699ed8662cde6a48f50/tests/integration/serializers/base/serialize-ids-test.js#L22

For backwards compatibility you want to opt-in to this feature like this:

```js
  let ApplicationSerializer = Serializer.extend({
    serializeIds: 'always'
  });
```

PR #817, opened 2016-07-22
Introducing utils/normalize-name.js to concentrate the naming logic

### LINKED ISSUE #793: Collection name not respecting ember-inflector rules
Hello there,

I've configured the following inflector rule on my ember app (doing so by following the ember guides):

```javascript
import Inflector from 'ember-inflector';

const inflector = Inflector.inflector;

inflector.irregular('tipo-documento', 'tipos-documento');

// Meet Ember Inspector's expectation of an export
export default {};
```

on my `config.js` I have:
```javascript
this.get('/sigoc/api/v1/tipos-documento', (schema, request) => {
  return schema.tiposDocumento.all();
});
```

which is erroring out with the following message:
`vendor.js:127893 Mirage: Your handler for the url http://localhost:8888/sigoc/api/v1/tipos-documento threw an error: Cannot read property 'all' of undefined` .

This is happening because inside `schema` there is a `tipoDocumentos` collection, which is the wrong inflection, as if mirage was completely ignoring my ember configuration.

Is this the intended behaviour? If so, how do I pass my inflection rules to mirage?

escobera:
Fixes #793

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -1,5 +1,6 @@
 import _isArray from 'lodash/lang/isArray';
-import { camelize, singularize, pluralize } from 'ember-cli-mirage/utils/inflector';
+import { singularize } from 'ember-cli-mirage/utils/inflector';
+import { toDbCollectionName } from 'ember-cli-mirage/utils/normalize-name';
 import BaseRouteHandler from '../base';
 
 export default class BaseShorthandRouteHandler extends BaseRouteHandler {
@@ -14,12 +15,12 @@ export default class BaseShorthandRouteHandler extends BaseRouteHandler {
 
     let type = _isArray(shorthand) ? 'array' : typeof shorthand;
     if (type === 'string') {
-      let modelClass = this.schema[pluralize(camelize(singularize(shorthand)))];
+      let modelClass = this.schema[toDbCollectionName(singularize(shorthand))];
       this.handle = (request) => {
         return this.handleStringShorthand(request, modelClass);
       };
     } else if (type === 'array') {
-      let modelClasses = shorthand.map((modelName) => this.schema[pluralize(camelize(singularize(modelName)))]);
+      let modelClasses = shorthand.map((modelName) => this.schema[toDbCollectionName(singularize(modelName))]);
       this.handle = (request) => {
         return this.handleArrayShorthand(request, modelClasses);
       };
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -1,6 +1,5 @@
 import _isArray from 'lodash/lang/isArray';
-import { singularize } from 'ember-cli-mirage/utils/inflector';
-import { toDbCollectionName } from 'ember-cli-mirage/utils/normalize-name';
+import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
 import BaseRouteHandler from '../base';
 
 export default class BaseShorthandRouteHandler extends BaseRouteHandler {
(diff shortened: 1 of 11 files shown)


locks:
The chosen name is a bit of a mouthful. Should we consider a shorter name, or use ES6's `as` import alias if desired to shorten it?

escobera:
I can't think of another name (actually that one was suggested by @samselikoff on #793 ), but I'm open to suggestions.

addon/route-handlers/shorthands/base.js
```diff
@@ -14,12 +15,12 @@ export default class BaseShorthandRouteHandler extends BaseRouteHandler {
 
     let type = _isArray(shorthand) ? 'array' : typeof shorthand;
     if (type === 'string') {
-      let modelClass = this.schema[pluralize(camelize(singularize(shorthand)))];
+      let modelClass = this.schema[toDbCollectionName(singularize(shorthand))];
```
samselikoff: think we can remove `singularize` here?

addon/route-handlers/shorthands/base.js
```diff
@@ -14,12 +15,12 @@ export default class BaseShorthandRouteHandler extends BaseRouteHandler {
...
       this.handle = (request) => {
         return this.handleStringShorthand(request, modelClass);
       };
     } else if (type === 'array') {
-      let modelClasses = shorthand.map((modelName) => this.schema[pluralize(camelize(singularize(modelName)))]);
+      let modelClasses = shorthand.map((modelName) => this.schema[toDbCollectionName(singularize(modelName))]);
```
samselikoff: and here

addon/serializer.js
```diff
@@ -82,7 +82,7 @@ class Serializer {
    * @public
    */
   keyForCollection(modelName) {
-    return pluralize(this.keyForModel(modelName));
+    return toDbCollectionName(modelName);
```
samselikoff: this helper should only be used when accessing a schema or db collection; I'd rather keep this the basic logic so people can see "oh, it's just munging the key used in the payload. I can do whatever I want here"

addon/serializer.js
```diff
@@ -118,7 +118,7 @@ class Serializer {
    * @public
    */
   keyForRelationship(modelName) {
-    return _compose(camelize, pluralize)(modelName);
+    return toDbCollectionName(modelName);
```
samselikoff: same ^

addon/serializers/rest-serializer.js
```diff
@@ -12,7 +13,7 @@ export default ActiveModelSerializer.extend({
   },
 
   keyForRelationship(type) {
-    return pluralize(camelize(type));
+    return toDbCollectionName(type);
```
samselikoff: same

addon/server.js
```diff
@@ -199,7 +200,7 @@ export default class Server {
 
     // Create a collection for each factory
     _keys(factoryMap).forEach(type => {
-      let collectionName = this.schema ? pluralize(camelize(type)) : pluralize(type);
+      let collectionName = this.schema ? toDbCollectionName(type) : pluralize(type);
```
samselikoff: i think these are the same now? originally schema collections were singularized and db plural, but not anymore

samselikoff:
great work! Let's rename to `toCollectionName` since both schema collections and db collections are the same (pluralized and camelized). Originally they weren't (schema were singular) but we changed that, so I think it would simplify to go with the same util for both cases.

escobera:
@samselikoff sorry for the delay, was on vacation.

I made the changes as you suggested, tests still pass and my app still works. So I think it's all good =)

samselikoff:
woot!

PR #823, opened 2016-07-25
JSONAPIAdapter shorthand routes need to include empty relationship values by default

### LINKED ISSUE #821: JSONAPIAdapter shorthand routes need to include empty relationship values by default
When setting up the shorthand routes, I think that the serializer for the JSONAPIAdapter needs to include blank (`null` or `[]`) values for all relationships configured within the Mirage model. I'm guessing this would require a little reflection of how relationships are setup within the models.

I may try hacking on this a little to see if I can add it in... 😮 

As a little background, here's a scenario where this has failed me in writing some acceptance tests.

I've set up a `page` model like this:

```javascript

export default Model.extend({
  draft: belongsTo()
});
```

Then configured the shorthand route in `config.js`:

```javascript
this.get('/pages');
```

In my acceptance test, I seed the database like so:

```
let draft = server.create('draft', { user });
server.create('page', { draft });
```

The Ember app then loads in the page, which has this payload when running `server.logging = true` in the test. All is good so far:

```json
{
  "data": [
    {
      "type": "pages",
      "id": "1",
      "attributes": { … },
      "relationships": {
        "draft": {
          "data": {
            "type": "drafts",
            "id": "1"
          }
        }
      }
    }
  ],
  "included": [ … ]
}
```

Another endpoint on the Mirage server nullifies the `page.draftId` and removes all drafts:

```javascript
page.draftId = null;
page.save();

schema.db.drafts.remove();
```

I then call the same `/pages` endpoint from before again to reload the data, and I get this payload, which in some ways is "correct" because the draft is "gone":

```json
{
  "data": [
    {
      "type": "pages",
      "id": "1",
      "attributes": { … }
    }
  ]
}
```

But in order for Ember to remove that associated draft from its store, I think it needs to see a `null` `draft` relationship, like so:

```json
{
  "data": [
    {
      "type": "pages",
      "id": "1",
      "attributes": { … },
      "relationships": {
        "draft": {
          "data": null
        }
      }
    }
  ]
}
```

chrisdpeters:
Looks like the tests were setup sufficiently to cover this. Did modify a couple tests to reflect the new behavior.

Closes #821

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -127,11 +127,11 @@ class JsonApiSerializer {
       let relationship = model[camelizedType];
       let relationshipKey = this.keyForRelationship(camelizedType);
 
-      if (isCollection(relationship)) {
-        if (!obj.relationships) {
-          obj.relationships = {};
-        }
+      if (!obj.relationships) {
+        obj.relationships = {};
+      }
 
+      if (isCollection(relationship)) {
         obj.relationships[relationshipKey] = {
           data: relationship.models.map(model => {
             return {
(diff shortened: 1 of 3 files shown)


samselikoff:
Awesome work - thanks!

PR #829, opened 2016-08-02
Remove unneeded argument

harianus:
`i` is defined but never used


locks:
Thanks for your contribution @harianus! Every bit counts :)

PR #836, opened 2016-08-08
Custom scenarios

ivanreyne:
Hi! First of all thanks a lot for the mirage server: I haven't been using it for long but it has already saved me many hours of work and headaches, so in my name and my dog's thanks for this extra time available in our lives! :-)

I have done a first take to having custom scenarios and different fixtures to initialise them (or to use on the tests). I have done it based on the requirements I see arising for a big project we are implementing at my work place where we will have a full team of people working on the application where we will need different data for the different features each developer is going to be using and also lots of different fixtures for out test cases. I hope that it will also be useful (or at least a good starting point) to other people. I also tried to do the minimum amount of changes possible, just in case, I didn't want to break anything without realising. Of course, the tests pass :-)

For the fixtures the idea behind the implementation is to load the them on demand rather than having everything loaded at start-up. Still, everything is backwards compatible and existing code should work in the same way. All the fixtures MUST reside inside mirage/fixtures (as now) but you can have folders and sub-folders inside! The data files MUST have the name of the model (as now). The way it works is that you can specify models with their relative path or you can even specify the name of the folder (ending with a slash) for all the model data files inside to be used.
So, you can have things like this (from the upgraded tests):
- mirage/fixtures/
- blog-posts.js
  -- case-00/
  --- blog-posts.js
  --- specific-00/
  ---- blog-posts.js
  --- specific-01/
  ---- blog-posts.js
  -- case-01/
  --- blog-posts.js
  Then on the scenario or test case you can do:
  server.loadFixtures('', 'case-01/', 'case-00/');
  server.loadFixtures('', 'case-00/', 'case-01/blog-posts');
  server.loadFixtures('', 'case-01/', 'case-00/blog-posts', 'case-00/specific-00/');
  The last (right-most) model specified is the one used when on different folders you have different data for the same model.

For the scenarios it is a bit more simpler (by the way, I didn't know how to create the acceptance test for them). The different scenario specifications MUST reside inside mirage/scenarios (as now) and there is a new key available on the mirage config.js file to specify which one we load.
So it will look something like this:
- mirage/scenarios/
  -- default.js
  -- ticket-0000.js
  -- ticket-0001.js
  -- specific-data-for-functionality-XXX.js
  In config.js:
  ...
  this.timing = 0;      // delay for each request, automatically set to 0 during testing
  this.scenario = 'ticket-0001.js';
  ...
  And, of course, inside the specific scenario file you can either use factories and/or the new loadFixture with specific data for each of the cases.

I hope you find this useful enough to add on the release at some point. Any questions and corrections please feel free to ask.

Again, many thanks for this great software little thingy! :-)

Regards,

Ivan.


samselikoff:
Interesting approach, thanks for the PR! I'd like to think more about the best approach to this - for now, can you use imports and the `server.loadFixtures` API in the main `scenario` file to share data?

Also, did you come up with a way to load different files?

ivanreyne:
Yes to both questions ... actually now I realise that I have had a problem with the formatting of the initial comment where I was trying to give an example of how things would look like and work. Let's see if this time I first read the markdown help page and I manage to put the examples with the right format :blush:

There are three separate aspects to use the custom scenarios and having different fixtures: the setup, the developers use case and the testers use case.

## Setup
For the example I am assuming an "ambitious" web application, something like a project manager, where we have different people, activities, projects, etc. Also let's imagine that we are working with user stories, a bug tracking system and that we have separate development and test teams. The folder structure for the fixtures could be something like this:
* mirage/fixtures
  * activity-types.js (global types for activities)
  * project-types.js (global types for projects)
  * user-types (global types for users)
  * users.js (default list of users we use in the development and test)
  * stories (folder to organise the data that we need to test specific user stories)
    * project-total-time-is-sum-of-activities-time
      * activities.js (a set of activities with known values for the story test)
      * projects.js (a set of projects with known values for the story test)
    * user-with-more-than-one-project-is-flagged-to-admin
      * projects.js (idem)
    * users-with-weak-password-cannot-access-system
      * users.js (modified users.js from main folder, some with weak passwords, to check the story)
  * bugs (folder to organise the data to replicate bugs and to make sure they don't happen again)
    * 0000-login-failure
      * users.js (snapshot of user table when the problem occurred)
    * 0001-activity-showing-wrong-time-format
      * case-00
        * activities.js (snapshot of activity table when the problem occurred)
        * projects.js (snapshot of project table when the problem occurred as we need to relationships with activities to match)
      * case-01
        * activities.js (crafted data to cover additional cases, relationships with projects match what comes by default from the project factory)
  * development (this folder is under gitignore and is where the different developers might have specific fixtures for the work they are doing. If that is later needed for a test then it can be put on a specific folder under "stories" or "bugs")
    * boring-new-activity-feature
      * activities.js (custom data for the new feature I am working on)
    * bug-0010
      * users.js (data to replicate the condition that triggers the problem)
      * projects.js (idem)
      * activities.js (idem)

## Developers Use Case
For the developer's work they will primarily use custom scenarios as a way to package the inclusion of the different fixtures combined with the data generated using the model factories.

So the **developers** would have something like this in their mirage scenarios:
* mirage/scenarios
  * default.js (the main development scenario that should not be touched)
  * my-work-on-user-story-000.js (the setup I need to work on a user story, on gitignore)
  * my-work-on-bug-0010.js (the setup I need to replicate bug 0010, on gitignore)

The default.js file will look like:
```javascript
    // Loading the tables that have constant content like the types.
    server.loadFixtures();

    // Creating some random data for development.
    server.createList('activity', 100);
    server.createList('project', 10);
}
```
The my-work-on-user-story-000.js file will look like:
```javascript
    // Loading the tables that have constant content like the types.
    server.loadFixtures();
    server.loadFixtures('development/boring-new-activity-feature/');

    // Creating some random data for development.
    server.createList('project', 10);
}
```
The my-work-on-bug-0010.js file will look like:
```javascript
    // Loading the tables that have constant content like the types.
    server.loadFixtures();
    server.loadFixtures('development/bug-0010/');
}
```
Note that by the way the fixtures get loaded (the last one wins) the my-work-on-bug-0010.js scenario will have the users from development/bug-0010 folder rather than the main fixture folder.
We could also write the loadFixtures call like this: `server.loadFixtures('', 'development/bug-0010/');` (even though having the empty string to load the default fixture folder looks a bit strange to me).
Note also, in the development use case, it is the developer's work to know in what feature he is working on at any time and updating the mirage/config.js accordingly to load the scenario he needs to do the job: `this.scenario = 'default';` or `this.scenario = 'my-work-on-user-story-000';`, etc. And they might have to consider if the mirage/config.js file should be or not under revision control to avoid a sloppy developer submitting it to the git repository with his own custom scenario (even thought that is a problem out of this scope).

## Testers Use Case
The testers will use directly the server.loadFixtures on their acceptance and integration tests. By the way test works they will not be using scenarios as such.

Using the above setup if we have an acceptance test where we simply go through all routes available on the application to check for the implementation of the login with ember-simple-auth we do not need any special fixtures apart from the standard list of users that gets loaded when including the default fixture folder:
```javascript
/* global server */
/* global visit, fillIn, click, andThen, currentURL */
/* jshint loopfunc: true */
import { test } from 'qunit';
import moduleForAcceptance from 'ambitious-app/tests/helpers/module-for-acceptance';

moduleForAcceptance('Acceptance | routes check login protection');

const appRoutes = [
    '/activity/1',
    '/activity-new',
    '/activities',
    '/user/1',
    '/users',
    '/project/1',
    '/projects'
];

for(let i = 0; i < appRoutes.length; i ++) {
    test(`visiting all routes: ${appRoutes[i]}`, function(assert) {
        server.loadFixtures();
        visit(appRoutes[i]);

        andThen(function() {
            assert.equal(currentURL(), '/login', 'Route has login protection');
        });
    });

    test(`visiting all routes and log in: ${appRoutes[i]}`, function(assert) {
        server.loadFixtures();
        server.createList('activity', 2);
        server.createList('project', 2);

        visit(appRoutes[i]);
        fillIn('input#username', 'steven-seagal');
        fillIn('input#password', 'p');
        click('button#login');

        andThen(function() {
            assert.equal(currentURL(), appRoutes[i], 'Route redirects properly after login');
        });
    });
}
```
For the project-total-time-is-sum-of-activities-time user story we would have an acceptance test setup like this:
```javascript
...
test(`making sure the project total time is the sum of all the activities`, function(assert) {
    server.loadFixtures(); // loading the standard fixtures
    server.loadFixtures('stories/project-total-time-is-sum-of-activities-time/'); // loading the specific fixtures for this test
...
```
For the bug 0000-login-failure we would have an acceptance/integration test setup like this:
```javascript
...
test(`login failure bug because of Scandinavian characters on password`, function(assert) {
    server.loadFixtures(); // loading the standard fixtures
    server.loadFixtures('bugs/0000-login-failure/'); // loading the specific fixtures for this test
...
```
For the bug 0001-activity-showing-wrong-time-format we would have an acceptance/integration test setup like this:
```javascript
...
test(`activity showing wrong time format because of the locale, original db dump`, function(assert) {
    server.loadFixtures(); // loading the standard fixtures
    server.loadFixtures('bugs/0001-activity-showing-wrong-time-format/case-00/'); // loading the specific fixtures for this test
...
}
test(`activity showing wrong time format because of the locale, test with Spanish locale`, function(assert) {
    server.loadFixtures(); // loading the standard fixtures
    server.loadFixtures('bugs/0001-activity-showing-wrong-time-format/case-01/'); // loading the specific fixtures for this test
...
}
```
As the tests are all run on a clean instance of the application you can have them all run under the test'em framework that Ember uses without anything getting mixed up at all.

## Final Considerations
You have the ability to load different fixture folders (and files) and the last one loaded wins. This means that you could do something like that on your development or test:
```javascript
...
    server.loadFixtures(); // loading the standard fixtures
    server.loadFixtures('bugs/0001-activity-showing-wrong-time-format/case-00/', 'stories/project-total-time-is-sum-of-activities-time/activities');
...
```
In this case you end up with the following fixtures on the mirage DB:
* activity-types.js (from mirage/fixtures)
* project-types.js (from mirage/fixtures)
* user-types (from mirage/fixtures)
* users.js (from mirage/fixtures)
* activities.js (from stories/project-total-time-is-sum-of-activities-time)
* projects.js (from bugs/0001-activity-showing-wrong-time-format/case-00)

It might not look useful in this particular case but it gives you a lot more flexibility when setting up the data needed for your test cases or development scenarios.

I hope I managed to explain properly how this implementation of custom scenarios works and how each role (developer and tester) should use them. Hopefully it adapts to the work flow structure of most development teams regardless of their methodology. I think it is not a big change in terms of the code they have to generate compared to now. It is a bigger change in terms of the setup of the fixtures folder structure and the gitignore file to avoid developers pushing their fixture modifications to the repository (but probably much better than now that there is only one big shared folder of fixtures for all people!).
If you have more questions/doubts please feel free to ask ... now I can format the answers properly! :smile:
Also, if you decide to use the PR, feel free to use this text as help page and/or modify as you feel fit.

Regards,

Ivan.

ivanreyne:
By the way, probably it is not clear from the examples I wrote but you do NOT need to call ```loadFixtures()``` with no parameters. You could have a test case (the same applies for a scenario) set up like this:
```javascript
...
test(`test description`, function(assert) {
    server.loadFixtures('dumpDir00/');
    server.createList('modelXXX', 5);
    server.createList('modelYYY', 15);
    ...
}
test(`test description`, function(assert) {
    server.loadFixtures('dumpDirBug001');
    ...
}
test(`test description`, function(assert) {
    server.loadFixtures('dumpDirBug001/specialCase00/');
    server.createList('modelXXX', 5);
    ...
}
...
```

samselikoff:
Heya, I just wanted to let you know that this is a very large issue, and I'm not going to be able to look at it for a while. I hope you can solve the problem within your own app for now, until I have time for this or you can recruit some other folks to discuss what you've proposed here.

samselikoff:
I'm going to close this for now. Next time we should probably do more of an RFC-style process for such a big change, so we can work through the design with the community.

Let me know if this is still pressing and we can start a discussion about it in an issue.

Issue #826, opened 2016-07-30
Problem with acceptance tests in ember 2.7


spruce:
Hey,
first of all thanks for this great tool to make testing easier. 

After I did all the transitions from ember 2.5 to 2.7 (ember, data, cli) which included updating some more deps. I noticed that the tests did work in the browser under /tests (I know this is discouraged to use) but did hang for like 30 secs in `ember test`. After going to the slack channel I found out that the problem was that my namespace was not starting with a slash. After changing it to start with a slash everything worked as expected.

This should be a heads-up for others running into the same problem and also to find a permanent solution for this.

samselikoff:
Would love some help with this, even if there's a quick bandaid solution! People keep running into it.

ohcibi:
I run into the same. A few things worth mentioning:

* If you did not set a namespace at all, the problem also happens
* If so just setting the namespace to `'/'` and leave the rest alone does not seem to work
* If you upgrade the config on using a namespace then don't forget to update the routes you may have overridden in your tests

givanse:
A good start would be to update this line from:

> make this `api`, for example,

to:

> make this `/api`, for example

acorncom:
Looks like we've got PRs that handle this, closing for now.  If I missed something please comment and we can re-open this issue

jrjohnson:
Yes, it looks like there is even a backport PR  for it, but I don't think that has been released yet.  We went ahead and over road the method until that gets released:
https://github.com/ilios/frontend/blob/master/app/mirage/config.js#L10-L39

PR #837, opened 2016-08-08
Allow a namespace of / to be used

jrjohnson:
If / is passed as the namespace it shouldn't end up as //

This is my attempt at a workaround for #826 it allows me to send `/` as my namespace and then tests work again.  I don't necessarily understand what all of these options do and what the default behavior should be, so please don't hesitate to tell me this is a mess.


samselikoff:
I think this is good - great test coverage. Thank you!

Issue #804, opened 2016-07-08
OPTIONS requests mocking


riklaunim:
Is there a way to mock OPTIONS requests? I use those to get model metadata used for form generation (field choices).

samselikoff:
Hi, I'm actually not sure off the top of my head. Assuming Pretender supports it, it should be relatively straightforward to add.

- Add `'options'` to [this line](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/server.js#L297)
- Add a very simple test to [this module](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/http-verbs-test.js)

Think you'd be up for giving this a shot? I'd be happy to answer any questions!

riklaunim:
Yes, I can check that out ;)

riklaunim:
I've added options:
* https://github.com/riklaunim/ember-cli-mirage/commit/14523a7d52163e9f6dd44acb4fd989f766e40bd6
* https://github.com/riklaunim/ember-cli-mirage/commit/e745afca41d07586c9c4b77462b4f303d9bdfd1d

And I get:
    TypeError: this.pretender[verb] is not a function

When trying to use it. Pretender has it: https://github.com/pretenderjs/pretender/blob/7e2d2dcef8fe6fe0dc00807a8da2ca337663a06d
/pretender.js#L53

So it's bit unexpected.

samselikoff:
If you open a PR with your changes we can discuss the code more easily there

riklaunim:
ok :)

acorncom:
@riklaunim Any additional info here?

riklaunim:
New pretender needed.

acorncom:
@riklaunim Newer than the 1.2 merged in in this PR? https://github.com/samselikoff/ember-cli-mirage/pull/881

riklaunim:
Hm.. it needs 1.3 that was released today :) Will update this PR today too.

acorncom:
We're tidying up issues, so am going to close this issue for now.  If it's still an issue, happy to reopen

atremblay-seniorlink:
Any updates on this issue? I'm getting ERR_CONNECTION_REFUSED responses during my Ember Data CORS preflight check. It would be great to have Mirage support the options method (unless it already does and I'm missing something).

samselikoff:
This would be nice to have, if someone has some time to spare I'd love to merge it into Mirage.

[...]

PR #841, opened 2016-08-13
[#804] OPTIONS request mocking support

riklaunim:
Initial commit for OPTIONS request mocking. Does not work yet, throwing:

```
TypeError: this.pretender[verb] is not a function
```

When options is trying to get mocked.


addon/route-handler.js
```diff
@@ -41,6 +41,8 @@ function createHandler({ verb, schema, serializerOrRegistry, path, rawHandler, o
     handler = new DeleteShorthandHandler(...args);
   } else if (verb === 'head') {
     handler = new HeadShorthandHandler(...args);
+  } else if (verb === 'options') {
+    handler = new GetShorthandHandler(...args);
```
riklaunim: This will rather require it's own handler.

samselikoff:
Looks like [Pretender supports OPTIONS](https://github.com/pretenderjs/pretender/blob/7e2d2dcef8fe6fe0dc00807a8da2ca337663a06d/test/registry_test.js#L15).

Regarding the shorthand handler, I agree it should have its own handler. Should there even be a shorthand for OPTIONS? What would the behavior be?

Perhaps we can start with just adding the verb, and letting users use the function handler, e.g.

```js
this.options('/some/url', () => {
  // normal function handler response
});
```

ksnyde:
@samselikoff I developed a PR but it looks too similar to this one to add it to the repo. the sad part though is that neither seems to address allowing a passthrough of a OPTIONS & POST pairing.

ksnyde:
Ok, my last comment is a bit misleading. I had in earlier testing removed the "options" from the my `config.js`. Now that it is correct to read:

    this.passthrough('https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyPassword', ['post', 'head', 'options']);

I'm back to my state from yesterday where it is fails on the "options" verb of the pretender object: 
https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/server.js#L179
image

ksnyde:
Sorry to be so verbose but ... another update ...

I switched the URL to a shortened version with wildcard characters:

    this.passthrough('https://www.googleapis.com/**', ['post', 'head', 'options']);

this DOES work with this PR. 👍

ksnyde:
Any chance we could merge this in? even with the fully qualified URL it is an improvement I think.

riklaunim:
I need to fix the style validation errors and see what else is failing. Either way you can use a git commit as the version in package.json if you just want to use it now :)

```
git+https://github.com/riklaunim/ember-cli-mirage.git#abae084fe2046648a7b26b31e49df2841c9f90fb
```
should do the trick.

ksnyde:
yup @riklaunim that's what I'm doing for now. glad you got to this 18 days before me. :) Thanks for the PR.

One last note, working off your code base I noticed that even though my shortened URL works the passthrough call will fail when explicitly asking for 'options'.

ksnyde:
@riklaunim in fact I _can't_ use yours yet as it fails on the same line as it was with base ember-cli-mirage. I had added one piece of defensive code into my clone of your PR:

    verbs.forEach(verb => {
      paths.forEach(path => {
        let fullPath = this._getFullPath(path);
        if(this.pretender[verb]) {
          this.pretender[verb](fullPath, this.pretender.passthrough);
        } else {
          console.warn(`Call to a non-existant pretender method: "${verb}". Pretender is: `, this.pretender);
        }
      });
    });

Not sure whether this is worth having in the final code once everything is sorted but it does get this to working.

riklaunim:
Well, I was wondering why Pretender doesn't have my options and other changes visible. Turns out the installed dependency isn't the one that is being used (that's from a docker container):

```
./bower_components/pretender/pretender.js
./node_modules/ember-cli-mirage/node_modules/pretender/pretender.js
```
node_modules/ember-cli-mirage/node_modules/pretender/pretender.js is used and not the version I install with my project as ./bower_components/pretender/pretender.js

And Pretender seems to need this https://github.com/pretenderjs/pretender/pull/166 to work. When I added that to the used pretender.js file - options do mock:

```
this.options('/campaigns', function () {
      return .....;
});
```

ksnyde:
@riklaunim got a bit confused on your npm versus bower **Pretender** versions comment. I'd expect that Mirage is only using the npm version but I think that's what you're saying. Anyway, glad to see your PR on Pretender as I pretty sure it had to originate from there based on what we were seeing hanging off the Pretender object. Sounds like you've nailed that.

ksnyde:
One other dimension to this that I've now noticed, is that as soon as I add the _passthrough_ to my `config.js` this somehow interferes with Mirage using my `scenarios/default.js` which is loading my fixtures and factories. It's very reproducible but haven't had time to dig into one might effect the other.

@samselikoff does anything jump out at you WRT to this?

riklaunim:
installing ember-cli-mirate adds pretender to bower dependencies so I expected it will use it from bower_components or if its want it in node_modules then it will be in node_modules and not /node_modules/subfolder/node_modules/ -- which is pretty hidden.

ksnyde:
Oh yes that is confusing. @samselikoff is there a reason for this?

ksnyde:
btw, removing my above comment until I can dig into it a bit further ... it's not as reproducible as i thought 😬

ksnyde:
@riklaunim I have forked your PR and pointed to your Pretender PR in the `package.json` so I get the full end-to-end effect. Works. I can send you the PR if you want but I'm guessing you'd prefer to keep these two PR separate even though the Mirage PR depends on the Pretender one.

ksnyde:
I can confirm though that a side effect does seem to be introduced. When I use this PR paired with the Pretender PR, I get no errors but Mirage _does not_ use my `config/default.js` configuration and therefore fixtures and factories are not used.

**Use Case 1**: PR breaks scenarios
**Use Case 2**: PR solves options problem

So to be clear, use case 1a:
- using published Mirage 0.2.1
- remove all passthrough configuration in `mirage/config.js` ... without which no OPTIONS calls will be made
- load an application page who's model ('forecast') has an associated fixture associated with it
- the `GET /forecasts` request results in the expected 80 rows that come from the fixture
- no error, expected outcome

Use case 1b:
- Use this PR alongside the Pretender PR
- exact same setup as 1a, outcome though is different ... no error, just no results


Use Case 2a:
- using default Mirage 0.2.1
- go to a page where `POST https://www.googleapis.com` is called (and indirectly the OPTIONS call)
- add passthrough --`this.passthrough('https://www.googleapis.com/**', ['post', 'head', 'options']);` -- to `mirage/config.js`.
- immediately fails on any page with the above mentioned "_this.pretender[verb] is not a function" error

Use Case 2b: 
- using Mirage and Pretender PR's (see [link](https://github.com/riklaunim/ember-cli-mirage/compare/master...ksnyde:master))
- No error, network request goes out as expected ... desired behaviour.

riklaunim:
Pretender needs to merge the PR and push a new version, then this can use it and not a github commit (which would be pretty odd for a production release).

ksnyde:
yes that was my point 👍 

I just did the fork for my immediate needs of seeing the two together.

ksnyde:
the issue above _does_ need addressing though as Mirage will not work correctly without addressing this

riklaunim:
I use it with a big client panel demo and it works. If the are problems - test should show them too - when the verb problem is solved by newer pretender.

ksnyde:
so you're loading fixtures and/or factories to initialise your state?

riklaunim:
Mostly factories, few custom responses. I have some passthrough in another project, can check it too.

ksnyde:
would you let me know how that comes out? I have to shift to a different project for a day but if you're not seeing this behaviour I'll dig into mine again and see what else could be causing it.

riklaunim:
Tested, works.

addon/route-handler.js
samselikoff: can we make this the FunctionHandler? I don't think shorthands make sense for options (or do they?)

samselikoff:
I'd love the tl;dr from your convo & experiments - what should our next steps be here? Are we blocked by the Pretender pr?

riklaunim:
At least some tests won't pass as Pretender.options does not exist. Then there may be some other options tests for fixing. It's also odd that node_modules/ember-cli-migrage/node_modules/.... subfolders shows up with Pretender - and not just the one in bower_components. I had to hand modify the Pretender code there instead of getting it from a git version in bower dependencies.

I'll see about the Pretender additional tests, then see those tests here.

samselikoff:
Awesome, thank you. So we should hold off on this PR until those are sorted out, right?

riklaunim:
Now fixture tests fail for some reason.

package.json
```diff
@@ -75,7 +75,7 @@
     "exists-sync": "0.0.3",
     "fake-xml-http-request": "^1.4.0",
     "faker": "3.0.1",
-    "pretender": "^1.2.0",
+    "pretender": "^1.3.0",
```
samselikoff: I think we can remove this change from this PR now? Once you do the tests will rerun and we can investigate any errors.

Let's get this puppy merged!

riklaunim: 1.2.0 doesn't have OPTIONS support. 1.3.0 is the first one to have it.

samselikoff: right but `master` is now on 1.4 so if you remove this & rebase we should be set

riklaunim: ah, rebase to master. OK :)

samselikoff:
You have some failures, are your tests passing locally?

riklaunim:
nop, as I commented above - few fixture related tests fail and I don't why.

riklaunim:
Can you help me with this? It seems like even adding options to the list - https://github.com/samselikoff/ember-cli-mirage/pull/841/commits/14523a7d52163e9f6dd44acb4fd989f766e40bd6#diff-3da129bcd0b0367d1b65b28e9ecc34c9 - causes the fixtures tests to fail. For some reason visit() return 404 - like if some mock or something stopped working.

samselikoff:
Closing due to inactivity

PR #844, opened 2016-08-20
Backport namespace fix

cibernox:


cibernox:
Strangely I had to fix a few JSHint warnings to make tests pass. I think that one implicit upgrade in JSHint made it more strict.

samselikoff:
Interesting. This looks great, thanks for opening!

1. Why the `aServer`, was it conflicting with global?
2. Did you mean to PR this against `fix-ember-lodash`?

This addresses the 2.7 issue, correct?

samselikoff:
Addresses #826

cibernox:
1. Yes. `server` is lister as a global in the .jshintrc, and theferore was an illegal name according to JShint.
2. Yes, the latest published version of 0.1.X was the branch, so in the absence of a 0.1 branch, that was the best target I found.

I'm using my fork with 2.7 now without any problem.

samselikoff:
Sounds like I should create an `0-1-stable` branch, do you agree? Then we can PR against that?

cibernox:
Yes, that sounds good. You can merge this PR and then create the branch or you can create the branch and I can change the target of this PR, the result will be the same.

cibernox:
@samselikoff can you public 0.1.14 with this fix?

samselikoff:
All set - sorry about the delay

Issue #719, opened 2016-04-26
Ensure getter/setters are defined for new attributes


samselikoff:
We need to call `_definePlainAttribute(attr)` in a model's `update` method, since it can add new attributes that the model wasn't aware of at runtime.

On [this line](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L81), call `_definePlainAttribute(attr)` before setting the attr. Also, refactor [`_definePlainAttribute`](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L179) to be idempotent (not redefine the getter/setter if it's already been defined).

Leooo:
+1, this is a big bug

Leooo:
`````
  updateAttrs: function (hash) {
    // https://github.com/samselikoff/ember-cli-mirage/issues/719
    Object.keys(hash).forEach((attr) => {
      if (!this.attrs.hasOwnProperty(attr)) {
        this._definePlainAttribute(attr);
      }
      this.update(attr, hash[attr]);
    });
    return this;
  },
`````

Leooo:
Also, would be good to have an `init` hook in the models to setup default attributes

Leooo:
OK, I think I know why this bug exists: model attributes are supposed to be defined in Factories. But then there is a big lack of documentation about what to do: define only the relationships in `models`, and all other attributes (without relationships) in Factories, and finally create models through server.create("modelName"), **NOT** schema.modelNames.create().

samselikoff:
Attributes actually don't have to be defined; mirage models are schemaless. That being said, when an instance of a model is instantiated, we need to call `_definePlainAttribute` on each attr, to set up some bookkeeping.

Factories are there to make it easy to seed the db with some values, not to define the schema. If that makes sense. For instance, if you didn't have a factory at all, and then your Ember app POST'ed with `{ some: 'data' }`, the models would be set up correctly.

Going to take a look at your PR

Leooo:
@3-john-4 we are not using `.get()` and `.set()` like in Ember, you should just use `ct.id`or `ct.update(propertyName, value)`

3-john-4:
I have a component that I want to pass this mirage test data to. I assign`model = ct` and then pass `model` as input to my component. Then in my code I'm doing a `model.get('id')`.  `model.id` would not work.

samselikoff:
[...]

PR #648, opened 2016-03-18
[WIP] Add association support for hasOne

samselikoff:
Affects: 
- #613

PR #846, opened 2016-08-23
Allow new attributes to be defined in #update

Leooo:
right now `model.update('status', true)` is not working, as only the `model.status` field is updated, and not `model.attrs.status` (which is the field used in the `save()` method: `this._schema.db[collection].update(this.attrs.id, this.attrs);`).

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -79,7 +79,9 @@ class Model {
     }
 
     Object.keys(attrs).forEach(function(attr) {
-      this.attrs[attr] = this[attr] = attrs[attr];
+      this.attrs[attr] = attrs[attr];
+      this._definePlainAttribute(attr);
+      this[attr] = attrs[attr];
     }, this);
 
     this.save();
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -79,7 +79,7 @@ class Model {
     }
 
     Object.keys(attrs).forEach(function(attr) {
-      this[attr] = attrs[attr];
+      this.attrs[attr] = this[attr] = attrs[attr];
     }, this);
 
     this.save();
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -79,7 +79,6 @@ class Model {
     }
 
     Object.keys(attrs).forEach(function(attr) {
-      this.attrs[attr] = attrs[attr];
       this._definePlainAttribute(attr);
       this[attr] = attrs[attr];
     }, this);
@@ -212,9 +211,6 @@ class Model {
    * @private
    */
   _definePlainAttribute(attr) {
-    if (this[attr] !== undefined) {
-      return;
-    }
 
     // Ensure the attribute is on the attrs hash
     if (!this.attrs.hasOwnProperty(attr)) {
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -223,6 +223,7 @@ class Model {
 
     // Define the getter/setter
     Object.defineProperty(this, attr, {
+      configurable: true,
       get() {
         return this.attrs[attr];
       },
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -212,6 +212,12 @@ class Model {
    */
   _definePlainAttribute(attr) {
 
+    // Ensure the property hasn't already been defined
+    let existingProperty = Object.getOwnPropertyDescriptor(this, attr);
+    if (existingProperty && existingProperty.get) {
+      return;
+    }
+
     // Ensure the attribute is on the attrs hash
     if (!this.attrs.hasOwnProperty(attr)) {
       this.attrs[attr] = null;
@@ -219,7 +225,6 @@ class Model {
 
     // Define the getter/setter
     Object.defineProperty(this, attr, {
-      configurable: true,
       get() {
         return this.attrs[attr];
       },
--- a/bower.json
+++ b/bower.json
@@ -6,7 +6,6 @@
     "ember-cli-shims": "0.1.0",
     "ember-cli-test-loader": "0.2.2",
     "ember-qunit-notifications": "0.1.0",
-    "sinonjs": "~1.17.1",
-    "Faker": "3.0.1"
+    "sinonjs": "~1.17.1"
   }
 }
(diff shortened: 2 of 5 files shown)


samselikoff:
#719

samselikoff:
Thanks for kicking this off!

I don't think we need to modify the associations - they're taken care of, because associations must be defined on models. What we need to do is call [_definePlainAttribute](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L212) on attrs that are passed to `update` that the model wasn't aware of when it was instantiated.

I think we want to add a line [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L81) in the forEach that calls `_definePlainAttribute` for each attr. I believe `_definePlainAttribute` is idempotent (it can be called multiple times safely), but we should add a test for this.

Think you could update your PR?

Leooo:
I'm not sure how to use `_definePlainAttribute`, because I'm not sure what it does (define `get` / `set` on each attribute?, but then we still need to `set` the attribute later in the `update` hook, what's the difference with just writing `this.attrs[attr] = this[attr]` then??

The code on the belongsTo is just picked up from https://github.com/samselikoff/ember-cli-mirage/pull/648, just to insist this is really in need to be merged..

Leooo:
@samselikoff updated to use `_definePlainAttribute`, but we still need to add a line before`this.attrs[attr] = attrs[attr]` to make sure that `this.attrs.hasOwnProperty(attr))` is true. Also added a test. 

Removed the relationship part which was not part of this fix, but the PR in https://github.com/samselikoff/ember-cli-mirage/pull/568 seems very sensible to me (you intend to have a model returned by `server.create` but that's not the case for now and things need to work).

L

Leooo:
Also added the `configurable` option without which `Object.defineProperty` can not be called twice (and corresponding test which is failing without configurable).

addon/orm/model.js
```diff
@@ -221,6 +223,7 @@ class Model {
 
     // Define the getter/setter
     Object.defineProperty(this, attr, {
+      configurable: true,
```
samselikoff: let's drop this and change line 215 from 

```js
if (this[attr] !== undefined) {
  return;
}
```

to

```js
if (this.hasOwnProperty(attr)) {
  return;
}
```

bower.json
```diff
@@ -6,6 +6,7 @@
     "ember-cli-shims": "0.1.0",
     "ember-cli-test-loader": "0.2.2",
     "ember-qunit-notifications": "0.1.0",
-    "sinonjs": "~1.17.1"
+    "sinonjs": "~1.17.1",
+    "Faker": "3.0.1"
```
samselikoff: don't need this, correct?

addon/orm/model.js
```diff
@@ -79,6 +79,8 @@ class Model {
     }
 
     Object.keys(attrs).forEach(function(attr) {
+      this.attrs[attr] = attrs[attr];
```
samselikoff: I don't think we need this once you make the change below

tests/integration/orm/update-test.js
```diff
@@ -71,3 +71,19 @@ test('it can update a hash of attrs immediately', function(assert) {
...
+
+test('it can update a non-existing attribute', function(assert) {
+  var link = this.schema.users.find(1);
+  link.update({ name: 'Evil link', evil: true, reallyEvil: 'absolutely evil' });
+
+  assert.deepEqual(link.attrs, { id: '1', name: 'Evil link', location: 'Hyrule', evil: true, reallyEvil: 'absolutely evil' });
+  assert.deepEqual(this.db.users.find(1), { id: '1', name: 'Evil link', location: 'Hyrule', evil: true, reallyEvil: 'absolutely evil' });
+});
+
+test('it can override an existing attribute', function(assert) {
```
samselikoff: do we need this test? Seems similar to L67 above

samselikoff:
Awesome, thank you for your work, I think we're almost there!

addon/orm/model.js
Leooo: true

addon/orm/model.js
Leooo: done, but wouldn't it be dangerous for users trying to set directly the property without using update? would mess things up then right?

bower.json
Leooo: as you saw in https://github.com/samselikoff/ember-cli-mirage/issues/850 we do

Leooo: correct

tests/integration/orm/update-test.js
Leooo: not if you removed the `undefined` test above, indeed

addon/orm/model.js
Leooo: Answer: no, I added a test for it

Leooo: To make the test I added pass, it was necessary to remove the checks before Object.defineProperty, which means it will be run every time. Should be fine to do that I suppose.

addon/orm/model.js
```diff
@@ -210,9 +211,6 @@ class Model {
    * @private
    */
   _definePlainAttribute(attr) {
-    if (this[attr] !== undefined) {
```
samselikoff: Instead change this to

```js
let existingProperty = Object.getOwnPropertyDescriptor(this, attr);
if (existingProperty && existingProperty.get) {
  return;
}
```

then we can remove `configurable: true`

samselikoff:
Starting to think we should have used a `.attrs` field on models. `link.someNewProp = 'foo'` is technically part of the api but this actually doesn't work:

```js
link.someKnownProp = 'foo';
link.someNewProp = 'bar';

link.save();
```

Save won't persist the unknown prop. 😕 

I guess the the best we can do for now is make `update` work with new attrs. Possible eventual paths forward are (1) tell users to interact with `.attrs` (2) use `.set` or (3) allow users to define a schema.

samselikoff:
One last comment then we can merge this. Thanks for all your work!

addon/orm/model.js
Leooo: @samselikoff done

samselikoff:
🎊

PR #848, opened 2016-08-25
Allow models to be updated by mirage created via a factory

tim-evans:
We were seeing consistently a case where creating a new model, then
updating via the UI wouldn't persist changes to mirage's db.

I tracked this down to `this.save()` being called with `this.attrs`.
There were missing Object descriptors on the model due to mirage
skipping properties that were not undefined, even though there were
no descriptors for them.

To fix this, I used Object.getDescriptors() to introspect on the model
which properties have been defined. This fixes the issue and allows
for updates to pass through smoothly.

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -210,13 +210,14 @@ class Model {
    * @private
    */
   _definePlainAttribute(attr) {
-    if (this[attr] !== undefined) {
+    let descriptor = Object.getOwnPropertyDescriptor(this, attr);
+    if (descriptor != null) {
       return;
     }
 
     // Ensure the attribute is on the attrs hash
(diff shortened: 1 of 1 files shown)


kashiif:
We are having the same issues and I was actually going to look into this today. Glad that I first looked into merge requests :). Thanks @tim-evans

samselikoff:
Thanks for the PR! I think we actually need to call `_definePlainAttribute` when `update` is used - I explained why in [this comment](https://github.com/samselikoff/ember-cli-mirage/pull/846#issuecomment-243001419). If you'd like, you could make that change here (with a test) and then I'd be happy to merge this. It should solve the issue you were running into, as well.

tim-evans:
This looks like it's fixed! 🎉

PR #849, opened 2016-08-25
Feature/add seed data builder

Leooo:
This PR adds to models the ability to create flexible seed data for their relationships.

This is a powerful way to create seed data by combining this with mirage factories / routes:

Example of seed data (a customer has many premises, a premise has many energyAccounts and servicesAccounts).

`````
 server.create('customer')
  .anySeedFunctionDefinedInTheCustomerModel() // <= defined on customer, returns customer
  .hasOne('premise', { // 
    status: hasPassedScreening || null
  }) // <= defined on customer, returns premise
  .hasNo('servicesAccount') // <= defined on premise, returns premise
  .hasOne('energyAccount') // <= defined on premise, returns energyAccount
  .isJointInvoiced() // <= defined on energyAccount, returns energyAccount
  .anySeedFunctionDefinedInTheEnergyAccountModel(); <= defined on energyAccount, returns energyAccount
`````

Each model has default seed data defined in Model.default()

We made `server` available as a global in the `development` environment to make it accessible inside Mirage models (was only a global in `test` before).

Haven't added tests yet but could do if you want to merge the PR.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -3,7 +3,8 @@
     "document",
     "window",
     "-Promise",
-    "faker"
+    "faker",
+    "server"
   ],
   "browser": true,
   "boss": true,
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -3,6 +3,7 @@ import _assign from 'lodash/object/assign';
 import { capitalize, camelize } from 'ember-cli-mirage/utils/inflector';
 import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
 import assert from 'ember-cli-mirage/assert';
+import Ember from 'ember';
 
 class BelongsTo extends Association {
 
@@ -76,13 +77,20 @@ class BelongsTo extends Association {
           - sets the associated parent (via model)
       */
       set(newModel) {
-        if (newModel && newModel.isNew()) {
+        if (!Ember.isEmpty(newModel) && typeof newModel === 'object' && typeof newModel.isNew === 'function' && newModel.isNew()) {
+          // newModel is a model
           this[foreignKey] = null;
           association._tempParent = newModel;
-        } else if (newModel) {
+        } else if (!Ember.isEmpty(newModel) && typeof newModel === 'object' && !Ember.isEmpty(newModel.id)) {
+          // newModel is a simple object with id property
           association._tempParent = null;
           this[foreignKey] = newModel.id;
+        } else if (!Ember.isEmpty(newModel)) {
+          // newModel is an id
+          association._tempParent = null;
+          this[foreignKey] = newModel;
         } else {
+          // newModel is not present
           association._tempParent = null;
           this[foreignKey] = null;
         }
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -79,7 +79,7 @@ class Model {
     }
 
     Object.keys(attrs).forEach(function(attr) {
-      this[attr] = attrs[attr];
+      this.attrs[attr] = this[attr] = attrs[attr];
     }, this);
 
     this.save();
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -25,9 +25,9 @@ class Model {
     this.modelName = modelName;
     this.fks = fks || [];
     attrs = attrs || {};
-    attrs.childrenAssociations = attrs.childrenAssociations || [];
     this._setupAttrs(attrs);
     this._setupRelationships(attrs);
+    this.childrenAssociations = this.childrenAssociations || [];
     // the associations in this list will be destroyed in beforeDestroy
     // TODO: define cleaner hasOne / belongsToMany relationships
 
@@ -109,11 +109,8 @@ class Model {
     * @public
     */
    _beforeDestroy() {
-     console.log('beforeDestroy', this.modelName, this.associationKeys);
      this.associationKeys.forEach((relName) => {
-       console.log('beforeDestroy 2', this.modelName, this.childrenAssociations, relName);
-       if (this.childrenAssociations.contains(relName)) {
-         console.log('hasNo beforeDestroy', this.modelName, relName);
+       if (this.childrenAssociations.indexOf(relName) > -1) {
          this.hasNo(relName);
        }
      });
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by samselikoff** [comment]
Interesting approach, thanks for opening this. I think I'd like to go in more of a factory_girl direction - I really like working with their API. It might look something like this:

```js
// models/post.js
export default Model.extend({
  comments: hasMany()
});

// models/comment.js
export default Model.extend({
  post: belongsTo()
});

// factories/post.js
import { Factory, trait } from 'ember-cli-mirage';

  title() {
    return faker.lorem.sentence()
  },

  withComments: trait({
    afterCreate(post, server) {
      server.createList('comment', 3, { post });
    }
  })
  
})
```

Use would be something like

```js
server.create('post', 'withComments');
```

We can also add an `association` helper so users can seed relationships directly:

```js
// factories/post.js
import { Factory, trait } from 'ember-cli-mirage';

  title() {
    return faker.lorem.sentence()
  },

  comments: association({
    count: 3
  })
  
})
```

which would set up the association by default.

Leooo:
whichever way would work I suppose, as long as we have the ability to flexibly change the relationships (right now the `afterCreate` hook in factories can just take care of one case).

What I like in my approach is the ability to write very user-friendly seed data for test cases ("post.hasOne('comment').isAdmin().hasMany('features')") (where `isAdmin()` and `features` are defined on `comments`) which seems to need access to models themselves.

In your code above, with `server.create('post', 'withComments');`, I don't know then how to access these comments to setup their own properties / relationships. In my case I can do `server.create('post').hasMany('comments', {isEditable: true})`, and it will return the different comments on which I can define more associations. This is nicely chained.

samselikoff:
I think `withComments` would be a trait that creates comments in `afterCreate`, and that's where you could specify overrides/traits for those comments:

```js
// factories/post.js

  withComments: trait({
    afterCreate(post, server) {
      server.createList('comment', 3, 'someTrait', { an: 'override' });
    }
  })
});
```

But its true you couldn't specify traits on the comments from the test. You'd need a different trait on the post for that.

Leooo:
@samselikoff ;) so shall we go in this direction then? BDD tests written like this are just awesome: `customer.hasAnAccount.isUsingFixedDirectDebit.etc`.

Should you wish to go in this direction I'm ready to add tests for the `hasOne`, `hasNone`, `hasMulti` methods above. What was missing to make this PR work was:
* Make `server` available as a global in development (or make it available inside Models at least).
* An `init` hook inside models (using a `default()` function right now but it's ugly, i didn't get how to properly override the constructor as `_super` didnt work).

?

samselikoff:
Added some comments in #852, want to get your response but let's discuss there.

Leooo:
@samselikoff , using my `hasOne`, `hasMulti`, `hasNone` etc. methods on Mirage models now and it's just great. Still, I'm seeing now that I would ideally need observes on Mirage models (to update properties when children change) etc., and the more I think about it, the more I'm realizing that actually Mirage models should just be Ember-data models, with a different `test` store (accessible from `server`), a specific `testSerializer` etc. This way we could leverage the whole Ember-data infrastructure (ember inspector etc.). You may have already thought about that when building Mirage but I would be curious to know your thoughts about that: it seems to be that you are going in a way with the Mirage models where you will actually have to build many more things to make the whole stuff coherent (normalizers etc.), while it would be simpler to just leverage  the framework. 

?

Leooo:
:thumbsdown:

samselikoff:
Hey, I am responding right now, didn't mean to close without commenting. Writing...

samselikoff:
Closing this for now since we've merged traits.

Thanks for your thinking on this, keep it coming. Yes we actually had this very discussion many months ago, and a few other times again since then. Originally the idea was vetoed mostly based on instinct: Ember Data is a library built for a few specific reasons: (1) client-side identity map (2) fundamentally asynchronous (talk to multiple hosts, deal with HTTP lifecycle etc). Mirage's db/orm layer originally wasn't either of these, so I didn't want to force a round peg into a square hole. Also many users don't use Ember Data, and originally it was very important for Mirage to be minimal in terms of its API and complexity so that folks would actually use it.

Obviously over the past year Mirage's complexity has grown, and we now have an ORM and the db is effectively an identity map/data store. Mirage-in-the-browser is still fully synchronous, but there's been talk about allowing it to run in node (we'd get the Network devtools for free, for one). So, it's true that it seems like the needs of Mirage's data layer have come closer and closer to those of Ember Data.

Some more random thoughts:

- I'd like to extract core Mirage from Ember CLI Mirage one day (when? I don't know), as Mirage itself is unaware that it's ever talking with an Ember app. It's true that we could do this while still relying on standalone Ember Data but it would make more integration work for us here, make Mirage core rely on another external dependency etc.
- Ember Data objects are observable, and while you cited that as something you'd like to have in Mirage I'm not convinced it's needed (definitely willing to discuss). Despite in the increase in complexity I still want Mirage to be as simple as possible. Data creation should be focused on setting up server scenarios for dev/testing, that's about it.

Anyway that's kind of a brain dump, I'm happy to keep the discussion going. I have a contract to round out the ORM with the missing relationships and I'll most likely be doing that by working on the existing data layer. Bringing in Ember Data is a big step and would take some more thought.

Thanks again for your input.

samselikoff:
Oh and I closed this PR not to shut down the convo but because we merged traits in. I'm glad your API is working for you and your team, I actually think that's good and teams should feel like Mirage's primitives are low-level enough to allow them to make higher-level helpers specific to their domain, which makes setting up scenarios easy.

Leooo:
Thanks for your response. All makes sense, especially the asynchronous nature of Ember-data (except with `{async: false}` relationships? And except with custom serializers / adapters, see `localStorageAdapter`) ?

In my case the need for observers comes from the fact that we can have different server responses to `PUT` or `POST` requests, with complex logic based on, say, answers to a survey. 

I also saw the need in this PR to have `init` methods, `beforeDestroy` hooks etc.

but yes, I realize that this actually makes things closer to Ember-Data, and yes it would be a big work to include ED into Mirage (probably a new add-on?).

In case you may want to go into that direction (i.e. in the direction of this PR or even towards combining Ember-data), I could be interested to take part.

samselikoff:
Yah. I'll definitely continue to think about it, and thanks for offering your help. I'll keep it in mind.

Again in most of the use cases I've seen folks don't need those advanced hooks, so if it's only 5-10% of users needing them I'd rather push that out to those specific users. However this could be colored by my personal use of Mirage. In my projects I've never felt this need, but I should probably take a survey and see what features folks think are missing.

PR #858, opened 2016-09-01
Add `sort` to orm/collection

lukemelia:
`sort` returns a new collection with models sorted according to the function as an argument to the method.

@samselikoff Should docs be in a separate commit?

### DIFF
--- a/addon/orm/collection.js
+++ b/addon/orm/collection.js
@@ -82,6 +82,18 @@ export default class Collection {
   }
 
   /**
+   * @method sort
+   * @param f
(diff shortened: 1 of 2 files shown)


samselikoff:
Wonderful! Docs would be awesome, could you go [here](http://www.ember-cli-mirage.com/docs/v0.2.x/collections/) and click "Improve this page"? Also would you mind throwing `filter` in there? I didn't realize we already had that.

Eventually we should add All The Enums and just proxy to array/lodash methods (basically what Backbone does).

PR #862, opened 2016-09-01
reuse getForeignKey in getForeignKeyArray method in BelongsTo class

Azdaroth:
Not a big thing, but I think it looks more expressive this way besides being DRY.


samselikoff:
yessir!

❤️

PR #863, opened 2016-09-02
peg faker at 3.0.1

### LINKED ISSUE #850: faker.random.uuid in a factory doesn't work with Faker 3.1
[A bug/change](https://github.com/Marak/faker.js/issues/368) in Faker 3.1 causes exceptions if you have a mirage factory that uses `faker.random.uuid`.  The reason for this is that the context of `self` inside faker's uuid function [here](https://github.com/Marak/faker.js/blob/e8231413cc38a3ce49d712823dadbffbd1849acc/lib/random.js#L93) is the factory object itself when called from mirage [here](https://github.com/samselikoff/ember-cli-mirage/blob/216250b8d6c8530474edb7440697c8de2efdd359/addon/factory.js#L88).

Prior to https://github.com/samselikoff/ember-cli-mirage/pull/813 it was easy enough for consumers of ember-cli-mirage to peg their preferred version of Faker in their bower.json, but subsequent to #813 that doesn't seem possible.

I'm not sure if there is anything specific we could do with ember-cli-mirage to avoid this, but I thought it was worth opening an issue in case others have encountered this problem and had some insight on how to proceed.

_It's also worth noting that Faker 3.1 also includes at least [one other breaking change](https://github.com/Marak/faker.js/issues/356) and that in general Faker doesn't seem to follow semver._

Dhaulagiri:
Fixes #850

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ember-lodash": "0.0.9",
     "exists-sync": "0.0.3",
     "fake-xml-http-request": "^1.4.0",
-    "faker": "^3.1.0",
+    "faker": "3.0.1",
     "pretender": "^1.1.0",
     "route-recognizer": "^0.1.11"
   }


samselikoff:
Thank you! 🙇

johanbove:
Thanks for the fix, but would have been cool if this change was mentioned in the changelog. We're using the mirage "faker" library to generate our data and in version 2.2.0 we're missing a couple of methods like faker.internet.exampleEmail() all of a sudden.

samselikoff:
@johanbove you're right - I think I merged this in haste, and realize that users are no longer able to swap out their own version. Is that correct? If so it seems like it's a problem.

Dhaulagiri:
There is definitely a bit of a catch-22 here.  Users who got used to new features 3.1.0 are having a bad time now but users who had breaking changes with 3.1.0 are in a better spot.

samselikoff:
If we relax the version in mirage's package.json, would that allow end users to be able to specify their own resolutions in their app's package.json?

johanbove:
Thanks for the follow up. Appreciate it! I believe it will be okay to change the Faker dependency if this is announced in the changelog and the version number for Mirage reflects a "breaking" change according to the semver contract.

Dhaulagiri:
@johanbove see #932 for a potential solution that will could make everyone happy 🎉

PR #864, opened 2016-09-02
Update upgrading.md

adamjmcgrath:
Noticed that the mirage folder wasn't being removed from our production build after upgrading to 0.2.x because we were using the `directory` feature to point it at the old `/app/mirage` directory.


samselikoff:
Thanks, good note to have. Although I guess we should ideally write broccoli code to strip it when not included, from whatever the configured directory is

Issue #852, opened 2016-08-30
Factory traits


samselikoff:
Traits let you have different "versions" of a factory:

```js
// factories/post.js
import { Factory, trait } from 'ember-cli-mirage';

export default Factory.extend({

  title: 'Lorem ipsum',

  published: trait({
    isPublished: true,
    publishedAt: '2010-01-01 10:00:00'
  }),

  withComments: trait({
    afterCreate(post, server) {
      server.createList('comment', 3, { post });
    }
  })
  
});
```

Invoke traits during `server.create`, by passing in a variable-length list of trait names:

```js
let defaultPost = server.create('post');
let publishedPost = server.create('post', 'published');
let postWithComments = server.create('post', 'withComments');
let publishedPostWithComments = server.create('post', 'published', 'withComments');
```

You can still pass in overrides, by passing in an object as the last argument:

```js
let publishedPost = server.create('post', 'published', { title: 'Rails is omakase' });
```

Azdaroth:
I will give it a try this weekend if nobody has started working on traits yet, definitely a super useful feature.

samselikoff:
Would be amazing!

Some folks had asked about a chaining API instead, but I can't think of one I like.

Leooo:
@Azdaroth @samselikoff yes, I think https://github.com/samselikoff/ember-cli-mirage/pull/849 is better.
 
You want to write real-life-like scenario for your seed-data, and this has to be flexible to encompass multiple cases / trees: so `server.create('customer').hasOne('premise').hasNo('energyAccount').hasOne('serviceAccount').hasOne('appointment', attrsForAppointment)` really seems the natural way to go

* Factories are responsible for defining and creating seed data for attributes
* Models are responsible for defining and creating relationships

What's wrong with that?

Azdaroth:
Maybe I got too much used to using factory_girl, but still I would say traits are more natural to work with and the API is less complex, they work just as an extra extensions. 

At the end of the day, I think the difference between working with traits and chaining API would be  `server.create('article', 'withComments', 'published')` vs `server.create('article').withComments().published()`. Not really sure if it's a good idea to have this setup split between the factories and  the models, it forces you to check both factory and the model if you want to understand what is actually handled where. For the associations I certainly see some benefits like ability to pass overrides instead of doing it in `afterCreate` hook, but something like `published` which doesn't handle associations, but only adds some extra attributes, definitely looks like some trait that should belong to factory.

[...]

PR #866, opened 2016-09-03
Implement traits

Azdaroth:


Azdaroth:
Solves https://github.com/samselikoff/ember-cli-mirage/issues/852

Works pretty fine, but passing arguments and extracting them based on their types in `create` and `build` is a bit painful... I was wondering if it's worth checking the sequence of the arguments (e.g. that there: traits, attr overrides and collectionFromCreateList) and not (collectionFromCreateList, traits, overrides), but seemed not really worth it comparing to the complexity it would introduce. Currently it actually doesn't matter and all these variants would work. The only surprising thing would be passing more than 1 `collectionFromCreateList` to `create` or passing more than 1 `overrides` as only the first argument of given type would be picked. But again, didn't want to make the arguments extraction too complex, I think it is quite unlikely there would be any issue with that.

Let me know what you think.

samselikoff:
Amazing work man! Going to wait just a bit in case there are more comments on #852, then will take a look and get this merged. Super excited to get this in here.

samselikoff:
Great work. I think the extraction is okay for now, `collectionFromCreateList` is an internal thing that is used so `db.collection` doesn't need to be called each time in the loop for `createList` (`db.collection` does a JSON.stringify copy of the entire db collection, so it's expensive).

samselikoff:
🎊

Azdaroth:
🎉

PR #871, opened 2016-09-04
consistently use native methods instead of lodash

Azdaroth:
I saw some methods were inconsistently used in quite a few places (`_keys` vs. `Object.keys`, `_isArray` vs `Array.isArray`), thought it's better to keep it consistent and used the native ones where possible. 

Had to add extra conditional here (https://github.com/samselikoff/ember-cli-mirage/compare/master...Azdaroth:methods-consistency?expand=1#diff-3da129bcd0b0367d1b65b28e9ecc34c9R139) as `keys` from `lodash` works fine with `null`, but `Object.keys` doesn't.


samselikoff:
c07911adf08369c21a66d7ea17b34e2dd685b94a closes (had to fix a few conflicts)

Thanks!

PR #873, opened 2016-09-06
Small fixes

Leooo:
* Add a few jshint fixes (at least for my jshint).
* Add route tests (was initially to find a failing case for https://github.com/samselikoff/ember-cli-mirage/pull/568 but it seems the failing test would have been too specific, still good to have tests on this part).
* Added a `&& isNew` check in `orm/model.js` which was causing a very nasty bug on my side in phantomjs only (parent was an `id` instead of a model). Feel free to not merge this one but I feel this could save massive headaches to other ppl.

L

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,3 +39,7 @@ Check out the [Docs](http://www.ember-cli-mirage.com/docs/v0.2.x/)!
 ## Support
 
 Having trouble? Open an issue!
+
+## Contributing
+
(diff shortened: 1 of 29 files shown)


samselikoff:
This is a lot of stuff. Some thoughts:

- you should configure your jshint to use the project's `.jshint` in the root of the project, so we're all using the same one. According to that there are no current failures (or else the build would be failing)
- I like to minimize the amount of files in the dummy app, since those are effectively global and will just grow over time. Better to isolate the server setup within a test, if at all possible.
- I'm not sure what the test you've added is demonstrating. As a first step, it might be good to open a new issue/PR with the behavior you think is currently broken, along with a failing test. That way we can discuss before you start implementation.

Leooo:
ok I will just close this one then.

PR #880, opened 2016-09-10
implement association helper

Azdaroth:
Another feature I was happy to see in the roadmap as it's super useful :). Works with name override, traits and attribute overrrides.

I think traits / association handling may need some refactoring, the responsibilites are a bit split betweem the factory itself and the server and it's not really clear what should happen where (e.g. some attributes are deleted in factory, others in server). And the building / creating could be actually delegated to some other object so that there's no so much logic directly in the server. But that's probably out of the scope of this PR.

### DIFF
--- a/addon/association.js
+++ b/addon/association.js
@@ -1,0 +1,10 @@
+let association = function(nameOverride, ...traitsAndOverrides) {
+  let __isAssociation__ = true;
+  return {
+    nameOverride,
+    __isAssociation__,
+    traitsAndOverrides
+  };
+};
+
+export default association;
--- a/addon/association.js
+++ b/addon/association.js
@@ -1,7 +1,6 @@
-let association = function(nameOverride, ...traitsAndOverrides) {
+let association = function(...traitsAndOverrides) {
   let __isAssociation__ = true;
   return {
-    nameOverride,
     __isAssociation__,
     traitsAndOverrides
   };
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,5 +1,6 @@
 import Factory from './factory';
 import trait from './trait';
+import association from './association';
 import Response from './response';
 import faker from './faker';
 import Model from './orm/model';
@@ -21,6 +22,7 @@ function belongsTo(...args) {
 export {
   Factory,
   trait,
+  association,
   Response,
   faker,
   Model,
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -288,5 +288,8 @@ class Model {
 }
 
 Model.extend = extend;
+Model.findBelongsToAssociation = function(associationType) {
+  return this.prototype.belongsToAssociations[associationType];
+};
 
 export default Model;
--- a/addon/orm/schema.js
+++ b/addon/orm/schema.js
@@ -105,6 +105,15 @@ export default class Schema {
   }
 
   /**
+   * @method modelFor
+   * @param type
+   * @public
+   */
+  modelFor(type) {
+    return this._registry[type];
+  }
+
+  /**
    * @method new
    * @param type
    * @param attrs
--- a/addon/server.js
+++ b/addon/server.js
@@ -2,7 +2,11 @@
 
 import { pluralize, camelize } from './utils/inflector';
 import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
+<<<<<<< HEAD
 import Ember from 'ember';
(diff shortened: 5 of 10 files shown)


samselikoff:
Awesome work.

I'm confused about the argument to association, for example `kind: association('category')`. What is `category` in this case, is it because the association on the original model is named differently from the model's type? Can't we introspect that?

Also, in the future, instead of using factory functions for objects let's use ES6 classes, to keep with the style of most of the other objects in Mirage.

Azdaroth:
>  What is category in this case, is it because the association on the original model is named differently from the model's type? Can't we introspect that?

Yeah, exactly, just like this example shows:

```
post: Model.extend({
  author: belongsTo(),
  kind: belongsTo('category')
})
```

so it would need to be:
```
kind: association('category')
```

I wanted to be explicit here and avoid extra coupling. We probably could introspect it, but that would require coupling to models - we currently can create / build new records without even registering the models, it's just ` this.factoryFor(type)` and we have a factory for given type without involing enything else. So not sure if it's worth doing. What do you think?

samselikoff:
Hm, well the factory relationships were intended to be used with the ORM. I mean the ORM is what makes the relationship stuff possible at almost every level - the relationship methods on models, the serializer relationship lookups etc.  I don't think folks should be using `association` without using Models, this is not really supported.

I think the name of the association should be the named relationship on the model (`kind` here), I don't think the factory should take in `type`. Again if a model had even more information on that named relationship (inverse, polymorphic etc) the factory layer should not be concerned with this. It should just delegate to the ORM and say, "I want an associated `kind` for this model, whatever that may be."

Thoughts?

Azdaroth:
Well, I agree with that, but I think it mostly comes to down to the question if we want to couple factories to models. Currently factories are independent from the models and the ORM provides the further integration - factories just build / create some objects that are inserted to db. And we can actually create whatever we want with factories - it doesn't have to be a model. Not sure if it's a valid and/or intended use case for factories, but it's worth keeping in mind. 

The `association` helper in current implementation is not literally for associations but rather embedded objects created by factories,  `association` is just a name that makes most sense here.

So for either case, there is some trade-off: coupling to models would be more cohesive probably, but it will mean less flexibility in the kinds of objects we can create. If the direction is to use factories only for models, I guess it would be a good idea to introduce some coupling and introspect info for `association` from models.

Azdaroth:
@samselikoff Thoughts?

samselikoff:
I'd rather couple to models here. I think

```
Model.extend({
  author: belongsTo('user')
})

Factory.extend({
  author: association('user')
})
```

breaks the mental model for factories too much - author _is_ a `user`, thats all it ever can be, so you shouldn't have to declare this type. I think factories should be thought of as corresponding to model attributes and relationships, and Mirage's flexibility in other areas has at times caused confusion about how the pieces should fit together.

There's a possibility we'll want `association` to take an arglist of traits, i.e. `author: association('admin')`. For now folks can just do this using afterCreate.

Also, if it comes up, we could always add a `factory` helper which does what your current version of `association` does.

So if you could rebase & introspect factory type based on the named attr, this should be good to merge! Thank you so much for your work and patience, let me know if you have any more thoughts.

Azdaroth:
Ok, if that's the clear intention then sure thing, I'll make it dependent on models, expect some new stuff today or tomorrow :).

Btw. it's already possible to pass the list of traits and overrides to `association` helper ;)

samselikoff:
> Btw. it's already possible to pass the list of traits and overrides to association helper ;)

you rule. 👊

Azdaroth:
@samselikoff Got some great news here, just check the code, it's pure magic now with models' introspection for getting some data about association :D

Can't wait to use it in some apps 💪

Azdaroth:
This is beyond the scope of this PR, but I think `server` could use some refactoring here, there's just too many implementation details and some stuff like building, creating etc. could just be delegated to some separate objects, how does it sound?

samselikoff:
Definitely `server` can use refactoring, let's break it down so we can do it in small steps.

Is this ready?

Azdaroth:
Yeah, `association` helper is ready.

Server refactoring is for sure something for another PR, I guess I could give it a try this week.

samselikoff:
💥 

Thanks for all your work on this dude!

Azdaroth:
Awesome, some docs might be useful here, coming soon ;).

Issue #845, opened 2016-08-22
Mechanism for per-test request assertions


aquamme:
First off, thanks for the nice addon @samselikoff.

In acceptance tests, I oftentimes want to assert that my app made a request to the right endpoint with the right parameters. This can be accomplished by overriding request handlers in the test, like so:

```javascript
test('Item should be hidden when the hide button is selected', function (assert) {
  let item = server.create('item');
  server.put('/items/:id', (schema, request) => {
    assert.equal(request.params.id, item.id);
    // perform the usual logic for this endpoint, then...
    return item;
  });

  // test
  visit(...);
});
```
This pattern works, but it entails duplicating whatever logic you may have in the request handler you overrode.

To solve this problem, I propose adding a hook that lets you examine the parameters passed to mirage before the request handler is invoked. Something like:

```javascript
  server.beforePut('/items/:id', request => {
    assert.equal(request.items.id, expectedItemId);
  });
```
By default, the `beforeVERB` hooks would be NOOPs.
Any return value would be discarded.
All `beforeVERB` hooks would be reset to their default NOOP between tests.

I'm glad to do the work to add the feature, but wanted to discuss the idea and the potential API first.

samselikoff:
Thanks for opening this, and you're welcome!

Lots of folks have asked about this so I'm happy to brainstorm. Reading through your comment made me think the ideal API really has nothing to do with redefining route handlers, but with assertions. Just thinking out loud, I wonder if there's a possibility to bring along some custom assertions with Mirage?

```js
test('Item should be hidden when the hide button is selected', function(assert) {
  let item = server.create('item');

  visit(`items/${item.id}`);
  click('.hide-item');
  click('.save');

  andThen(() => {
    assert.server.received('PUT').toURL(`/items/${id}`).withParams({
      hidden: true
    });
  });
});
```

This would be pretty involved (lots to think through in the API here), but I know it'd be valuable.

FWIW if my endpoints are doing run-of-the-mill crud I often will just assert on Mirage's server state directly:

```js
assert.equal(server.db.items.find(id).hidden, true);
```

aquamme:
That custom assertion syntax is certainly aesthetically pleasing -- I'm a fan :smiley: I'll spend some time pondering and looking over how it might be used in the app I'm working on. A few questions that come to mind immediately:

How might we handle asserting that a request didn't happen?
```javascript
assert.server.didntReceive('POST').toURL('/posts');
[...]

PR #882, opened 2016-09-12
Add request assertion feature to server

aquamme:
Implements #845 

The server tracks all handled requests and provides an API to allow
users to check whether a request matching the given parameters has
been made since the server was started.

``` javascript
  server.received.get.to('/contacts', { name: 'Darth Vader' })
  server.didNotReceive.post.to('any');
  server.received.get.to(/\/secrets.*/);
  server.didNotReceive.request.to('/robot/5');
  server.received.put.with({ name: 'Jack' });
```

I ended up adding RegExp support mostly because it was simpler than adding support for paths like "/posts/:id".

There is also a getter for the last request made to the server:

``` javascript
  server.lastRequest
```

Comments and suggestions are welcome!


addon/request-assertion.js
```diff
@@ -0,0 +1,90 @@
+/* * RequestAssertion implements a simple DSL for asserting that specific requests were made.
```
aquamme: Not sure if `RequestAssertion` is the best name since 'assertion' is a loaded term. Maybe something like `RequestChecker`?

addon/request-assertion.js
```diff
@@ -0,0 +1,90 @@
...
+  return foundMatch === expectMatch;
+}
+
+class Verb {
+  constructor(requests, received) {
+    ['get', 'post', 'put', 'delete', 'head', 'request'].forEach(verb => {
+      Object.defineProperty(this, verb, {
+        get() {
+          let f = to.bind(null, requests, received, verb);
+          return { to: f, with: f };
```
aquamme: `with` is an alias for `to`, only because it seems to read better in cases like this `server.received.put.with({ name: 'Todd' })` -- could remove it though.

aquamme:
A few other things I'd like to do/consider/I missed:
* Add an option to cap the number of requests that the server holds on to
* Add support for comparing POST payloads with nested attributes
* Have the feature report the number of matching requests instead of a boolean -- might want to rework the DSL a bit if this idea is pursued -- `server.received.post.to('/posts')` doesn't look like it would return a number

samselikoff:
Sorry about delay, hoping to get to this soon.

Just glancing over, will folks want to assert against content type or other things? Should we

```
assert.ok(this.server.received.post.to('/contacts', {
  with: {
    body: { name: 'Bilbo', age: 111 },
    headers: {
      'ContentType': 'foo'
    }
  })
);
```

Don't have to change right now, just a quick thought.

aquamme:
I like the idea. I'm on vacation until the 4th, so I won't be able to update the PR until then. Glad to hear any other feedback you may have when you get to it. Thanks!

samselikoff:
Few suggestions:

- Make `_requests` public: `server.requests`. This way folks can inspect during dev, or even use w/their assertions (`assert.equal(server.requests.length, 3);`)
- Move `server.lastRequest` to `server.requests.last` (I think you can define a property on the array)
- I think the DSL currently doesn't allow `server.received.post`, right? You have to specify a url? ... thinking about this, I think it's okay.
- Originally I said enable only in testing by default, but I think we should enable in all envs by default, and maybe add an option to disable if memory gets out of hand. (We can do this in a new PR).

aquamme:
Great feedback -- I should have time to make those changes tomorrow evening.

aquamme:
All right, this is ready for another peek.

`server.received...` returns the number of matching requests, while `server.didNotReceive` returns a boolean. So you can do things like
```javascript
// Two GETs are made to '/contacts', then...
assert.equal(server.received.get.to('/contacts'), 2);
assert.ok(server.received.get.to('/contacts'));
assert.ok(server.didNotReceive.post.to('/contacts'));
```
You can assert against any FakeXHR attributes:
```javascript
assert.ok(server.received.get.to('/contacts', { requestHeaders: { 'Foo': 'Bar' } }));
```

Also made `requests` a public property and added a `last` property.

> I think the DSL currently doesn't allow server.received.post, right? You have to specify a url? ... thinking about this, I think it's okay.

I don't see how we could make that syntax work while still allowing further chaining with 'to'. That aside, I did have a mechanism for allowing any URLs but I removed it because I didn't like the API (`server.received.request.to('any')`), but it could easily be added back in some form.

aquamme:
@samselikoff -- any further changes you'd like to see on this? I can write up some documentation once you give it the 👍

samselikoff:

PR #911, opened 2016-10-03
Update version of resource helper to 0.2.2

morhook:
I've stumbled at using this.resource, as it was not clarified what version it was added to mirage (still using 0.2.1)


acorncom:
@morhook The normal way we're adding these types of notes is with a comment after the first usage in the docs (see screenshot).  Could you tweak this PR to note it that way instead?
screen shot 2016-10-05 at 8 28 36 am

acorncom:
Thanks, that looks good to me.  Will merge that now /cc @samselikoff

Issue #710, opened 2016-04-21
Add pagination support to JsonApiSerializer


samselikoff:

Azdaroth:
Hey Sam, have you had some new ideas on pagination?

That's something I deal occasionally with and would be really nice to have something working out of box. Actually, I even needed it for an addon I released last week for dedicated pagination for JSONAPI backend and that's how I handled it in the dummy app: serializer (https://github.com/BookingSync/ember-cli-jsonapi-pagination/blob/master/tests/dummy/mirage/serializers/rental.js) + customized route handler (https://github.com/BookingSync/ember-cli-jsonapi-pagination/blob/master/tests/dummy/mirage/config.js). 

JSONAPI doesn't enforce any particular strategy for pagination, so not really sure if all strategies mentioned in the docs should be supported (page-based, offset-based and cursor-based) or only 1 or 2 of them, it seems (arguably) the page-based is the most commonly used one. What were your plans here?

Wondering if it's something that should be handled on serializer-only level,  seems more natural to have it on route handler level which is closer to what actually happens when interacting with real backend.

Thoughts?

samselikoff:
I haven't thought much about this one yet. I believe when I opened it I thought there was only one way to do pagination in json:api, so I thought we should add it to the serializer.

Since there's not, it might make sense to try to tackle the general "meta" problem, that is, how to get metadata to the serializer layer. Then, the serializer can decide where to put it.

Azdaroth:
How about handling pagination logic in route handlers? Actually the same thing applies to ordering and filtering as well. In real world scenario all these things happen server-side, so it might make sense to reflect the same thing in route handlers.

For the meta data, this could probably come from route handlers, just like it comes from the backend  and then maybe passed as the 3rd argument to `serialize` method and the serializer would do the rest?

samselikoff:
Yah - would you like to propose an API? We need to think of a way to get the meta from the route handler to the serializer. Problem is, you need to be able to respond with a POJO or POJA from route handlers and have that go through. 

We might need to extend Mirage.Response, and maybe you return an instance of that that has the metadata attached to it.

Azdaroth:
Yeah, I could give it a try and see what works best in this case. If I have enough time I'll play with it this weekend.

hbrysiewicz:
Feature requests are being moved to Canny. 
[...]

PR #916, opened 2016-10-13
implement paged paginator for GET requests

Azdaroth:
I've played a bit with pagination (originally https://github.com/samselikoff/ember-cli-mirage/issues/710) and here's what I think made most sense so far.

There are parts I'm not 100% happy with, especially modifying original collection (see splice part in PagedPaginator) and just assigning meta as an attribute. Maybe some decorator for a collection would be useful to make it clear that it may contain meta, having meta directly on collection seems a bit odd. And there's no easy yet to inject custom paginator to route handler.

The current pagination and some naming is a bit arbitrary (page[number] and page[size] were suggested in JSONAPI docs: http://jsonapi.org/format/#fetching-pagination, total-records and total-pages seem (arguably) to be used commonly), so not sure if it's fine like that.

Thoughts?

### DIFF
--- a/addon/route-handlers/shorthands/base.js
+++ b/addon/route-handlers/shorthands/base.js
@@ -1,5 +1,6 @@
 import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
 import BaseRouteHandler from '../base';
+import PagedPaginator from 'ember-cli-mirage/utils/paged-paginator';
 
 export default class BaseShorthandRouteHandler extends BaseRouteHandler {
 
@@ -10,6 +11,7 @@ export default class BaseShorthandRouteHandler extends BaseRouteHandler {
     this.serializerOrRegistry = serializerOrRegistry;
     this.shorthand = shorthand;
     this.options = options;
+    this.paginator = options.paginator || new PagedPaginator(serializerOrRegistry);
 
     let type = Array.isArray(shorthand) ? 'array' : typeof shorthand;
     if (type === 'string') {
--- a/addon/route-handlers/shorthands/get.js
+++ b/addon/route-handlers/shorthands/get.js
@@ -30,9 +30,11 @@ export default class GetShorthandRouteHandler extends BaseShorthandRouteHandler 
         return model;
       }
     } else if (this.options.coalesce && request.queryParams && request.queryParams.ids) {
-      return modelClass.find(request.queryParams.ids);
+      let collection = modelClass.find(request.queryParams.ids);
+      return this.paginator.paginate(request, collection);
     } else {
-      return modelClass.all();
+      let collection = modelClass.all();
+      return this.paginator.paginate(request, collection);
     }
   }
 
@@ -64,5 +66,4 @@ export default class GetShorthandRouteHandler extends BaseShorthandRouteHandler 
 
     return modelClasses.map((modelClass) => modelClass.all());
   }
-
 }
--- a/addon/serializer-registry.js
+++ b/addon/serializer-registry.js
@@ -84,4 +84,8 @@ export default class SerializerRegistry {
     );
   }
 
+  serializeMetaForPagination(totalPages, totalRecords, response) {
+    let serializer = this.serializerFor(response.modelName);
+    return serializer.serializeMetaForPagination(totalPages, totalRecords);
+  }
 }
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -45,6 +45,9 @@ class Serializer {
 
     } else if (primaryResource) {
       let [ resourceHash, newIncludes ] = this.getHashForPrimaryResource(primaryResource);
+      if (primaryResource.meta) {
+        resourceHash.meta = primaryResource.meta;
+      }
       let newDidSerialize = (this.isCollection(primaryResource) ? primaryResource.models : [ primaryResource ]);
 
       return this.buildPayload(undefined, newIncludes, newDidSerialize, resourceHash);
@@ -432,6 +435,13 @@ class Serializer {
     }
 
     return formattedAttrs;
+  }
+
+  serializeMetaForPagination(totalPages, totalRecords) {
+    return {
+      'total-pages': totalPages,
+      'total-records': totalRecords
+    };
   }
 }
 
--- a/addon/server.js
+++ b/addon/server.js
@@ -108,6 +108,26 @@ function extractRouteArguments(args) {
   return args;
 }
 
+class RouteHandlerOptions {
+  constructor(options = {}) {
+    this.options = options;
+  }
+
+  optionsForRouteHandler(method, path) {
+    let { options } = this;
+    return options[method] && options[method][path] || {};
+  }
+
+  mergeOptionsForRouteHandler(method, path, overrides = {}) {
+    let { options } = this;
+    options[method] = options[method] || {};
+    options[method][path] = options[method][path] || {};
+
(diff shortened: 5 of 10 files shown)


Azdaroth:
Failing specs on CI, but 100% green on local machine, cool ;). Will fix it with other potential enhancements / updates after review.

ashrafhasson:
:+1:

ashrafhasson:
this is a great addition, thanks!
How would you go about manually paging through the result set and returning a serialised stub if one wanted to override the default shorthand route handler, say in a test?

Azdaroth:
Hmmm, I'm not sure I get it, could you give a code example?

ashrafhasson:
So I have an acceptance-test for a controller of a route that has ember-infinity model loader and wanted to test that setting a filter on the controller will cause the ember-infinity call to the paginated endpoint includes the filter. It works but I'm not entirely sure how to test that. ember-infinity loads more data when it has reached the the bottom of the viewport. I wanted to override the route handler and inspect the queryParams has the filter on the second request to the API. Here's what I have in the controller:

```javascript
    selectedCoonditionsFilter: Ember.observer('selectedUseConditions.@each', function() {
        let filteredVehicles = [];
        let filterConditions = [];

        this.get('selectedUseConditions').forEach((condition) => {
            if (this.get('model')) {
                this.get('model').map((vehicle) => {
                    if (Ember.get(vehicle, 'useCondition') === condition.value) {
                        filteredVehicles.pushObject(vehicle);
                        filterConditions.pushObject(condition.value);
                    }
                });
            }
        });

        this.send('set_query_params', {useConditions: filterConditions});

        if (this.get('selectedUseConditions').length) {
            this.set('filteredVehicles', filteredVehicles);
        } else {
            this.set('filteredVehicles', this.get('model'));
        }
    }),

```

The route receives the filter object and sets the bound param for ember-infinity like so:

```javascript
import Ember from 'ember';
import InfinityRoute from 'ember-infinity/mixins/route';

export default Ember.Route.extend(InfinityRoute, {
    perPageParam: "page[size]",
    pageParam: "page[number]",
    totalPagesParam: "meta.total-pages",

    model() {
	    return this.infinityModel("vehicle", {perPage: 10, startingPage: 1}, {useConditions: "useConditions"});
	  },
    infinityModelUpdated() {
		    // fired up when data is loaded into the model.
		    console.log('updated with more vehicles');
	  },
    infinityModelLoaded() {
		    // fired up when there's no more data to load from the backend.
		    console.log('no more vehicles to load');
    },

    useConditions: null,

    actions: {
        set_query_params(filter) {
            for (var key in filter) {
                if (filter.hasOwnProperty(key)) {
                    this.set(key, filter[key]);
                }
            }
        }
    }
});

```

In the test, I try to assert the queryParam is present:

```javascript
test('setting useCondition causes ember-infinity to include filter as bound query param in request', function(assert) {
    server.createList('vehicle', 10);
    server.createList('vehicle', 1, {useCondition: 'new'});

    let serializerStub = {
        serializeMetaForPagination(totalPages, totalRecords, collection) {
            return {
                'total-pages': totalPages,
                'total-records': totalRecords
            };
        }
    };

    let paginator = new PagedPaginator(serializerStub);

    let requestCount = 0;
    server.get('/vehicles', (schema, request) => {
        let paginatedCollection = paginator.paginate(request, schema.vehicles.all());

        if (requestCount > 0) {
            assert.equal(request.queryParams.useConditions, ['new']);
        }

        requestCount++;
        return paginatedCollection;
    });

    visit('/search');
    click('.option-expander');
    andThen(() => assert.equal(find('.vehicle-post').length, 10, 'finds 10 vehicles prior to filtering'));

    click('.use-condition-checkbox:first');
    triggerEvent('#test-this', 'scroll', [{isInTestEnvironment:test}]).then(() => {
        console.log('here');
    });
    pauseTest();
    andThen(() => {
        assert.equal(find('.vehicle-post').length, 1, 'finds 1 vehicle post-filtering');
    });
});
```

to be honest, I don't know if this is correct or the best way for testing. It's currently failing the last assert and I feel it's because the page need to scroll to the bottom to bring the infinity-loader into the viewport.

Anyways, this might be more suited for a SO question but any comment on this would also be appreciated.
Thanks,

Azdaroth:
Hmm, not sure if I can help much here, I've never used ember-infinity, but based on what it does I wouldn't really make any extra assertions in route handlers and checking query params as it is more of an implementation detail, I'd rather make some integration / acceptance testing, so I'd write one test for the initial content, then somehow force the scrolling and check if some extra elements were added to the page by ember-inifinity, let me know if it makes sense to you.

samselikoff:
This is a great start and thanks for kicking off the work!

Your implementation here will definitely work for some folks but I'd like to think a bit about how we can get general lower-level abstractions in. If we merge this first then work towards lower-level abstractions it could make it hard to undo some of these API decisions. Since Mirage is such a general-purpose tool (aims to support all kinds of backends), it's important to give folks an escape hatch with lower-level primitives.

I love the idea of a paginator, I think this is a great abstraction, and I like the idea of being able to write your own.

I think the fundamental decision we need to make here is, how to attach metadata to a response.

samselikoff:
So I think before we make progress here, we should work on the metadata question.

I've opened https://github.com/samselikoff/ember-cli-mirage/issues/956 to discuss. Feel free to propose an API there, I will also be thinking about this soon.

ashrafhasson:
@Azdaroth thanks for your reply and yes that's what I was attempting as well, if you see in the test's second `andThen`. Agreed, this might be an implementation detail but what if I needed to override the paginator to control what's returned based on the query param (which actually is also the case in my test). As per the current implementation, that's probably possible by overriding the route, I thought.

Azdaroth:
@ashrafhasson Awh, unfortunately there's is no way to inject custom paginator at this moment, in `BaseShorthandRouteHandler` you could pass it as an option, but I haven't figured out it yet how it can be controlled from the route handler level. Check the code where the paginator is instantiated, maybe you will come up with some nice idea how to handle it.

Azdaroth:
@samselikoff I was thinking about some way to customize paginator and later to customize ordering  / filtering processors (I think there were some plans to add those as well), so I added RouteHandlerOptions, would be awesome if you could check the latest commit and provide some feedback about it. If it looks good and the API for that seems nice, I'll add more tests.

Azdaroth:
Hmm, there is some nasty inter-dependency going on in the tests... basically `TestPaginator` is injected in only one of the tests, but it looks like it sticks in other tests as well, even though the nest instance of server is created in each of them :/. No idea so far how it is possible.

Azdaroth:
Wow, that was some nasty mutation going on, gotta watch out how  `_assign` is used :P.

samselikoff:
oof. Good catch, almost seems like we could lint for that (always produce new object instead of augmenting old ones).

jfo84:
What's the status of this? Great work on this repo, but we need first class pagination to use the library. I can help if help is wanted

acorncom:
@jfo84 if you're still up for it, I think we'd all love to see good pagination support :+1:

scottkidder:
I'm interested in adding pagination support here as well. Any other work on this feature?

samselikoff:
@scottkidder I believe this PR is pretty old and was also a bit too aggressive...

What's the use case you're looking for? (What's the server pagination conventions you're using right now?)

samselikoff:

Issue #915, opened 2016-10-12
"You created a second Pretender instance while there was already one running"


denchen:
I started getting this error today with `ember-cli-mirage@0.2.2`, and I'm fairly certain it has to do with [Pretender 1.4.0](https://github.com/pretenderjs/pretender/commits/1.4.0) (that has a [Notify on second Pretender fix](https://github.com/pretenderjs/pretender/commit/5b95cd70c529f7355b01c5f9247a9f1873473683))  that was released today.

```
beforeEach failed on visiting /myroute/foo: You created a second Pretender instance while there
was already one running. Running two Pretender servers at once will lead to unexpected
results!Please call .shutdown() on your instances when you no longer need them to respond.
```

This was in my acceptance test. I have failures in my integration tests as well, which look like:

```
Promise rejected before it renders: Assertion Failed: You cannot use the same root element
(#ember-testing) multiple times in an Ember.Application
```

So what is the appropriate way to shutdown the Mirage server? I currently have this in my acceptance tests:

```
afterEach() {
  Ember.run(application, 'destroy');
  server.shutdown();
}
```

Do I need something similar in my integration tests?

denchen:
I found a workaround for both my acceptance and integration tests

In acceptance tests, my `beforeEach()` now looks like:

      beforeEach() {
        server.shutdown();
        application = startApp();
     },

I'm not sure why `server.shutdown()` in `afterEach()` doesn't properly shutdown the server.

And for my integration/unit tests, I've modified the code from [Manually starting up Mirage](http://www.ember-cli-mirage.com/docs/v0.2.x/manually-starting-mirage/) to this:

    // See: http://www.ember-cli-mirage.com/docs/v0.2.x/manually-starting-mirage/
    import mirageInitializer from '../../initializers/ember-cli-mirage';
    
    export default function startMirage(container) {
      if (this.server) {
        console.warn('Shutting down Mirage server before re-initializing');
        this.server.shutdown();
      }
      mirageInitializer.initialize(container);
    }

All my tests now pass. I'm not sure if the above modifications are what I should be doing or not, however.

netes:
Same here.

netes:
**Update:**
It seems that updating `module-for-acceptance.js` to the newest version (from Ember CLI 2.8) and adding this snippet to integration tests which use Mirage:
```js
    afterEach() {
      server.shutdown();
    },
```
does the job - everything is fine then.

dustinfarris:
[...]

PR #921, opened 2016-10-17
fix pretender version

Leooo:
@samselikoff https://github.com/samselikoff/ember-cli-mirage/issues/915 happens to be a big problem with us, I think https://github.com/pretenderjs/pretender/pull/178 should have been a major version bump. This PR to fix the pretender version to a previous one until things settle down.

Thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-cli-mirage",
-  "version": "0.2.2",
+  "version": "0.2.3",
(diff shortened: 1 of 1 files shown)


cspanring:
This is the workaround we are using to fix the pretender issue in our application. Would be good to issue an "official" mirage patch. Thanks!

Leooo:
Closing as Pretender just released a minor version which solves the problem.

L

PR #924, opened 2016-10-18
Add support for promises in route functions

dustinfarris:
This enables users to work with promises in route function handlers, e.g.:

```js
this.get('/users', () => {
  return new Promise(resolve => {
    resolve(new Response(200, { 'Content-Type': 'text/csv' }, 'firstname,lastname\nbob,dylan'));
  });
});
```


dustinfarris:
Bumped min pretender to 1.4.1 which does not throw errors on overlapping pretender servers.

samselikoff:
Looks great!

PR #925, opened 2016-10-18
[DOCS] Advise user to shutdown pretender servers

### LINKED ISSUE #915: "You created a second Pretender instance while there was already one running"

dustinfarris:
Updating old documentation.  I'm not sure why the original docs advised running the initializer method; the startMirage function should work just fine—no helper file should be necessary.

Closes #915 (not automatically, though)

### DIFF
--- a/docs/v0.2.x/manually-starting-mirage.md
+++ b/docs/v0.2.x/manually-starting-mirage.md
@@ -5,27 +5,19 @@ version: v0.2.x
 
 Currently, Mirage's server boots up in an Ember initializer. This means it will not automatically start in unit or integration tests, since there's no Ember app running.
 
-Eventually we'll extract Mirage's initializer, but for now you can use this workaround. To run your Mirage server during a unit or integration test, first create a helper:
-
-```js
-// tests/helpers/setup-mirage-for-integration.js
-import mirageInitializer from '../../initializers/ember-cli-mirage';
-
-export default function startMirage(container) {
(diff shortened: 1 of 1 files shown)


samselikoff:
The initializer method code was taken from a @rwjblue gist... Robert, is it ok to just use

```
import { startMirage } from 'yourapp/initializers/ember-cli-mirage'
```

and call it without passing in the container?

PR #927, opened 2016-10-20
fix(config-docs): document requirement to define namespace in cordova/electron envs

anulman:
Per #497


samselikoff:
Awesome - super helpful. Thank you!

Issue #905, opened 2016-09-28
Relation is serialized to the model name instead of the attribute name


jessedijkstra:
When I upgraded from 0.2.1 I noticed that all of the sudden certain relationships with on the wrong key. It uses the model key instead of the attribute key.

I defined the relationship with `include` on a model that is serialised with the `ActiveModelSerializer` serialiser as followed:

``` js
import ApplicationSerializer from './application';

export default ApplicationSerializer.extend({
  include: ['author']
});
```

The model includes the relationship as followed:

``` js
export default Model.extend({
  author: belongsTo('user'),
  user_answers: hasMany('user_answer'),
});
```

The factory sets the author with afterCreate as followed:

``` js
  const { user_answers, author } = user_question;

  if (!author) {
    user_question.author = server.create('user');
  }
```

The returned JSON with 0.2.1 stores the relation in the author key:
<img width="560" alt="screen shot 2016-09-27 at 10 44 11" src="https://cloud.githubusercontent.com/assets/226130/18866293/a5fa1048-849f-11e6-90bc-c2c7eac0fe9e.png">

In 0.2.2 the relation is stored in the user key:
<img width="559" alt="screen shot 2016-09-27 at 10 45 42" src="https://cloud.githubusercontent.com/assets/226130/18866300/aaaf649e-849f-11e6-88e1-da6160c095b2.png">

jessedijkstra:
@samselikoff any information on this? It's a breaking bug.

aquamme:
I've encountered this problem too. I ended up just renaming my model as a workaround.

saheljalal:
In the `_hashForModel` method in `serializer.js` it seems the attribute key isn't used when embedding resources.

```javascript
if (this.embed) {
  let newDidSerialize = _assign({}, didSerialize);
  newDidSerialize[model.modelName] = newDidSerialize[model.modelName] || {};
  newDidSerialize[model.modelName][model.id] = true;

  this.getKeysForIncluded().forEach(key => {
    let associatedResource = model[key];
    if (!_get(newDidSerialize, `${associatedResource.modelName}.${associatedResource.id}`)) {
      let [ associatedResourceHash ] = this.getHashForResource(associatedResource, true, newDidSerialize, true);
      let formattedKey = this.keyForResource(associatedResource);
      attrs[formattedKey] = associatedResourceHash;

      if (this.isModel(associatedResource)) {
        let fk = `${camelize(associatedResource.modelName)}Id`;
[...]

PR #930, opened 2016-10-21
Serialize models on the relation key instead of the modelName

jessedijkstra:
The models of a relations with serialized on the modelName. This makes it impossible to correctly serialize the relations.

See https://github.com/samselikoff/ember-cli-mirage/issues/905

**Todo:**
- [x] Fix and test the json-serializer
- [x] Fix and test the active-model-serializer

### DIFF
--- a/tests/integration/serializers/base/associations/custom-key-collection-test.js
+++ b/tests/integration/serializers/base/associations/custom-key-collection-test.js
@@ -1,0 +1,157 @@
+// jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+import Serializer from 'ember-cli-mirage/serializer';
+import { hasMany, belongsTo } from 'ember-cli-mirage';
+import Schema from 'ember-cli-mirage/orm/schema';
+import Model from 'ember-cli-mirage/orm/model';
+import Db from 'ember-cli-mirage/db';
+import SerializerRegistry from 'ember-cli-mirage/serializer-registry';
+import { module, test } from 'qunit';
+
+module('Integration | Serializers | Base | Associations | Custom Key', {
+  beforeEach() {
+    let db = new Db();
+    this.schema = new Schema(db);
+    this.schema.registerModels({
+      wordSmith: Model.extend({
+        draftBlogPosts: hasMany('blogPost', { inverse: 'wordSmithee' })
+      }),
+      blogPost: Model.extend({
+        wordSmithee: belongsTo('wordSmith')
+      }),
+    });
+
+    let link = this.schema.wordSmiths.create({ name: 'Link', age: 123 });
+    link.createDraftBlogPost({ title: 'Lorem' });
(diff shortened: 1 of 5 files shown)


jessedijkstra:
Saw that the serializer has been rewritten in it's entirety again. Rebasing now.

jessedijkstra:
@samselikoff Rebased and fixed

jessedijkstra:
@samselikoff Now fixed all the serialisation issues in the implemented serialisers. Would love to get some feedback on this PR with possible solutions that are less dependent on storing the key along with the resource when serialising associations.

This is quite a breaking bug and completely stops us from upgrading to the newest version of Mirage.

samselikoff:
I think I'd rather not change all these method signatures, I know we spoke about making this code more clear overall but for this PR, could we just make the minimum change needed to get the tests passing?

Also if you want I can do this this weekend - your tests are valuable so let's keep them here.

jessedijkstra:
@samselikoff Yeah, was looking for the minimum change. The biggest problem is that we loose the key the model is serialised on the moment we enter `getHashForIncludedResource`.

If you have multiple keys with the same model it's impossible to know what the key the resource should have based on the model definition.

Should we add the key to the resource rather then adding it as a `{ key, resource ]` pair? That does however require a change where we deepClone the model because it introduces resource depended state to the model making it unreliable and unsafe.

tests/integration/serializers/base/associations/sideloading-model-test.js
```diff
@@ -91,7 +91,7 @@ test(`it can sideload a named has-many association`, function(assert) {
       name: 'Link',
       postIds: ['1']
     },
-    blogPosts: [
+    posts: [
```
samselikoff: I think this is wrong. Top-level collections should use the model name - what if this `blogPost` is also related to another model?

Attribute keys should only affect their model

samselikoff:
I'm working on this now, I think I'll just push my changes up to your branch :)

tests/integration/serializers/base/associations/custom-key-collection-test.js
```diff
@@ -0,0 +1,157 @@
...
+  let blogPost = this.schema.blogPosts.find(1);
+  let result = registry.serialize(blogPost);
+
+  assert.deepEqual(result, {
+    blogPost: {
+      id: '1',
+      title: 'Lorem',
+      wordSmitheeId: '1'
+    },
+    wordSmithees: [
```
samselikoff: This is also wrong

samselikoff:
I think the bug is only for embedded, but you changed sideloaded as well.

Going to clean up the tests because I try to avoid adding new tests if we don't need them.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/commit/1e4a5ef6cecf37cf6976328bb98705b4bee1d7d6 closes - thank you Jesse!

PR #932, opened 2016-10-25
Let faker float from 3.0.0

Dhaulagiri:
This PR addresses concerns from #863 around a lack of flexibility around which version of faker is used.  This PR lets the range float from 3.0.0 so by default users will get the latest version, but if they want to use a specific version they can specify that version in their own package.json.

I tested this out a few ways:
 * No Faker version in my app's package.json, I get Faker 3.1.0
 * Faker 3.0.1 in my app's package.json, I get Faker 3.0.1
 * Faker ^3.0.0 in my app's package.json, I get Faker 3.1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ember-lodash": "0.0.9",
     "exists-sync": "0.0.3",
     "fake-xml-http-request": "^1.4.0",
-    "faker": "3.0.1",
+    "faker": "^3.0.0",
     "pretender": "^1.4.1",
     "route-recognizer": "^0.1.11"
   }


samselikoff:
Nice, should it just be `~3` or `~3.0`? Should we leave out patch version? (I'm not sure here, just looking for reasoning...)

Dhaulagiri:
@samselikoff I'm always braced to be trolled by some part of npm I don't understand, but `^3.0.0` should consistently give people the latest version of the `3.x` series of Faker with the option of specifying an older version themselves.  `~3` seems to yield the same in my local testing whereas `~3.0` installs `3.0.1`.

samselikoff:
👍  awesome, thank you!

Dhaulagiri:
@samselikoff we should mention in the changelog or docs how this works.

samselikoff:
Changelog sounds like the right place for starters. Would you have time to add a line w/explanation?

Dhaulagiri:
@samselikoff sure, where's the best place to put this? Start a new section for a 0.2.4 release?

samselikoff:
I usually start with `#master` heading, in case it ends up being a minor bump.

PR #936, opened 2016-10-28
Add regression tests for reflexive operations.

abuiles:
@samselikoff based on our chat in slack


samselikoff:
thank you. Should be covered by 0-3 series.

PR #941, opened 2016-11-12
HasMany association on its own model fix

designblooz:
I ran into a problem with ember-cli-mirage >= v0.2.2 when a model has a hasMany relationship to its own model that references to a its own belongsTo relationship with a foreignKey. It was due to `fksAddedFromThisModel` already including fk on the inverse relationship since it was referencing to a belongsTo relationship on its own model. So the fix was by adding `foreignKey` option to the hasMany association and checking if `fk !== assocation.opts.foreignKey` before trying to check if current `fk` is included in the `fksAddedFromThisModel[fkHolder]`.

Example of the model structure I had issue with is added in the acceptance test.

### DIFF
--- a/addon/orm/schema.js
+++ b/addon/orm/schema.js
@@ -68,13 +68,17 @@ export default class Schema {
         let [ fkHolder, fk ] = association.getForeignKeyArray();
 
         fksAddedFromThisModel[fkHolder] = fksAddedFromThisModel[fkHolder] || [];
-        assert(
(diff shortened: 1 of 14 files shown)


samselikoff:
Thank you. This should be addressed in the 0-3 series, you can read more [here](http://www.ember-cli-mirage.com/blog/2017/01/09/0-3-0-beta-series/).

Let me know if you run into any more issues related to this!

PR #947, opened 2016-11-19
server.contacts => server.schema.contacts

mfazekas:
contacts.length => contacts.models.length
```contacts.isAdmin = false``` => ```contacts.update({ isAdmin: false })```

see also #948


samselikoff:
Thanks 👊

PR #948, opened 2016-11-19
Collection#length

mfazekas:


samselikoff:
I'm going to merge this bc it's so convenient - we used to have it - but my only hesitation is if it will trick people and/or libraries into thinking Collection is an array. That was the original problem we had - Collection was actually a subclass of `[]`, which nudged folks that they could use, e.g. Lodash methods on it. Some would work and then mysteriously, some wouldn't.

So let's merge this, but keep our eye out for people being confused about this. The `.models` property should still be seen as the way to get at "the JS array of models" if you need it.

Thanks!

PR #949, opened 2016-11-19
Upgrade route-recognizer to decode dyanmic path segments

mfazekas:
See https://github.com/tildeio/route-recognizer/pull/91

Right now if id contains a special character like ```/``` or ```:``` then we need to use a workaround like:

```
this.get('/posts/:id', (request, schema) {
    let decodedId = decodeURIComponent(request.params.id);
    return schema.posts.find(decodedId);
});
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "exists-sync": "0.0.3",
     "fake-xml-http-request": "^1.4.0",
     "faker": "^3.0.0",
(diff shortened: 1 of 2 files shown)


mfazekas:
There might be a more appropriate location for the tests.

samselikoff:
Thank you! ✨

PR #954, opened 2016-11-28
Typo fix: Remove errant "i" from subdirectory msg

bantic:
This removes the letter "i" from the message. It appears to be a typo that snuck in. (A previous version of this file doesn't have the i: https://github.com/samselikoff/ember-cli-mirage/blob/f95aa53262164953f70c970812d8462aa13feab2/addon/utils/read-modules.js#L32).


samselikoff:
thank you sir!

PR #957, opened 2016-12-04
Add slice method to collections

alexparker:
Adds slice method to collections. If someone else gets to a solution for proxying enumerable methods to Array and lodash first, great. But this would be helpful too.

### DIFF
--- a/addon/orm/collection.js
+++ b/addon/orm/collection.js
@@ -94,6 +94,19 @@ export default class Collection {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


samselikoff:
We originally had Collection extend `[]`, but that cause problems (bc you can't really do that), but now that it's not a fake subclass I think it makes sense to start adding these methods back in.

Maybe we can look at how Backbone collections proxy methods to lodash.

Issue #885, opened 2016-09-15
Creating an empty object rejected as invalid JSON-API payload (type, but no attrs/rels)


nruth:
I think there's a bug with creating empty objects. We currently have some code that does this, then creates other objects associated to it using the ID given back by the API.

`this.get('store').createRecord('something', {})` gives me

`ember.debug.js:19746 Mirage: Error: You're using a shorthand or #normalizedRequestAttrs, but your serializer's normalize function did not return a valid JSON:API document.` 

https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/route-handlers/base.js#L45
is asserting json.data.attributes or relationship must be present, but json-api says only type is necessary for post/create requests. http://jsonapi.org/format/#crud-creating

Ember Data is sending it to faker as something like `requestBody: "{"data":{"type":"somethings"}}"` (found by dropping a `debugger` into the library code in node_modules).

I've been getting this on 0.2.1 and master.

Changing the assert to `json.data && (json.data.type || json.data.attributes || json.data.relationships)` fixes our app, but seems like a bad solution. It should probably check if it's a post/create or other kind of request?

acorncom:
@nruth Would it be possible to get a failing test case for this?  Sam's been getting swamped, so trying to figure out how to have us all help him help us ;-)

samselikoff:
I agree with you - I think we need to just change that check to match for `json.data && json.data.type`, based on what you've found in the spec.

Would you PR this?

nruth:
Yes I will try to add tests and make the change, but need to clear another deadline first. Please poke me if I haven't submitted something by mid November.

I think `{ data: { type: '' }}` is only valid if it's a create action, the others need more, but will double-check when I get back to it.

samselikoff:
Looking for someone to take this one on.

alisdair:
If I'm reading the spec properly, `{"data":null}` would be a valid document. This is explicitly recommended in [the section about clearing a one-to-one relationship](http://jsonapi.org/format/#crud-updating-to-one-relationships):

> And the following request clears the author of the same article:
> 
>     PATCH /articles/1/relationships/author HTTP/1.1
>     Content-Type: application/vnd.api+json
>     Accept: application/vnd.api+json
>    
>     {
>       "data": null
>     }

And in [the top-level document description](http://jsonapi.org/format/#document-top-level):

> A document **MUST** contain at least one of the following top-level members:
> - `data`: the document’s “primary data”
>
[...]

PR #958, opened 2016-12-04
Adds single test case to allow payload with type

alexparker:
Opening this up re: https://github.com/samselikoff/ember-cli-mirage/issues/885 I'm not sure it's sufficient enough - I don't have enough context of the entire codebase, but wanted to at least get something started to help out.


samselikoff:
Definitely an improvement - thank you!

PR #966, opened 2016-12-09
include note about import path in an Ember addon

noslouch:
This bit me on a recent project. `import 'yourapp/initializers/ember-cli-mirage'` is not a valid path inside an ember addon. Needs to be:

```js
import { startMirage } from 'dummy/initializers/ember-cli-mirage';
```


samselikoff:
thank you!

PR #981, opened 2017-01-11
clarify error message when using 'association' helper for nonexistent relationship

Azdaroth:


samselikoff:
How about

```js
`You're using the `association` factory helper on the '${associationAttribute}' attribute of your ${modelType} factory, but that attribute is not a `belongsTo` association. Read the Factories docs for more information.`
```

I'd love to get a permalink to different parts of docs that's versionless, so we could link right to it.

Azdaroth:
Sounds much better, I like it 👍 

I don't think it's possible currently to get a permalink like that, but `http://www.ember-cli-mirage.com/docs/v0.2.x/factories/#factories-and-relationships` should be good enough here, at least for now.

samselikoff:
rebuilding master to see if the problem is there

Azdaroth:
@samselikoff All tests are passing now, sounds good to merge?

samselikoff:
Yessir go for it

Azdaroth:
💪

PR #984, opened 2017-01-16
[BUGFIX] Blueprint.prototype.insertIntoFile is async

stefanpenner:
This prevents the termination of the process and updating these files from racing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
 before_install:
   - export PATH=/usr/local/phantomjs-2.0.0/bin:$PATH
   - "npm config set spin false"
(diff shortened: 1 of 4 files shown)


stefanpenner:
@samselikoff i believe tests are failure due to node 0.12 stuff, which is EOL. Should i bump node, or move from `=>` to old schol functions + `this` or `bind`

samselikoff:
@stefanpenner bump node would be perfect

stefanpenner:
failing tests appeared to be a bug in ember or ember-data 2.3, bumped to a more recent version and everything appears fine locally. Lets see if CI confirm.

samselikoff:
👏  thank you so much man

PR #780, opened 2016-06-17
Adds support for subdirectories when reading modules

jrhe:
Should fix #663 #540. Needs testing but wanted some guidance on what the test cases should be and where they should go to suit the project. All existing tests are green.

PR #992, opened 2017-01-19
implementing subdirectories - Fixes #540

### LINKED ISSUE #540: Subdirectories are not supported
When running:

 `ember g mirage-model foo/bar`

The following error occurs when trying to access the route:

`Assertion Failed: Subdirectories under foo are not supported`

A related question:

How would you see Pods structure working with ember-cli-mirage?

Thanks!

mansona:
Howdy 👋 

I ran across the same issue described in https://github.com/samselikoff/ember-cli-mirage/issues/540 and have essentially attempted to implement the tests requested in the closed pull request https://github.com/samselikoff/ember-cli-mirage/pull/780 (the implementation is almost exactly the same as @jrhe initially implemented)

This essentially works but if you take a look at the mirage config you can see that I have had to do some pretty awful stuff to make it work 😭 

I'm not entirely sure where to go from here. Ideally, I would like to use the cleaner helper API for the config like this `this.get('/user/profiles');` and for it to *"know"* that it is supposed to use the 'user/profiles' model. 

Any suggestions you have on how to proceed would be appreciated.

### DIFF
--- a/addon/utils/read-modules.js
+++ b/addon/utils/read-modules.js
@@ -5,8 +5,7 @@
 'use strict';
 
 import Ember from 'ember';
-import _camelCase from 'lodash/string/camelCase';
-import { pluralize } from 'ember-cli-mirage/utils/inflector';
+import { pluralize, camelize } from 'ember-cli-mirage/utils/inflector';
 
 const { assert } = Ember;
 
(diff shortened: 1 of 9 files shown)


brunowego:
Hi @mansona, thanks for this great feature. You will continue this work?

mansona:
@brunowego I'm happy to keep working on this but I think I will need some guidance to continue. I can't quite remember but I think I had a conversation with @samselikoff on the Slack channel and he mentioned that this wasn't priority ahead of the 0.3.0 release

Although I could be taking that up completely wrong 😖

shanebruggeman:
Hi @mansona, now that 0.3.0 is out is there more priority for this? Thanks!

mansona:
@shanebruggeman I get the feeling this whole thing will become obsolete, or handled in a **very** different way when the module unification PR/RFC lands. 

I think for now we ought to close this and if it comes up again we can look at it then. For my particular use case, the client I was working with has flattened the directory structure of the models so it no longer affects me.

PR #993, opened 2017-01-20
Good Enough™ `hasAndBelongsToMany`

chancancode:
Hello!

I implemented a Good Enough™ `hasAndBelongsToAny` that seems to work for my app.

### Usage

```javascript
// mirage/models/app.js
export default Model.extend({
  owners: hasAndBelongsToMany('user', { inverse: 'ownedApps' }),
  collaborators: hasAndBelongsToMany('user', { inverse: 'sharedApps' })
});

// mirage/models/user.js
export default Model.extend({
  ownedApps: hasAndBelongsToMany('app', { inverse: 'owners' }),
  sharedApps: hasAndBelongsToMany('app', { inverse: 'collaborators' })
});
```

### API

```javascript
app.owners; // => [user1, user2, ...]
app.owners = [user1, user2, ...];
app.ownerIds; // => [userId1, userId2, ...];
app.ownerIds = [userId1, userId2, ...];

app.addOwner(user);
app.addOwnerId(userId);
app.addOwners(user1, user2, ...);
app.addOwnerIds(userId1, userId2, ...);

app.removeOwner(user);
app.removeOwnerId(userId);
app.removeOwners(user1, user2, ...);
app.removeOwnerIds(userId1, userId2, ...);

// Called without arguments removes everything... bad iDear?
app.removeOwners();
app.removeOwnerIds();
```

### Demo app

https://github.com/tildeio/mirage-habtm-demo/

### Implementation

It works by generating an implicit join model for each habtm relationship (a single model is shared by the two sides). The join model simply stores the IDs of the two sides as plain attributes (not FKs) and therefore does not leak the join model into the other classes.

#### Does

✅ Seem to work
✅ Seem good enough
✅ Require both sides to be saved (not `isNew()`) in order for it to work
✅ Have HAX

#### Does not

❎ Cache
❎ Have tests
❎ Handle attaching unsaved parent/children (currently an error)
❎ DRY

#### Probably

❓ Have bugs
❓ Does not scale

I probably don't have time nor the expertise to polish this enough to be mergable, but I figured this would be a useful starting point for those who do.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -11,12 +11,16 @@ import JSONAPISerializer from './serializers/json-api-serializer';
 import RestSerializer from './serializers/rest-serializer';
 import HasMany from './orm/associations/has-many';
 import BelongsTo from './orm/associations/belongs-to';
+import HasAndBelongsToMany from './orm/associations/has-and-belongs-to-many';
 
 function hasMany(...args) {
   return new HasMany(...args);
 }
 function belongsTo(...args) {
   return new BelongsTo(...args);
+}
+function hasAndBelongsToMany(...args) {
+  return new HasAndBelongsToMany(...args);
 }
 
 export {
@@ -32,12 +36,14 @@ export {
   JSONAPISerializer,
   RestSerializer,
   hasMany,
-  belongsTo
+  belongsTo,
+  hasAndBelongsToMany
 };
 
 export default {
   Factory,
   Response,
   hasMany,
-  belongsTo
+  belongsTo,
+  hasAndBelongsToMany
 };
--- a/addon/orm/associations/has-and-belongs-to-many.js
+++ b/addon/orm/associations/has-and-belongs-to-many.js
@@ -1,0 +1,308 @@
+import Association from './association';
+import Collection from '../collection';
+import Model from '../model';
+import _compact from 'lodash/array/compact';
+import { capitalize, camelize, singularize } from 'ember-cli-mirage/utils/inflector';
+import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
+import assert from 'ember-cli-mirage/assert';
+
+function diffIds(oldIds, newIds) {
+  oldIds = oldIds.sort();
+  newIds = newIds.sort();
+
+  let add = [];
+  let remove = [];
+
+  let i = 0;
+  let j = 0;
+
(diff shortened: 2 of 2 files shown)


samselikoff:
Very cool!

Have you seen http://www.ember-cli-mirage.com/blog/2017/01/09/0-3-0-beta-series/ by the way? Originally I was going to add habtm but ended up changing the ORM's fk implementation. Many-to-many on 0-3 series is now working.

chancancode:
:clap: we will investigate that as well!

dylankb:
We tried upgrading to Mirage 0.3.0, but we had difficulty getting many-to-many relationships working. The problem seems to occur when there are multiple `hasMany` relationships between two models, but I'm not sure.

I forked one of your twiddles to demonstrate the problem, and here is the [Gist output](https://gist.github.com/dylankb/13dd2b66a6290bf0f502e37096743391/revisions). Sorry for all the revisions, just pay attention to the last three. Also here's the [twiddle](https://ember-twiddle.com/13dd2b66a6290bf0f502e37096743391?openFiles=mirage.models.user.js%2C).

The third revision from the most recent revision currently works. `users` have many `apps`, and `apps` have many `users`.

The second revision also introduces `ownedApps` to `users` and `owner` to `app`. This is where the relationship stops working correctly, and the JSON shows the ids are no longer updating.

The current version (ACA5E30) adds a piece of owner test data, but this does not update the `user` side with any `owner` ids.

Commenting out line 6 on `/mirage/models/user.js` shows the `app` & `user` relationship correctly working again.

dylankb:
We did some more investigating and isolated the problem further. It seems like the problem is when there are multiple relationships of any kind (`hasMany/belongsTo`) of the **same type**; it fails to find the right inverse, thus failing to keep the two sides in sync.

Here is the [twiddle](https://ember-twiddle.com/13dd2b66a6290bf0f502e37096743391/28def03624902e426de2d08f5d1245c9572bff00?openFiles=mirage.models.user.js%2C)

samselikoff:
Closing this for now, I think double-sided hasMany means we don't need habtm (ember data style).

@dylankb tracking your bug in https://github.com/samselikoff/ember-cli-mirage/issues/1004

samselikoff:
@dylankb bug is fixed, try `0.3.0-beta.5`

chancancode:
:clap:

PR #995, opened 2017-01-23
Invoke _getAttrsForRequest with correct model name. Fixes #994

### LINKED ISSUE #994: FunctionRouteHandler calls getModelClassFromPath with wrong argument on PUT/PATCH, breaking customized normalize methods.
Given a function route handler like this:

```
this.put('/contacts/:id', function({ contacts }, request) {
  let id = request.params.id;

  return contacts.find(id).update(attrs);
});

```
And a request like `PUT /contacts/2`...

`FunctionRouteHandler.normalizedRequestAttrs()` calls `getModelClassFromPath()` with `request.url`, which is `'/contacts/2'`. `getModelClassFromPath()` then returns `'2'` as the model class, which is obviously incorrect. The effect of this is that model-specific normalizers are not run.

The fix is simple: Call `getModelClassFromPath()` with the route path, instead of the request URI.

PR to follow.

bwbuchanan:


samselikoff:
Going to go ahead and merge this, since normalizedRequestAttrs is covered the way it is used by public.

Thank you @bwbuchanan!

Issue #901, opened 2016-09-27
Make namespace/host support more robust


samselikoff:
Folks keep running into issues if their namespace/host is
- undefined
- empty string
- starts with a leading slash

or if their routes have or omit a leading slash. We should make the logic robust to all these cases.

samselikoff:
Seems there's also issues now that baseUrl is deprecated. I want to get all this stuff sorted out, would love someone's help.

backspace:
Ah, this is happening to me [here](https://github.com/travis-ci/travis-web/pull/723)! Removing `baseURL` has caused test failures that I don’t know how to fix apart from adding a leading `/` in the adapter’s namespace, which results in unsightly `//`s in production API calls. 🤔

0000marcell:
I having this issue too, if i remove baseURL the tests won't pass, if i add baseURL i get a deprecation warning, for now i'm leaving the deprecation warning, is there any better solution ?

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1006 closes

PR #1006, opened 2017-01-31
Make host/namespace support more robust

zinyando:
Fixes issue #901. 

Adds tests to reproduce issues brought up in issue #901 and makes them pass. These test for 4 conditions for different namespace and urlPrefix inputs

1. with a leading slash
2. without a leading slash
3. with a trailing slash
4. as empty strings

Any feedback is welcome.


zinyando:
@samselikoff any feedback on this? Just wondering what I need to do so that this gets merged in :)

samselikoff:
This looks great! Thanks @zinyando !

PR #1013, opened 2017-02-16
Add documentation for findBy

davewasmer:
TIL mirage has `findBy`, and I can stop writing `users.where({ username }).models[0]` 😜

### DIFF
--- a/docs/v0.2.x/models.md
+++ b/docs/v0.2.x/models.md
@@ -93,7 +93,7 @@ let posts = blogPosts.find([1, 3, 4]);
 
 ### findBy(*query*)
 
-Returns the first model in the database that matches the key-value pairs in the `query` object. Note that a string comparison is used. `query` is a POJO.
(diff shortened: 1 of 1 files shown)


docs/v0.2.x/models.md
```diff
@@ -91,6 +91,14 @@ let post = blogPosts.find(1);
 let posts = blogPosts.find([1, 3, 4]);
 ```
 
+### findBy(*query*)
+
+Returns the first model in the database that matches the key-value pairs in the `query` object. Note that a string comparison is used. `query` is a POJO.
+
+```js
+let posts = blogPosts.where({ published: true });
```
samselikoff: is it supposed to be `findBy` instead of `where`

davewasmer: Derp :man_facepalming:

locks:
Psst, what's a POJO?

samselikoff:
thanks! ❤️

PR #1014, opened 2017-02-16
Mirage as express middleware

mfazekas:
This PR adds the option to use express instead of pretender.

### Usecases: 
- use real HTTP traffic, use browser's network inspector
- let multiple instances talk to mirage backend that survives reload
- get fastboot apps running without a real backend, fastboot testing

### Implementation:

We compile `ember-cli-mirage` and app's `mirage` dir to node.js using babel and borccoli. Since `ember-cli-mirage` uses `ember` some places we have `ember-compat` that provides the used parts of `Ember` in node.js

~this is piggybacks on `Fastboot`, the main issue with is that current implementation of mirage runs in browser and we need it to run it on `node.js`. Fastboot already does that for the whole ember app. So we install an express middleware so that on first network request we can lazily create a Fastbooted (sandboxed) app instance, and we pass an express router to it. Then the sandboxed app can use the passed in express router to respond to real HTTP requests.~

~It also depends on universal serving intruduced in ember `2.12` apps needs to include `ember-cli-fastboot`, so .js assets are compiled to node.js env.~

### TODO

- tests
- reload

### How to use

1. set `useExpress: true` in your ENV['ember-cli-mirage']  in `config/environment.js`

For example app see: https://github.com/mfazekas/fastboot-mirage-test/

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -4,12 +4,12 @@ import { pluralize, camelize } from './utils/inflector';
 import { toCollectionName } from 'ember-cli-mirage/utils/normalize-name';
 import Ember from 'ember';
 import isAssociation from 'ember-cli-mirage/utils/is-association';
-import Pretender from 'pretender';
 import Db from './db';
 import Schema from './orm/schema';
 import assert from './assert';
 import SerializerRegistry from './serializer-registry';
 import RouteHandler from './route-handler';
+import createPretender from './interceptors/create-pretender';
 
 import _pick from 'lodash/pick';
 import _assign from 'lodash/assign';
@@ -19,48 +19,12 @@ import _isInteger from 'lodash/isInteger';
 
 const { RSVP: { Promise } } = Ember;
 
-/**
- * Creates a new Pretender instance.
- *
- * @method createPretender
- * @param {Server} server
- * @return {Object} A new Pretender instance.
- * @public
- */
-function createPretender(server) {
-  return new Pretender(function() {
-    this.passthroughRequest = function(verb, path, request) {
-      if (server.shouldLog()) {
-        console.log(`Passthrough request: ${verb.toUpperCase()} ${request.url}`);
-      }
-    };
-
-    this.handledRequest = function(verb, path, request) {
-      if (server.shouldLog()) {
-        console.log(`Mirage: [${request.status}] ${verb.toUpperCase()} ${request.url}`);
-        let { responseText } = request;
-        let loggedResponse;
-
-        try {
-          loggedResponse = JSON.parse(responseText);
-        } catch(e) {
-          loggedResponse = responseText;
-        }
-
-        console.log(loggedResponse);
-      }
-    };
-
-    this.unhandledRequest = function(verb, path) {
-      path = decodeURI(path);
-      assert(
-        `Your Ember app tried to ${verb} '${path}',
-         but there was no route defined to handle this request.
-         Define a route that matches this path in your
-         mirage/config.js file. Did you forget to add your namespace?`
-      );
-    };
-  });
+function createInterceptor(server) {
+  if (server.options.createInterceptor) {
+    return server.options.createInterceptor(server);
+  } else {
+    return createPretender(server);
+  }
 }
 
 const defaultRouteOptions = {
@@ -168,7 +132,7 @@ export default class Server {
     let hasFactories = this._hasModulesOfType(config, 'factories');
     let hasDefaultScenario = config.scenarios && config.scenarios.hasOwnProperty('default');
 
-    this.pretender = this.pretender || createPretender(this);
+    this.interceptor = this.interceptor || createInterceptor(this);
 
     if (config.baseConfig) {
       this.loadConfig(config.baseConfig);
@@ -255,7 +219,7 @@ export default class Server {
     verbs.forEach((verb) => {
       paths.forEach((path) => {
         let fullPath = this._getFullPath(path);
-        this.pretender[verb](fullPath, this.pretender.passthrough);
+        this.interceptor[verb](fullPath, this.interceptor.passthrough);
       });
     });
   }
@@ -409,7 +373,7 @@ export default class Server {
   }
 
   shutdown() {
-    this.pretender.shutdown();
+    this.interceptor.shutdown();
     if (this.environment === 'test') {
       window.server = undefined;
     }
@@ -481,7 +445,7 @@ export default class Server {
     let fullPath = this._getFullPath(path);
     let timing = options.timing !== undefined ? options.timing : (() => this.timing);
 
-    this.pretender[verb](
+    this.interceptor[verb](
       fullPath,
       (request) => {
         return new Promise(resolve => {
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -38,7 +38,7 @@ function _shouldUseMirage(env, addonConfig) {
   to initialize Mirage.
 */
 function _defaultEnabled(env, addonConfig) {
-  let usingInDev = env === 'development' && !addonConfig.usingProxy;
+  let usingInDev = env === 'development' && !addonConfig.usingProxy && !addonConfig.useExpress;
   let usingInTest = env === 'test';
 
   return usingInDev || usingInTest;
--- a/index.js
+++ b/index.js
@@ -4,6 +4,49 @@ var path = require('path');
 var mergeTrees = require('broccoli-merge-trees');
 var Funnel = require('broccoli-funnel');
 
+function toNodeJSBuilder(projectRoot, mirageDirectory) {
+  var esTranspiler = require('broccoli-babel-transpiler');
+  var broccoli = require('broccoli-builder');
+
+  var appEnvPath = path.join(projectRoot, 'config', 'environment.js');
+  var appEnv = require(appEnvPath);
+
+  var emberCliMiragePath = __dirname;
+  var mirageAddon = new Funnel(emberCliMiragePath, {srcDir: 'addon', destDir: 'ember-cli-mirage'});
+  var appMirageSrc = new Funnel(mirageDirectory, {srcDir:'.', destDir: 'mirage'});
+  var nodejsSrc = new Funnel(emberCliMiragePath, {srcDir:'nodejs', destDir: 'nodejs'});
+  var combinedSrcTree = mergeTrees([mirageAddon, nodejsSrc, appMirageSrc]);
+  var transpiledTree = esTranspiler(combinedSrcTree, {
+    browserPolyfill: true,
+    resolveModuleSource: function(source, filename) {
+      var result = source;
+      var result_abs;
+
+      var dirname = path.dirname(path.join(transpiledTree.outputPath,filename));
+      if (source === 'ember') {
+        result_abs = path.join(transpiledTree.outputPath,'nodejs/ember-compat');
+      } else if (source === 'pretender') {
+        result_abs = path.join(transpiledTree.outputPath,'nodejs/no-pretender');
+      } else if (source.startsWith('ember-cli-mirage/')) {
+        result_abs = path.join(transpiledTree.outputPath,source);
+      } else if (source.startsWith('../mirage/')) {
+        result_abs = path.join(transpiledTree.outputPath,'na',source);
+      } else if (source === '../config/environment') {
+        result = appEnvPath;
+      }
+      if (result_abs) {
+        result = path.relative(dirname, result_abs);
+        if (!(result.startsWith('../') || result.startsWith('./'))) {
+          result = './' + result;
+        }
+      }
+      return result;
+    }
+  });
+  var builder = new broccoli.Builder(transpiledTree);
+  return { builder, appEnv };
+}
+
 module.exports = {
   name: 'ember-cli-mirage',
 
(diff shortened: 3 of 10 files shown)


samselikoff:
I see. Could you reach out to the FastBoot team and ask if there's a way to get babel-transpiled code into the Node environment?

mfazekas:
@samselikoff pls review - i think this is complete now

samselikoff:
Ok so - this depends on FastBoot? Can you walk me through it a bit. I wasn't thinking we'd need FB for this

mfazekas:
@samselikoff so issue we need to solve is loading ember code which normally runs in browser to Node.js env. Fastboot already solves this, so in a way it makes sense to build on that. These are the dependencies on fastboot:

1. `ember-cli-mirage` itself depends on `fastboot` which is just code used by `ember-cli-fastboot`, including it will not change the apps behaviour in any way
2. apps wanting to use mirage via express needs to use on `ember-cli-fastboot`

Which dependency are you having issue with? I'd say `1.)` is not an issue, `2.)` is more problematic, but it still means one can use the express backed for testing  fastboot apps, and others not using `ember-cli-fastboot` can decide if the `express` backed worth enough to add that or not.

We could try to do without fastboot, but that requires locating the `vendor.js` and `<appname>.js` then loading them to node.js and reload them on change. I'd assume just using require is probably not going to work we need a lot of what fastboot does which loads the code using eval `https://github.com/ember-fastboot/fastboot/blob/master/src/ember-app.js#L130`

Removing the fastboot dependency would not change the way this PR works it would only change `serverMiddleware` and `postBuild` parts.

samselikoff:
Thanks for the response, I've been pressed for time but hoping to look at this in more detail soon.

Conceptually my thinking was to read in the mirage code from disk and require it in node, and use it to define those route handlers within Ember CLI's existing express server.

The reason to avoid adding fastboot or ember-cli-fastboot is just because they are more dependencies and deps can cause issues. Again I'll need to go through and see how they are used.

What if we started with getting it working in Ember CLI's express but without automatic reloading?

mfazekas:
> Thanks for the response, I've been pressed for time but hoping to look at this in more detail soon.

> Conceptually my thinking was to read in the mirage code from disk and require it in node, and use it to define those route handlers within Ember CLI's existing express server.

I've tried that route but mirage code depends on parts of ember which is not something you can just `require` and run inside node asis as they were meant to run in the browser. As i recall such examples are string operations, promises.

Otherwise we define the routes on Ember CLI's existing express server. Actually we define the routes on an `express.Router` instance object and we add that to the express server started by ember-cli.

> The reason to avoid adding fastboot or ember-cli-fastboot is just because they are more dependencies and deps can cause issues. Again I'll need to go through and see how they are used.
> What if we started with getting it working in Ember CLI's express but without automatic reloading?

I hear you i'll look into getting it working without fastboot by doing the same eval stuff done by fastboot, but that change should affect small part of this PR.

samselikoff:
@mfazekas Ah I see - well Mirage was designed explicitly not to rely on Ember (for precisely this feature and future features related to node-land only stuff) so if there are Ember dependencies we should remove them.

At the most I expect some utilities, no Ember.Object is used anywhere. Perhaps that should be a first step.

addon/server.js
```diff
@@ -167,7 +130,7 @@ export default class Server {
     let hasFactories = this._hasModulesOfType(config, 'factories');
     let hasDefaultScenario = config.scenarios && config.scenarios.hasOwnProperty('default');
 
-    this.pretender = this.pretender || createPretender(this);
+    this.interceptor = this.interceptor || createInterceptor(this);
```
samselikoff: `this.pretender` was basically a public API, it let folks get access to the underlying pretender server. I'm thinking we should add code here to to create a `this.pretender` alias if `this.intercepter instanceof Pretender`.

app/initializers/ember-cli-mirage.js
```diff
@@ -23,7 +25,7 @@ export function startMirage(env = ENV) {
   let modules = readModules(env.modulePrefix);
   let options = _assign(modules, {environment, baseConfig, testConfig});
 
-  return new Server(options);
+  return new Server(Object.assign({createInterceptor: createPretender}, options));
```
samselikoff: Let's use Lodash assign, since Object.assign doesn't work in Phantomjs
```
import _assign from 'lodash/object/assign';

_assign({ createInterceptor: createPretender }, options));
```

index.js
```diff
@@ -112,9 +176,6 @@ module.exports = {
   },
 
   _shouldIncludeFiles: function() {
-    if (process.env.EMBER_CLI_FASTBOOT) {
```
samselikoff: I believe this was here so Mirage doesn't blow up when using fastboot. Does this no longer happen?

tests/integration/database-test.js
```diff
@@ -1,5 +1,5 @@
 import {module, test} from 'qunit';
-import Server from 'ember-cli-mirage/server';
+import Server from 'ember-cli-mirage/pretender-server';
```
samselikoff: Shouldn't these be able to use `ember-cli-mirage/server` and by not passing `useExpress` they'll default to using Pretender?

index.js
```diff
@@ -4,6 +4,47 @@ var path = require('path');
...
+function toNodeJSBuilder(projectRoot) {
+  var esTranspiler = require('broccoli-babel-transpiler');
+  var broccoli = require('broccoli-builder');
+
+  var appEnvPath = path.join(projectRoot, 'config', 'environment.js');
+  var appEnv = require(appEnvPath);
+
+  let emberCliMiragePath = __dirname;
+  let mirageAddon = Funnel(emberCliMiragePath, {srcDir: 'addon', destDir: 'ember-cli-mirage'});
+  let appMirageConfig = Funnel(projectRoot, {srcDir:'mirage', destDir: 'mirage'});
```
samselikoff: Mirage's dir can be configurable, we might need to look it up here... I kind of wish it wasn't though. shrug

samselikoff:
Some comments reading through it. Excellent work on this, I'm so stoked for it to land!!

Was wondering if we should ship it with 0.3 (which I was about to cut), or get it in for the last 0.2.x series.

addon/server.js
mfazekas: Fixed: `createPretender` is now setting pretender on `Server`.

https://github.com/mfazekas/ember-cli-mirage/blob/express/addon/interceptors/create-pretender.js#L48

mfazekas:
I think we should target `0.3` i'll update my branch

mfazekas:
PR updated. Not it fixes issues noted in review. 

i'm a bit confused bout code climate, it shows 2 new in files i've not modified. Also it's not clear to me if codeclimate is using jslint in addition to eshint, so of the messages sounded jshint and adding a `.jshintrc` with `{esnext: true}` did solve, but i don't see anyhting about jshint in their doc.

```
Use '!==' to compare with 'null'. NEW
      return links[relationshipKey] != null;
Found in addon/serializers/json-api-serializer.js

Use '===' to compare with 'null'. NEW

    if (key == null) {
Found in addon/orm/model.js
```

I

briankulp:
Any update on this getting merged now that `0.3` is out? I'm excited to be able to use mirage in our fastboot environment :)

samselikoff:
Going to take a look at this soon. CodeClimate is new and I'm not super familiar with it so it won't be a blocker. I (or someone) should definitely take time to get it configured properly though, or else it's just noise.

app/initializers/ember-cli-mirage.js
```diff
@@ -23,7 +25,7 @@ export function startMirage(env = ENV) {
   let modules = readModules(env.modulePrefix);
   let options = _assign(modules, {environment, baseConfig, testConfig});
 
-  return new Server(options);
+  return new Server(_assign({createInterceptor: createPretender}, options));
```
samselikoff: Don't think we need this or either of the `create*` functions here, the `createPretender` is already a default option within `Server`.

mfazekas:
> The bottom pane is from running npm ls lodash, looks like ember-cli-mirage is using 3.x which is causing issues?

That is very strange are we using lodash 3.x?? What's strange that using yarn it installs 4.x and i don't see issue on ember serve 

```
ember new ember-cli-mirage-e1-demo
cd ember-cli-mirage-e1-demo
ember install mfazekas/ember-cli-mirage#express
# ./node_modules/lodash is 3.x
yarn
# ./node_modules/lodash is 4.x
```

samselikoff:
I get the build error when setting `useExpress: true` in my config, you don't?

samselikoff:
I think the problem is the `lodash` module is supplied by `ember-lodash` which isn't available in node.

mfazekas:
I was not getting the error when i used `yarn` vs `npm` to install modules. `yarn` seems to have chosen 4.0 and not 3.0 of `lodash` not sure why.

samselikoff:
Gotchya. We'll need to get it working with npm as well.

samselikoff:
ping @mfazekas. we're so close!

mfazekas:
@samselikoff sorry i was busy lately with other stuff. 

So about lodash version, i'm a bit stuck it seems that with ember 2.12 projects are getting lodash 3.1* by default, and even if you add a package requiring lodash-4.* it will not be upgraded.

With ember 2.13 a new project seems to get "4.17.4" even with npm.

mfazekas:
I'm also working on rebasing this to current version, and we'll need to figure out some automated way to test. Since we're compling to different env (node.js) it will be something easy to brake esp if new emberjs or other dependency is introduced.

kellyselden:
Not sure if it is known, but I just tried out this branch and got:

```
Error during build for node.js: TypeError: (0 , _utilsInflector.singularize) is not a function
    at GetShorthandRouteHandler.getModelClassFromPath (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/route-handlers/base.js:40:116)
    at GetShorthandRouteHandler.BaseShorthandRouteHandler (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/route-handlers/shorthands/base.js:32:35)
    at new GetShorthandRouteHandler (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/route-handlers/shorthands/get.js:35:90)
    at createHandler (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/route-handler.js:79:15)
    at new RouteHandler (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/route-handler.js:107:20)
    at Server._registerRouteHandler (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/server.js:602:26)
    at Server._this5.(anonymous function) [as get] (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/server.js:578:18)
    at Server.exports.default (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/mirage/config.js:51:8)
    at Server.loadConfig (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/server.js:269:14)
    at Server.config (/Users/kelly.selden/code/jackrabbit/tmp/broccoli_persistent_filterbabel-output_path-hxNMndi8.tmp/ember-cli-mirage/server.js:206:14)
Build for node.js finished!
```

Hope it helps.

samselikoff:
@mfazekas there was just a lodash PR merged, perhaps it addresses the issue you were running into? 

I've been trying to focus mirage time on landing polymorphic associations so I haven't been able to dive into this. Also stanley was hacking on getting mirage running in a Service Worker, not sure if there are any implications for that here

mfazekas:
@samselikoff  sorry i try to get some time to work in on the weekend. 

There is one issue that sometimes `lodash-3.*` is installed by npm. I'm not that concerned about that as it works with never ember versions, i'll also try with latest mirage.
Other thing is that i need proper emulate stuff like singularize, etc. under `node.js` i assume mirage is using ember-inflector, i'd probably need to use `lodash-inflection` on the node.js side.

rtablada:
a potential issue with using lodash inflector is that custom inflection rules would not be properly inflected.

mfazekas:
@samselikoff The real reason this PR is stuck at the moment that ember-fastboot do have nontrivial ember depencencies. One example is inflector to pluralize, etc strings. While we can do some emulation with `lodash-inflection` it will not be 100% compatible, like custom rules mentioned above.

Another alternative is going back to drawing board and piggyback on fastboot, as fastboot let's Ember code run on the npm side.

kratiahuja:
@samselikoff what's the direction you want to take with this PR? Or what needs to be done to have mirage supported for FastBoot? Happy to help with it.

mfazekas:
@kratiahuja this PR would make mirage work with Fastboot. But it's not tied to fastboot. There was a version where i did use fastboot to run mirage code on node.js, but in this version we're using broccoli directly to transpile mirage code to node.js. Ideally mirage would not have ember dependency, but it does have some use of Ember's `String` for example which makes this approach less than ideal.

xcambar:
> [...] less than ideal.

Less than ideal is better than none. As far as I'm concerned, I'm really looking forward to seeing this PR land in Mirage. The benefits of this PR are fantastic! Being able to debug "as if" we were hitting the proper API and doing analyses with the Network Panel are really worth any tradeoff.

Please land this 🙏

izelnakri:
Hi @mfazekas, I tried to run this branch on a fastboot project(up-to-date) and got this error:
TypeError: Cannot read property '_passthroughCheck' of undefined.

Kuzirashi:
Hello, is this work here going to be continued?

jkeen:
Is there a new strategy that's being looked into for mirage + fastboot, @samselikoff and @mfazekas? I picked up this branch and [rebased it with master](https://github.com/jkeen/ember-cli-mirage/tree/express) and am working on getting it going with a project I'm using [ember-cli-fastboot-testing](https://github.com/embermap/ember-cli-fastboot-testing) on but I'm running into some rough broccoli transformation errors.

It looks like mirage now has fastboot as a dependency, should we revive the initial idea of using fastboot to do all that transforming?

ming-codes:
I think Mirage as Service Worker sounds great. Would it be a better approach to modify `ember-cli-mirage` to be a plugin for `ember-service-worker`? It may even be possible to build a separate addon to bridge the gap without modifying mirage itself.

samselikoff:
@ming-codes good question. I don't know much about service workers yet. I think the APIs I'll be adding to get Mirage working in node will also make it work in a SW.

samselikoff:
I apologize for not communicating more around this issue & letting it linger. The reality was that this expands the problem space Mirage must support to a degree we weren't able to sustain with the resources we had on the project. Maybe this was my fault for not recruiting and empowering more folks to work on the project.

In any case, we've picked this up for some related work, and FastBoot & Node support is a priority of ours at the moment. I'm going to close this PR for now since we're taking a slightly different path, relying on`esm` to do the heavy lifting instead of custom Broccoli code we have to maintain.

@mfazekas Thank you so much for kicking off this work and providing a place for folks to chime in, it helped us understand the need much more! Sorry again about letting this languish but hopefully we will end up with something that works for everyone soon.

Anyone interested in the status of the feature can keep an eye on the [FastBoot support](https://github.com/samselikoff/ember-cli-mirage/issues/1411) issue.

PR #1015, opened 2017-02-17
refactor(deps): rm deprecated lodash dep

anulman:
This one threw me for a loop while debugging. `lodash` is now accessed via `ember-lodash`, so we can rm this dep


samselikoff:
Do these tests pass on your local machine?

anulman:
They did!

I saw those SimpleConcat errors locally too when building; can't recall what cleared them but it was an internal e-cli / broccoli issue

acorncom:
Restarting the build ...

Issue #808, opened 2016-07-14
Polymorphism


kagemusha:
We are trying to model our polymorphic object structure in Mirage and wondering how best to do this.  Our object structure looks like:

```
Variation
    content: DS.belongsTo('content',{polymorphic:true,async:true}),

Content (a superclass)
    (all relations in the subclasses)

StaticImage extends Content
    variation: DS.belongsTo('variation'),

CountdownTime extends Content
    variation: DS.belongsTo('variation'),

(and some other subclasses of Content)

Asset
    content: DS.belongsTo('content',{polymorphic:true,async:true}),
```

We looked at the previously opened issues on polymorphism whic seem unresolved, but apparently there are workarounds?

samselikoff:
Sure, hopefully I can help! Could you write down an example of a simple HTTP response that involves these?

kagemusha:
This is greatly simplified but something as below.  And we are using the ActiveModelSerializer.

The assets call returns:

assets: [
               {id: 10, pic: {type: "static_image", id: 1}, ....}
               ...
             ]

Followed by: 

    {static_images: [
        { id: 1, asset_id: 10, ...}
        ....
      ]}

In the store, we end up with static_image models no content models.  Is that enough detail on the HTTP responses?

samselikoff:
I have limited time to help right now, but I think for this I would try to just keep the modeling simple:

```js
// mirage/models/asset.js
export default Model.extend({
});

// mirage/models/static-image.js
export default Model.extend({
  asset: belongsTo()
});
```

Now `static-image` has a foreign key pointing to an asset. Create the data like this:

```js
// mirage/scenarios/default.js
server.create('asset');
server.create('static-image', { asset });
```

Now the static image should serialize with the asset foreign key. To get the asset payload correct, let's futz with the asset serializer:

```js
// mirage/serializers/asset.js
export default ApplicationSerializer.extend({
 
  serialize(asset, request) {
    let json = Serializer.prototype.serialize.apply(this, arguments);

    json.assets.forEach(asset => {
      let relatedContent = this.registry.schema.staticImages.where({ assetId: asset.id })[0];

      asset.pic = {
        type: relatedContent.modelName,
        id: relatedContent.id
      };
    });
  }
  
});
```

Now of course we're only searching `staticImages` here, so as you add more types you'll need to add some more code here to "find" the first model that has the matching `assetId`.

[...]

Issue #677, opened 2016-04-04
Support for Polymorphic relationships


JakeDluhy:
Unless it's just undocumented, I don't think there's support for polymorphic relationships with the ORM. I think the best solution would be an implementation similar to the way it's represented in rails (e.g. have polymorphicId and polymorphicType attributes on the factory, that the serializer is smart about finding).

I'd definitely be willing to take a crack at it, once I get a bit of time, as one of the projects I'm working on has those data requirements.

p.s. LOVE 0.2.x right now! I'm finding it very easy to mock my data

samselikoff:
:+1: so glad to hear!

Yes we need to add this, and also hasOne (almost done), and hasAndBelongsToMany.

These are blocked by finalizing the serializer API, which should be done soon.

samselikoff:
Oh and if you want to take a crack at it, please do! Once you dive in I'm happy to dive in.

dukejeffrie:
Hi, folks, is this still going to be implemented? Is there a workaround for a polymorphic hasMany()? I've been trying to hack my own, without success.

samselikoff:
Quick update, been busy but got some sponsorship to work on polymorphic. So will be starting on this soon.

bauerjon:
@samselikoff is there any update on the progress of this feature? thanks!

samselikoff:
There has been progress, you can follow along here: https://github.com/samselikoff/ember-cli-mirage/projects/1

I expect another month or two before the ORM will be ready. For now, I'd suggest hand-mocking polymorphic routes, or hacking it into the current serializer layer.

hbrysiewicz:
Closing as this has been added to Mirage. You can view more information about this feature [here](http://www.ember-cli-mirage.com/blog/2017/06/05/polymorphic-associations-and-auto-discovery-of-ember-data-models/)

PR #1016, opened 2017-02-17
[WIP] Feat/polymorphism

anulman:
This is a working reference implementation of polymorphic relationships in mirage (#677 #808). It breaks a bunch of tests, so should not be merged immediately.

Turn on polymorphic behaviour by declaring `polymorphic: true` in your `hasMany` / `belongsTo` opts object. `hasMany` rels will benefit from declaring a specific inverse. For example:

```javascript
// mirage/models/cart.js

export default Model.extend({
  lineItems: hasMany({ polymorphic: true, inverse: 'cart' })
});

// mirage/models/line-item.js

export default Model.extend({
  cart: belongsTo({ polymorphic: true })
});
```

Subsequent models may implement cart-like behaviour (e.g. via mixin or inheritance), and all objects will retain access to their parent / children objects.

Prior discussion recommended mucking with serializers; I avoided doing so here as serialization will always be application-specific (e.g. do I prefix as `cart/foo`, `cart-bar`, or omit the prefix for `baz`?).

Behaviour appears stable in my consuming app, so I think it's worth discussing:
- If this is the preferred implementation, or if there are better proposals waiting to be implemented;
- Edge cases not yet accounted for; and
- How / why tests are failing

Includes a commit from https://github.com/samselikoff/ember-cli-mirage/pull/1015.

### DIFF
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -31,7 +31,7 @@ class BelongsTo extends Association {
    * @public
    */
   getForeignKey() {
-    return `${camelize(this.key)}Id`;
+    return camelize(this.key);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


samselikoff:
I was planning on working on polymorphic support after landing 0-3 stable, this definitely should be based off the ORM changes in that branch (one-way-associations). Thanks for putting this work out there, ping me on slack if you want to discuss further. I think we need to fix a few bugs in 0-3 first.

samselikoff:
Polymorphism support has been added to Mirage

PR #1024, opened 2017-02-23
Replace Number.isInteger with lodash isInteger

ilucin:
Number.isInteger is not well supported in older browser environments so we should use lodash isInteger method instead. Specifically, my CI tests in phantomjs were failing because of this.


samselikoff:
Thanks!

Issue #1021, opened 2017-02-22
Association doesn't create new records


yratanov:
Hi,

I have following factories:

```js
//factories/role.js
import { Factory, association } from 'ember-cli-mirage';

export default Factory.extend({
  user: association()
});
```

```js
//factories/user.js
import { Factory, faker, trait } from 'ember-cli-mirage';

export default Factory.extend({})

```

```js

//models/role.js
import { Model, belongsTo } from 'ember-cli-mirage';

export default Model.extend({
  user: belongsTo()
}); 
```
```js

//models/user.js
import { Model, hasMany } from 'ember-cli-mirage';

export default Model.extend({
  role: hasMany()
}); 
```

When I create several `role` records:

### Unexpected
```js
let role = server.create('role');
role.userId; // 1

let role1 = server.create('role');
role1.userId; // 1
```

### Expected:
```js
let role = server.create('role');
role.userId; // 1

let role1 = server.create('role'); //new user should be created here
role1.userId; // 2 
```

samselikoff:
This definitely seems like a bug.

Could you make a Copy (File > Copy Twiddle) of the [dev boilerplate](https://ember-twiddle.com/eedfd390d8394d54d5bfd0ed988a5d0f) and demonstrate this issue?

Also make sure the version is the same as what you're using, you can change it in `twiddle.json`.

yratanov:
@samselikoff 
Here it is: https://ember-twiddle.com/8faac59f1e2b1ab9a9a84d2284964eaf?openFiles=templates.application.hbs%2C

samselikoff:
Yep, looks like a bug.

Would you have time to take a look? I would start by looking here: https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/server.js#L586-L596

I think this module would be good for adding a test to make it easier to develop & fix https://github.com/samselikoff/ember-cli-mirage/blob/5b0f55e2dfbb091edaf89bc17e63e2b700442861/tests/unit/server-test.js#L1053

yratanov:
@samselikoff ok, I'll take a look.

samselikoff:
#1025 closes

PR #1025, opened 2017-02-24
fix factory association feature

yratanov:
`_mapAssociationsFromAttributes` method modifies factory's attrs field.

Example:
*Before* this method callled attrs can be `{ author: { ..association } }`.
*After* this method called attrs will be `{ authorId: 1 }`

So association will be generated only first time.

As a fix I copy the original factory to apply associations.

FIx for #1021


samselikoff:
thank you!

PR #1026, opened 2017-02-24
Fixes generated code for destroy app helper

NLincoln:
I ran into this issue when writing unit tests for an initializer.


HeroicEric:
Thanks for the help @NLincoln 😄  Can you please provide a description of the problem that this solves?

It also looks like you accidentally commited a `yarn.lock` file

NLincoln:
Initializer tests use `startApp` and `destroyApp`, but mirage doesn't create the server global in those cases, causing errors.

NLincoln:
Lockfile has been removed

samselikoff:
Thanks @NLincoln!

Thank you for the review @HeroicEric , it really helps 😄

PR #1033, opened 2017-03-01
Add a note about dynamic attribute support

danevans:
Display which versions to expect support for referencing other attributes from within dynamic attributes.

Otherwise the docs can be misleading because the feature is half-implemented (the attributes are there but always return `undefined` because the getter is only half implemented) prior to 2.4.


samselikoff:
👍 Thank you very much!

PR #1039, opened 2017-03-09
add support to serialize hasMany relationships

sescobb27:
Hi, I just found that when you get the body of a request, even though you can send a has many relationships in it doing this.

```js
// models/order.js
  details: DS.hasMany('order-detail', {
    inverse: 'order'
  })
})

// models/order-detail.js
  order: DS.belongsTo('order', {
    polymorphic: true,
    inverse: 'details'
  })
})

export default DS.JSONAPISerializer.extend({
  attrs: {
    details: {
      serialize: true
    }
  }
});
```
you don't recognize it, this PR add's support to that case.

### DIFF
--- a/addon/route-handlers/base.js
+++ b/addon/route-handlers/base.js
@@ -58,8 +58,11 @@ export default class BaseRouteHandler {
       Object.keys(json.data.relationships).forEach((key) => {
         let relationship = json.data.relationships[key];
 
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you for your contribution - https://github.com/samselikoff/ember-cli-mirage/pull/1086 closes this one. Sorry about the delay!

PR #1040, opened 2017-03-10
Import `require` to avoid issue in babel@6.

rwjblue:
While testing babel@6 upgrades with apps and addons, there are issues when relying on the global require.

In this case, we are just trying to dynamically grab something from the loader so we can use this `import require from 'require';`.

### DIFF
--- a/addon/utils/read-modules.js
+++ b/addon/utils/read-modules.js
@@ -5,6 +5,7 @@
 import Ember from 'ember';
 import _camelCase from 'lodash/camelCase';
 import { pluralize } from 'ember-cli-mirage/utils/inflector';
+import require from 'require';
 
 const { assert } = Ember;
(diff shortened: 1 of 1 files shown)


cibernox:
@samselikoff I don't know how releases work. This merged PR will appear in the 0.2 or in the 0.3 branch? If it's part of the 0.3, can we backport it to 0.2?

samselikoff:
Still 0.2, 0.3 hasn't been merged yet

samselikoff:
So if you want to you can go ahead and release

cibernox:
@samselikoff I did the release in github but I cannot publish to npm.

samselikoff:
@cibernox thanks - just gave you (and @rwjblue) publish bit

PR #1046, opened 2017-03-17
Documentation for findBy

williamhaley:
Seems like [findBy](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/db-collection.js#L97) exists on db collections, but is not documented on [the site](http://www.ember-cli-mirage.com/docs/v0.2.x/database/#find)


samselikoff:
👍  Thank you!

PR #1056, opened 2017-03-26
[FEATURE] Auto generate models from Ember Data models

offirgolan:
This PR adds the ability to auto generate Mirage models based on ones created via Ember Data. This feature is behind a config flag set in `config/environment.js` and is defaulted to `false`. 

```js
ENV['ember-cli-mirage'] = {
  discoverEmberDataModels: true
};
```

Concerns:

- ~`autoGenerateEmberDataModels` can be changed to whatever. I wasnt really sure what the best wording for it should be.~


samselikoff:
Awesome work — I know the community has been looking forward to this & will be very grateful!!

1. Let's go with `discoverEmberDataModels`, since "generate" could be confused with static file generation
2. `modelFor` isn't bad, do you have a concrete use case for this yet?

offirgolan:
@samselikoff 

> Let's go with `discoverEmberDataModels`, since "generate" could be confused with static file generation

Done. 

> modelFor isn't bad, do you have a concrete use case for this yet?

Honestly, I can't think of one lol. I just think that there should be a way to access these auto generated models so you are able disable or even add relationships or properties. Totally up to you if you want to keep it or not.

samselikoff:
@offirgolan if `modelFor` is speculative let's not push it - folks could come to rely on it before we learn how it should be used. Smaller API the better :)

Currently Mirage models are only for models and relationships, we don't have properties yet but maybe one day if we add default props to models it would be good.

Can you keep the code somewhere but remove the feature from the PR?

offirgolan:
@samselikoff makes sense to me. I removed `modelFor` from public API but still kept the code + test cases so that if we do want to pursue this path, we have a solid starting point.

samselikoff:
Will look at this tomorrow

nadnoslen:
Hey @offirgolan & @samselikoff,

Cool feature.  This could be better documented, say in the [Configuration section of the 0.3 documentation](http://www.ember-cli-mirage.com/docs/v0.3.x/configuration/) and the [0.2 - 0.3 upgrade guide](http://www.ember-cli-mirage.com/docs/v0.3.x/upgrading/#02x--03-upgrade-guide).  I found this PR and [the blog post](http://www.ember-cli-mirage.com/blog/#auto-discovery-of-ember-data-models) after digging around my errors from upgrading.

So why should we shore up the docs?  Sadly, my project is only taking advantage of fixtures...I'm from the 0.1 days.  Being able to turn off model discovery allows me to upgrade without having to incur a whole bunch of tech-debt for a feature I'm not yet using.  Obviously, auto-model detection should be the default, but being able to opt-out needs to be described somewhere.  

Figuring this out was not too difficult for me, but I'm certain others will be spinning in their chairs.

If I had some cycles I'd suggest the edits but I'm a bit slammed...sorry.

PR #1062, opened 2017-04-08
Upgrade faker.js

oskarrough:
Hi, the version of Faker.js currently included is around 70mb. From 4.0.0 it's down to 5mb again. So this upgrade faker.js

BUT, the tests wouldn't run on my machine (before or after) so I can't verify that it works. I hope someone can help verify so we can get some faster installs!

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
     "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
     "ember-data": "^2.10.0",
-    "ember-disable-proxy-controllers": "^1.0.1",
     "ember-disable-prototype-extensions": "^1.1.0",
+    "ember-disable-proxy-controllers": "^1.0.1",
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.5.1",
     "ember-resolver": "^2.0.3",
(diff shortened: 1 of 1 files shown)


acorncom:
@oskarrough Looks like the changes to Faker to reduce filesize (https://github.com/Marak/faker.js/pull/432 - specifically the ignoring of `build` files) means that the import of faker's built files in Mirage [here](https://github.com/oskarrough/ember-cli-mirage/blob/9f59d04e66e2c09de6c17c917486cf5404589d82/index.js#L18) will always fail. 

Looks like the built files are still being built on Github (see https://github.com/Marak/faker.js/commit/0a58d668beab0275636013bded29b7fad7f82279 for an example) but they aren't available in the npm package.  Not sure if the Faker folks intended for the `build` folder to disappear or not, but that seems to be what's breaking this PR at the moment ...

oskarrough:
@acorncom thanks!

If Mirage needs the files in the `build` folder I guess it can't be excluded? Or could mirage switch to use the source files?

acorncom:
@oskarrough I don't see why we wouldn't be able to switch to using the raw source files, do you have time to look into making that work?

oskarrough:
@acorncom I'm sorry but I won't be able to help with this one.

acorncom:
@oskarrough No problem and understood, I don't have the time either 😉 I'll close this PR for now and tag the issue as a help wanted. Thx for your work on it!

caseywatts:
I just tried to use `index.js` directly (instead of `build/build/faker.js`), and I'm getting errors that seem related to the line `var Faker = require('./lib');`.

image

I'm not sure we can import commonjs modules through our `nodeAssets` setup. (I really don't know what I'm doing with our asset pipeline things)

--- 

As a different approach to this, I'm asking if we can include that one file again in this PR:
https://github.com/Marak/faker.js/issues/575

I imagine that will help other people other than just us as well :)

PR #1068, opened 2017-04-22
Add Resource helper custom path parameter

rjschie:
This adds in the ability to specify a customized path in the resource helper.

Currently with a `server.get()` shorthand, the second parameter can be a string for the collection name, if it doesn't match the given path. This PR will bring the `server.resource()` helper in line with that customizability, although it places this option inside the second parameter's option hash as `path`.

```js
// given "blog-posts" resource name and a custom path:
server.resource('blog-posts', { path: '/posts' });

// returns shorthands:
server.get('/posts', 'blog-posts');
server.get('/posts/:id', 'blog-posts');
server.post('/posts', 'blog-posts');
server.put('/posts/:id', 'blog-posts');
server.patch('/posts/:id', 'blog-posts');
server.delete('/posts/:id', 'blog-posts');
```

You could also specify an absolute url as the custom path:

```js
server.resource('blog-posts', { path: 'https://localhost/posts' });

// returns shorthands:
server.get('https://localhost/posts', 'blog-posts');
server.get('https://localhost/posts/:id', 'blog-posts');
server.post('https://localhost/posts', 'blog-posts');
server.put('https://localhost/posts/:id', 'blog-posts');
server.patch('https://localhost/posts/:id', 'blog-posts');
server.delete('https://localhost/posts/:id', 'blog-posts');
```

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -424,7 +424,14 @@ export default class Server {
     }
   }
 
-  resource(resourceName, { only, except } = {}) {
+  resource(resourceName, routePath, { only, except } = {}) {
+    if (typeof routePath === 'object') {
+      only = routePath.only;
+      except = routePath.except;
+      routePath = resourceName;
+    }
+
+    routePath = routePath || resourceName;
     only = only || [];
     except = except || [];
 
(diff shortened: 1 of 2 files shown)


samselikoff:
nice! I feel like I went a bit heavy on the positional-args-as-API in a few places in Mirage, could we use the options hash and add a `path` option instead of using second arg?

rjschie:
Ya definitely! I like that, and it would make things so much simpler. I'll try and get that updated later today.

rjschie:
Ok, I've update the PR to use the `path` option. I also thought it might be nice to include the leading `/` if you define a path, making this open to absolute URL's such as `https://localhost/resource-name/`, rather than putting the leading `/` hardcoded into the resource handler object map.

samselikoff:
Nice work!

PR #1083, opened 2017-05-12
cloneDeep data when loading into a collection.

### LINKED ISSUE #1071: Associations persisting between separate tests
Greetings :-)

I'm attempting to upgrade from 0.2.4 to 0.3.1.

We mostly use a set of fixtures to run all of our acceptance tests. I've updated both sides of our associations throughout the fixtures, as per the upgrade notes.

We have teams and people in our app. A team has many people and a person belongs to a team.

When I run a suite of tests where a person is created in a test, subsequent tests seem to fail because the new person id exists for the team.personIds association, but the new person record has been cleared due to loadFixtures between tests.

An example of the error is:

_**Error: Mirage: You're instantiating a team that has a personIds of 1,2,3,4,1,f7dd226e-c361-46f2-8d2a-92a55d8825f8, but some of those records don't exist in the database.**_

Individually, the tests always run ok. On occasion, the tests pass together.

It's as if the team.personIds association is not being reset between tests. Is that possible? Any ideas on where I can look to resolve it? Happy to dig into the source code, with a bit of direction.

Thanks,
Graeme

stuartc:
Protects nested fixtures from getting mutated during tests.

Fixes #1071.

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -1,14 +1,11 @@
 import _assign from 'lodash/assign';
+import _cloneDeep from 'lodash/cloneDeep';
 import _map from 'lodash/map';
 import _isEqual from 'lodash/isEqual';
 import _sortBy from 'lodash/sortBy';
 
 function duplicate(data) {
-  if (Array.isArray(data)) {
-    return data.map(duplicate);
-  } else {
-    return _assign({}, data);
-  }
+  return _cloneDeep(data);
 }
 
 /**
--- a/tests/unit/db-test.js
+++ b/tests/unit/db-test.js
@@ -68,12 +68,14 @@ module('Unit | Db #loadData', function(hooks) {
 
   test('it can load an object of data', function(assert) {
     let data = {
-      contacts: [{ id: '1', name: 'Link' }],
+      contacts: [{ id: '1', name: 'Link', items: [{a: 1}] }],
       addresses: [{ id: '1', name: '123 Hyrule Way' }]
     };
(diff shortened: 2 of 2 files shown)


samselikoff:
restarting master build in case it's something there

samselikoff:
Hm, master seems fine. Could you investigate the failing tests?

stuartc:
@samselikoff yeah weird that! sure I'll see what's causing the failures.

HeroicEric:
@stuartc I'm running into this issue as well while trying to upgrade. Have you had a chance to dig into this any further? I'll likely be spending some time on Friday on this.

samselikoff:
Believe there's something here with ember-inflector breaking changes. Could someone could dig into this & summarize our possible options for me?

stuartc:
@HeroicEric unfortunately I haven't managed to get time to check it out - I'm using my fork for a project without any issues - I won't be able to get around to this until the weekend.

HeroicEric:
@samselikoff I reran the build and this seems to be passing now

stuartc:
I rebased against master a few days ago, and pushed - I see the Chrome Canary travis jobs are set to 'allow failures', so the tests pass. So this is good to go if you're happy @samselikoff

BnitoBzh:
nobody to do the necessary?

stuartc:
@samselikoff just rebased against master, would be great to get this in a release. :)

samselikoff:
This has been fixed since `loadData` does a _cloneDeep now.

Issue #1072, opened 2017-04-28
Shorthand `this.patch('users/:id');` does not persist changes in relationships


Herriau:
It seems like the shorthand for `this.patch()` properly persists attributes, but mutations of relationships, whether they are `belongsTo` or `hasMany` do not appear to be persisted.

Example:
[https://ember-twiddle.com/c496ddbf58986fd2082a2e4771a5987d](https://ember-twiddle.com/c496ddbf58986fd2082a2e4771a5987d)

Click the "Update Users" button, which removes the posts associated with the first user in the collection and then `save()` that user. The post counter next to the first user will temporarily change to "0 posts" while the request is still being handled by mirage, but will revert to 1 as mirage responds with the old list of related posts.

samselikoff:
I believe https://github.com/samselikoff/ember-cli-mirage/pull/1086 closes this

PR #1086, opened 2017-05-15
Normalize hasMany relationships

npafundi:
This commit updates `_getRequestAttrs` to normalize any `hasMany`
relationships in a request's JSON data.

By default, Ember Data will serialize `hasMany` relationships if they
are many-to-none, many-to-many, or if serialization is forced via
`serialize: true` in the `attrs` property on [DS.JSONSerializer](https://emberjs.com/api/data/classes/DS.JSONSerializer.html).
(One-to-many are not serialized by default).

Before this commit, Mirage only normalized attributes and `belongsTo`
relationships. Because `hasMany` relationships weren't normalized, many-
to-many relationships weren't saved via the shorthand route handlers or
any custom route handlers that used `this.normalizedRequestAttrs()`
(which calls `_getRequestAttrs`).

For example, assume a many-to-many relationship between `Authors` and
`Books`. In Ember, if additional `Books` are associated with an `Author`
and saved, i.e.:
```javascript
author.get('books').pushObjects([anotherBook, yetAnotherBook]);
author.save();
```
These changes wouldn't be saved and reflected in Mirage, because the
normalization didn't include the new Ids.

Please let me know if this is an appropriate solution, or if you'd like to see any changes. Thanks!

### DIFF
--- a/addon/route-handlers/base.js
+++ b/addon/route-handlers/base.js
@@ -58,7 +58,9 @@ export default class BaseRouteHandler {
       Object.keys(json.data.relationships).forEach((key) => {
         let relationship = json.data.relationships[key];
 
-        if (!Array.isArray(relationship.data)) {
+        if (Array.isArray(relationship.data)) {
+          attrs[`${camelize(singularize(key))}Ids`] = relationship.data.map(rel => rel.id);
+        } else {
           attrs[`${camelize(key)}Id`] = relationship.data && relationship.data.id;
         }
       }, {});
--- a/tests/unit/route-handlers/shorthands/base-test.js
+++ b/tests/unit/route-handlers/shorthands/base-test.js
@@ -62,6 +62,21 @@ test('_getAttrsForRequest works with attributes and relationships', function(ass
             'type': 'companies'
           }
         },
+        'employees': {
+          'data': [{
+            'id': '1',
+            'type': 'employees'
+          }, {
+            'id': '2',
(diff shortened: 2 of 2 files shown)


npafundi:
Codeclimate is failing with "Too many errors. (100% scanned).", but no errors listed. Let me know if I can address this with any changes.

nathanhammond:
This solves for the same goal as #1039 and should supersede it.

npafundi:
Thanks @nathanhammond.

This may also resolve #1072; at least partially. This should maintain many-to-many relationships, but one-to-many still won't persist because Ember Data doesn't seem to serialize those by default. My understanding is that you must save records on the "many" side for that to work properly; see [`_shouldSerializeHasMany`](https://github.com/emberjs/data/blob/v2.13.0/addon/serializers/json.js#L871) in the Ember Data json serializer.

mupkoo:
This is just what I need. I was looking to see if there is any PR that is does it and this looks like it.

Currently I have this in handler

```javascript
this.patch('/people/:id', function ({ people }, request) {
  let id = request.params.id;
  attrs.tagIds = JSON.parse(request.requestBody).data.relationships.tags.data.map(({ id }) => id);

  return people.find(id).update(attrs);
});
```

PS The other PR #1039 is using different naming `tagsId`

samselikoff:
Awesome - thanks everyone!

PR #1087, opened 2017-05-16
RestSerializer should always serialize ids

jrjohnson:
Override the default Serializer behavior in the RestSerializer to
convert ids in the expected way.

I'm fairly certain that this would be a breaking change and it has no tests, I just wanted to float this out there for some comments. The current behavior was a big surprise to me making the transition from 0.1 the RestSerializer was sending back `postId: 1` and `authorIds: [1, 2]` instead of `post: 1`, `authors: [1, 2]`. The `serializeIds` option is easy to use and well documented once I discovered the issue, I just feel like it should be the default for RestSerializer.


samselikoff:
Yes this has come up and you're correct, it would technically be a breaking change (even if it is more in line with most RestSerializer implementations).

I think we will do this for RestSerializer and ActiveModelSerializer for v1.0.

jrjohnson:
I don't know anything about ActiveModel, but I believe you. I updated that as well in the PR. I also added a default test for it, but it is *really* sparse. What I'm confused about is whey the integration tests for these two serializers didn't break. If anyone knows what is going on there I can update those tests and add tests for the other states for serializeIds.

samselikoff:
Thank you, and sorry about the delay!

jrjohnson:
No problem. Thanks for getting it over the finish line!

PR #1093, opened 2017-05-27
Update route-handlers.md

MartinMalinda:
Import Response correctly and add demo of JSON-API compliant error response


samselikoff:
Thanks! :)

PR #1096, opened 2017-06-01
Fixed existence of `relationships.data` if `links` are defined

### LINKED ISSUE #973: Specifying `links` in JSONAPISerializer removes relationship data
If a `links` method is specified in a subclass of JSONAPISerializer, the relationship object's `data` property will not be emitted. I believe we should be able to augment the relationship hash output with `links`, not have it be one or the other.

lancedikson:
JSON API spec says that a `relationships` object must contain at least one of the props, but it could contain all of them. So, `links` are able to be with `data` as well.
Link to the article of the spec: http://jsonapi.org/format/#document-resource-object-relationships
An article about relationships when they are defined, but empty: http://jsonapi.org/format/#document-resource-object-linkage (like a case in tests)

fixes #973

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -137,33 +137,32 @@ export default Serializer.extend({
     model.associationKeys.forEach((key) => {
       let relationship = model[key];
       let relationshipKey = this.keyForRelationship(key);
-      let relationshipHash;
+      let relationshipHash = {};
       hash.relationships = hash.relationships || {};
 
       if (this.hasLinksForRelationship(model, key)) {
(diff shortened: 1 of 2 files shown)


lancedikson:
Failed test seems an accident. I've taken care about tests :)

lancedikson:
@samselikoff, sorry for bothering, just a reminder ;)

lancedikson:
@samselikoff, I politely urge to merge it before this piece of code gets obsolete 😃

samselikoff:
Restarting the failed job, will merge when passes :)

lancedikson:
Great! 👍

PR #1098, opened 2017-06-02
Fix FastBoot 1.0

simonihmig:
The upcoming FastBoot 1.0 (incl. latest RC versions) include breaking changes, that don't work anymore with ember-cli-mirage. See https://github.com/ember-fastboot/ember-cli-fastboot/issues/387

This adds FastBoot tests (just testing that it does not break FastBoot rendering. They were failing at first with the latest FastBoot releases), and the required changes.

*Note: I set `engine` to `>= 4` to be able to use some ES6 in node. Hope that is ok!*


simonihmig:
The failing scenario in Travis seems to be not related, a rerun should hopefully fix this! Will have to look at the codeclimate complains though...

stefanpenner:
@simonihmig rerunning that one failed scenario.

stefanpenner:
@simonihmig looks like the CI gods have accepted our offering.

You may want to look into the codeclimate stuff though.

samselikoff:
Awesome -- thank you very much! (This would have taken me ages..)

I got rid of codeclimate so no worries there (it was too noisy)

🙇

simonihmig:
@samselikoff can we get a release with this? please 😛

samselikoff:
@simonihmig all set https://github.com/samselikoff/ember-cli-mirage/releases/tag/v0.3.4

simonihmig:
Awesome, thanks!

billdami:
Are there specific blockers/complications that require mirage to be disabled in fastboot? This seems to impede development quite a bit when running `ember serve` (which now runs fastboot), as livereloading while on a route that expects a mirage-backed model will 404 on the server-side, forcing you to either navigate back to that page once it renders on the client-side, or use `?fastboot=false` to disable fastboot rendering altogether.

simonihmig:
@billdami sorry, missed your question. Basically, Mirage works be using pretender.js to intercept ajax requests and fake the response. In FastBoot however this does not work, as there is no Ajax in node.js, HTTP requests are made with node APIs. I am very much looking forward to a world where Mirage would work in FastBoot out of the box, as probably many others. But it's not there yet. And this is unrelated to this PR. I believe #1014 would solve that issue, but no idea how this is going forward...

PR #1102, opened 2017-06-08
Extend modelRegexp to match models in pod structure

dguettler:


barryofguilder:
@dguettler, thanks for doing thing!

PR #1105, opened 2017-06-09
Documentation: change sequence starting int to zero

happycollision:
I am pretty sure this is actually how it works. I got an id of zero for my first instance of a model.


samselikoff:
Yes - thank you v much :)

Issue #1112, opened 2017-06-16
Unable to delete records with polymorphic belongsTo associations


omghax:
Okay, this is a little involved, so bear with me.

I was getting some really strange errors in my app when trying to delete a record with a polymorphic `belongsTo`. In my case, the record itself would be deleted, but it wouldn't be removed from the inverse relationship correctly.

After digging around through Mirage's source, I think I found the root cause. In the `BelongsTo#disassociateAllDependentsFromTarget` method, a [query is generated](https://github.com/samselikoff/ember-cli-mirage/blob/fe8ec9c838feec42270e950beb55ef0619ecd6c6/addon/orm/associations/belongs-to.js#L228-L235) which, in the case of a polymorphic association, ends up looking like:

```js
aCollection.where({ aForeignKey: { id: 1, type: 'some-model-type' } })
```

Ultimately this falls into `DbCollection#_findRecordsWhere` which, in the case that it's given an object, will attempt to [coerce all values to strings](https://github.com/samselikoff/ember-cli-mirage/blob/fe8ec9c838feec42270e950beb55ef0619ecd6c6/addon/db-collection.js#L295-L297). But, if we have a polymorphic association, these values will all be `{ id, type }` pairs, so coercing them into strings will result in `"[object Object]"`, and _every_ record will be matched.

I think the simplest way to fix this would be to pass a function to `where` instead (in fact, `HasMany#disassociateAllDependentsFromTarget` [already does this](https://github.com/samselikoff/ember-cli-mirage/blob/fe8ec9c838feec42270e950beb55ef0619ecd6c6/addon/orm/associations/has-many.js#L256-L268)).

samselikoff:
Oof - great sleuthing!

That sounds like a good fix to me. Would you be able to open a PR with the failing test, and then make the change? Even a PR with the failing test would be very helpful.

Thank you for taking all the time to open this issue!

omghax:
Thanks for the quick reply :)

Yeah, I already have the change ready, I just need to come up with a minimal failing test case. I'll try to get a PR out shortly.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1173 closes

PR #1113, opened 2017-06-16
Unable to delete records with polymorphic belongsTo associations

omghax:
This PR addresses #1112. @samselikoff I wasn't sure where the test should live, so I just threw it into `tests/integration/belongs-to` for now. It might be possible to reduce the test further, this was just the simplest analogue to my app, and caused the same sort of failure I was seeing.

### DIFF
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -232,7 +232,19 @@ export default class BelongsTo extends Association {
     }
 
     let dependents = this.schema[toCollectionName(owner)]
-      .where({ [this.getForeignKey()]: fk });
+      .where((potentialOwner) => {
+        let id = potentialOwner[this.getForeignKey()];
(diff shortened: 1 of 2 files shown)


anulman:
@omghax just submitted another polymorphism-related PR; may be helpful: https://github.com/samselikoff/ember-cli-mirage/pull/1153

samselikoff:
@omghax great work! I wanted to understand the bug so it took me a while but your solution is perfect. The user was being correctly disassociated, but then it was getting re-associated because that `dependents` search was returning the comment as a dependent of itself, leading to weirdness.

I guess in the future methods like this could use some more lower-level tests (given this setup, `getDependents()` returns the correct models). For now this is great.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1173 closes

Issue #1108, opened 2017-06-14
links regression in #1096


kellyselden:
Before commit:
```
Mirage: [200] GET /profiles?filter%5BfirstName%5D=asdf&filter%5BmiddleName%5D=asdf&filter%5BlastName%5D=asdf&filter%5BcityState%5D=asdf
Mirage: [200] GET /profiles/1/offers
```
After commit
```
Mirage: [200] GET /profiles?filter%5BfirstName%5D=asdf&filter%5BmiddleName%5D=asdf&filter%5BlastName%5D=asdf&filter%5BcityState%5D=asdf
Mirage: [200] GET /offers/1
Mirage: [200] GET /offers/2
...
```

The payload in the first request has the correct links property in both cases.

kellyselden:
cc @lancedikson

lancedikson:
Hm. That's weird. @kellyselden do you have `links` hooks for Mirage's serializers? Could you please share the response object in both cases? I guess it would help to find a problem.

kellyselden:
This regression has now released in v0.3.4. I recommend reverting, then merging in #1120, and then reattempting the fix with that regression test in place.

samselikoff:
@kellyselden Thanks for letting me know. I've been busy, is there someone who can fix?

@lancedikson did you see Kelly's PR with the failing test?

lancedikson:
Yes, I did, but I haven't tried to find the problem yet. I'm gonna have a look on it today.

lancedikson:
I made a research and it seems that ED doesn't work with included explicit relationships correctly (according to the spec). I started [a discussion topic on the JSONAPI forum](http://discuss.jsonapi.org/t/could-links-and-data-persist-in-compound-documents-relations/1023) to make sure that to have both `links` and `data` for included relations is as good as it's good for the primary data in the response (because for the primary data it's ok for sure).

Now I'm going to implement a workaround for cases when a resource is being got as included (`getHashForIncludedResource`). Then I continue the investigation and propose the fix to ED if it's needed.

samselikoff:
Sorry for the late reply...

I just checked out #1120  locally and the test didn't break. @kellyselden @lancedikson Do we know if this is still an issue?

samselikoff:
I believe this somehow sorted itself out, so closing for now. Happy to reopen if still an issue.

PR #1120, opened 2017-06-23
test for links

kellyselden:
A test for links mentioned introduced in #1096 and mentioned in #1108. I branched off the commit before the breaking commit, so CI should pass, but if you merge, it will fail.

cc @lancedikson


tests/dummy/mirage/serializers/blog-post.js
```diff
@@ -1,5 +1,12 @@
 import { JSONAPISerializer } from 'ember-cli-mirage';
 
 export default JSONAPISerializer.extend({
+  links(blogPost) {
+    return {
+      comments: {
+        related: `/blog-posts/${blogPost.id}/comments`
```
lancedikson: @kellyselden, could you please update the path here as well to `/api/blog-posts/${blogPost.id}/comments`?

lancedikson: Thanks :)

samselikoff:
Sorry for the late reply...

I just checked this out locally and the test didn't break. @kellyselden @lancedikson Do we know if this is still an issue?

samselikoff:
I believe this somehow sorted itself out, so closing for now. Happy to reopen if still an issue.

PR #1121, opened 2017-06-24
Normalize relationship ids when normalizeIds is set to true

### LINKED ISSUE #972: RESTSerializer throws on pushing record with relationship
Following docs ember-cli-mirages `RestSerializer` should be used "to fake backends that match Ember Data’s RestSerializer expected response format". But if using it together with default ember-data RESTAdapter and RESTSerializer it throws on pushing records with relationship:

> Mirage: Your handler for the url /comments threw an error: newModel.isNew is not a function

I've set up an ember twiddle demonstrating the issue: https://ember-twiddle.com/328e1448a6a50e8372f0709b4a4be3f6

I tried to debug the issue but I'm not quite sure whats the expected behavior. I figured out that ember-cli-mirage throws cause it expects an model instance [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/associations/belongs-to.js#L106-L118) but it's called with an id.

Also ember-cli-mirage RestSerializer does not seem to follow ember-data RESTSerializer as stated in docs. ember-cli-mirage RestSerializer seem to expect an "Ids" suffix[1] which ember-data RESTSerializer does not use by default[2]. But overriding `keysForRelationshipIds` method does not fix the issue. To test that one just replace `mirage/serializers/applications.js` in twiddle with this one:
```
import { RestSerializer } from 'ember-cli-mirage';
import { camelize, singularize } from 'ember-cli-mirage/utils/inflector';

export default RestSerializer.extend({
  keyForRelationshipIds(type) {
    return camelize(singularize(type));
  }
});
```

Digging deeper into it, `normalize` method of `ActiveModelSerializer` which is also used by `RestSerializer` doesn't seem to return a valid JSONApiDocument. In my example it's called with this payload:

```
{
  "comment": {
    "post": "2"
  }
}
```

And returns this object:

```
{
  "data": { 
    "type": "comments",
    "attributes": {
       "post": "2"
    }
  }
}
```

As far as I got JSON Api specs relationships should not be part of `attributes` object. They should be included in `relationships` object as `data` property.

Also it seem to be impossible to override the key used for a belongs-to relationship. It's suffixed by "Id" and there doesn't seem to be any `keyForRelationshipId` method.

[1] [addon/serializers/rest-serializer.js](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializers/rest-serializer.js#L22-L24) and [tests/integration/serializers/rest-model-serializer-test.js](https://github.com/samselikoff/ember-cli-mirage/blob/24f994f4419f1b175c82331b00c3089e3f65c038/tests/integration/serializers/rest-model-serializer-test.js#L50)
[2] [DS.RESTSerliazer normalize docs](http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_normalize)

jrjohnson:
Fixes #972 mainly with code I stole from @jelhan 

Changes the `normalize` method in `ActiveModelSerializer` to respect a `normalizeIds` property and convert REST type id arrays into JSONApi objects.

It isn't clear to me that we need a boolean to control this behavior as I don't know of any REST or ActiveModel APIs that *don't* expect data in this format. Hiding this behavior behind a flag does make this a non-breaking change though which is easier to manager.

Edit 6/30: Issues below are solved, leaving them in for posterity.

Some specific issues:
1) Should this transform be applied to ActiveModel type APIs. I think they are similar to REST in this way, but need to confirm with someone who knows.
2) ~Is piggybacking on `serializeIds` the right path here or should I add a new `normalizeIds` property to the serializer?~
3) This line is problematic: 
`let { belongsToAssociations, hasManyAssociations } = this.schema._registry[type].class.prototype;`
It depends on the private `_registry` and also breaks several tests because they will need mock schema objects injected into the serializer.

@samselikoff et al is there a better way to determine what the types of associations are for a given model?

### DIFF
--- a/addon/serializers/active-model-serializer.js
+++ b/addon/serializers/active-model-serializer.js
@@ -1,8 +1,8 @@
 import Serializer from '../serializer';
-import { underscore, pluralize, dasherize, singularize } from '../utils/inflector';
+import { underscore, pluralize, dasherize, singularize, camelize } from '../utils/inflector';
 
 export default Serializer.extend({
-
+  normalizeIds: false,
   keyForModel(type) {
     return underscore(type);
   },
@@ -38,6 +38,11 @@ export default Serializer.extend({
   normalize(payload) {
     let type = Object.keys(payload)[0];
     let attrs = payload[type];
+    let modelName = camelize(type);
+    let modelClass = this.schema.modelClassFor(modelName);
+    let { belongsToAssociations, hasManyAssociations } = modelClass;
+    let belongsToKeys = Object.keys(belongsToAssociations);
+    let hasManyKeys = Object.keys(hasManyAssociations);
 
     let jsonApiPayload = {
       data: {
@@ -48,11 +53,42 @@ export default Serializer.extend({
     if (attrs.id) {
       jsonApiPayload.data.id = attrs.id;
     }
+
+    let relationships = {};
+
     Object.keys(attrs).forEach((key) => {
       if (key !== 'id') {
-        jsonApiPayload.data.attributes[dasherize(key)] = attrs[key];
+        if (this.normalizeIds) {
+          if (belongsToKeys.includes(key)) {
+            let association = belongsToAssociations[key];
+            let associationModel = association.modelName;
+            relationships[dasherize(key)] = {
+              data: {
+                type: associationModel,
+                id: attrs[key]
+              }
+            };
+          } else if (hasManyKeys.includes(key)) {
+            let association = hasManyAssociations[key];
+            let associationModel = association.modelName;
+            let data = attrs[key].map(id => {
+              return {
+                type: associationModel,
+                id
+              };
+            });
+            relationships[dasherize(key)] = { data };
+          } else {
+            jsonApiPayload.data.attributes[dasherize(key)] = attrs[key];
+          }
+        } else {
+          jsonApiPayload.data.attributes[dasherize(key)] = attrs[key];
+        }
       }
     });
+    if (Object.keys(relationships).length) {
+      jsonApiPayload.data.relationships = relationships;
+    }
 
     return jsonApiPayload;
   },
--- a/tests/unit/serializers/rest-serializer-test.js
+++ b/tests/unit/serializers/rest-serializer-test.js
@@ -1,10 +1,18 @@
 import RestSerializer from 'ember-cli-mirage/serializers/rest-serializer';
 
 import { module, test } from 'qunit';
+import { Model } from 'ember-cli-mirage';
+import Schema from 'ember-cli-mirage/orm/schema';
+import Db from 'ember-cli-mirage/db';
 
 module('Unit | Serializers | RestSerializer', function(hooks) {
   hooks.beforeEach(function() {
-    this.serializer = new RestSerializer();
+    let schema = new Schema(new Db(), {
+      person: Model
+    });
+    this.serializer = new RestSerializer({
+      schema
+    });
   });
 
   test('it hyphenates camelized words', function(assert) {
--- a/tests/unit/serializers/active-model-serializer-test.js
+++ b/tests/unit/serializers/active-model-serializer-test.js
@@ -1,10 +1,23 @@
+import { Model, belongsTo } from 'ember-cli-mirage';
+import Schema from 'ember-cli-mirage/orm/schema';
+import Db from 'ember-cli-mirage/db';
 import ActiveModelSerializer from 'ember-cli-mirage/serializers/active-model-serializer';
 
 import {module, test} from 'qunit';
 
 module('Unit | Serializers | ActiveModelSerializer', function(hooks) {
   hooks.beforeEach(function() {
-    this.serializer = new ActiveModelSerializer();
+    let schema = new Schema(new Db(), {
+      contact: Model.extend({
+        address: belongsTo()
+      }),
+      address: Model.extend({
+        contact: belongsTo()
+      })
+    });
+    this.serializer = new ActiveModelSerializer({
+      schema
+    });
   });
 
   test('normalize works', function(assert) {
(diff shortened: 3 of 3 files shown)


jelhan:
Thanks for pushing this forward.

> Is piggybacking on serializeIds the right path here or should I add a new normalizeIds property to the serializer?
I'm quite sure if we really need this configuration at all. There are just a very few use cases in which a server ignores relationships presented in `POST` or `PATCH` request. It might be okay to not support these uses cases out of the box.

addon/serializers/active-model-serializer.js
```diff
@@ -38,19 +38,49 @@ export default Serializer.extend({
...
     if (attrs.id) {
       jsonApiPayload.data.id = attrs.id;
     }
+
     Object.keys(attrs).forEach((key) => {
       if (key !== 'id') {
-        jsonApiPayload.data.attributes[dasherize(key)] = attrs[key];
+        if (this.serializeIds === 'always') {
+          if (belongsToKeys.includes(key)) {
+            let association = belongsToAssociations.hasOwnProperty(key);
```
jelhan: `Object.prototype.hasOwnProperty()` returns `Boolean`.

jelhan: I think it should be `let association = belongsToAssociations[key];`.

jrjohnson: Thanks, bad copy paste I guess. I wonder why it worked like that at all, weird.

jrjohnson:
I tend to agree that (in REST type at least) the serializeIds should always be set to 'always'. However, since the option does exist there isn't a lot of reason not to make this configurable as well - given how simple the separation is to accomplish.

jelhan:
Please mind the difference between serialisation of a record for response and normalization of request payload. `serializeIds` option controls if relationship ids should be included in *response*. It's another question if relationships presented in payload should be ignored.

addon/serializers/active-model-serializer.js
jelhan: Your tests only cover if it throws not if records are actually persisted in server database.

jrjohnson:
@jelhan yes, that was my issue with using this configuration option again. I think I will add a new option to control the normalization behavior. Thanks.

jrjohnson:
@jherdman I've fixed tests and found a way to use public APIs to get to the relationship data. What do you think?

jrjohnson:
@jherdman / @samselikoff I believe this is ready. We've been using this code in our application serializer for some time successfully. Are there further changes I should make? Should I abandon this PR and just continue to rely on application level serializers for REST type mocking?

jrjohnson:
Just checking in here. I'm happy to rebase this again, but don't want to keep spending the time if it wont be merged. I'm not really sure how anyone could be using the RestSerializer without this code though.

samselikoff:
I apologize Jonathan - I should have said either yes or no, rather than leave you hanging.

Taking a look now!

addon/serializers/active-model-serializer.js
```diff
@@ -38,6 +38,11 @@ export default Serializer.extend({
   normalize(payload) {
     let type = Object.keys(payload)[0];
     let attrs = payload[type];
+    let modelName = camelize(type);
+    let model = this.schema.modelFor(modelName);
+    let { belongsToAssociations, hasManyAssociations } = model.class.prototype;
```
samselikoff: You can get these from [schema.modelClassFor(modelName)](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/schema.js#L214). (There's also `schema.associationsFor(modelName)` but since you need the hasMany and belongsTo split I think modelClassFor is the best approach).

addon/serializers/active-model-serializer.js
```diff
@@ -1,8 +1,8 @@
 import Serializer from '../serializer';
-import { underscore, pluralize, dasherize, singularize } from '../utils/inflector';
+import { underscore, pluralize, dasherize, singularize, camelize } from '../utils/inflector';
 
 export default Serializer.extend({
-
+  normalizeIds: false,
```
samselikoff: What is the behavior today if a relationships is passed to this serializer? If it's completely broken I think we could avoid adding this flag.

tests/unit/serializers/rest-serializer-test.js
```diff
@@ -1,10 +1,17 @@
...
 
 module('Unit | Serializers | RestSerializer', function(hooks) {
   hooks.beforeEach(function() {
-    this.serializer = new RestSerializer();
+    let schema = new Schema(new Db(), {
+      person: Model
+    });
+    this.serializer = new RestSerializer({
+      schema
+    });
```
samselikoff: Were these needed/used for anything?

samselikoff:
@jrjohnson I am also happy to make the changes I mentioned in my review and add as a commit on top of yours, if you'd like (understandably) move on from this pr! :)

addon/serializers/active-model-serializer.js
jrjohnson: My opinion is that it is broken and doesn't match what EmberData's ResetSerializer expects and #972 documents that. I don't mind avoiding the flag, I just didn't want to create a BC break unnecessarily.

jrjohnson:
Thanks @samselikoff I will make these changes (you have enough to do) if I get stuck I will ask for some help.

tests/unit/serializers/rest-serializer-test.js
jrjohnson: Yes. Without them this test blows up because there is no `schema` attached and then for various other reasons when any of these parts is removed.

addon/serializers/active-model-serializer.js
jrjohnson: Dug back in to this and realized that the big question for me is how ActiveModel should behave. I have no idea so this needs to be settable somehow. I think it should be merged with the flag and then the flag should be defaulted to true at some future point.

samselikoff:
🎉  Thank you @jrjohnson !

PR #1122, opened 2017-06-30
Remove broccoli-unwatched-tree dependency.

rwjblue:
This package is unused in ember-cli-mirage (and it has been deprecated upstream).


samselikoff:
thanks ya big bear

PR #1126, opened 2017-07-10
Add "yarn.lock" file

Turbo87:
This PR adds a `yarn.lock` file and configures CI to use it.

/cc @samselikoff


.travis.yml
```diff
@@ -36,14 +38,13 @@ before_install:
   - wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
   - sudo dpkg -i google-chrome*.deb
   - if [[ $(npm -v | cut -d '.' -f 1) -lt 3 ]]; then npm i -g npm@^3; fi
-  - npm config set spin false
-  - npm install -g bower phantomjs-prebuilt
+  - yarn global add bower phantomjs-prebuilt
 
 install:
-  - npm install
+  - yarn install --no-lockfile --non-interactive
```
bartocc: Wondering why we have the --no-lockfile option here. The purpose of the lockfile is to make sure every `yarn install` command on different machines (CI included) will result with the exact same node_modules directory.

Since we have the yarn.lock in the repo, why not use it?

Thanks
Julien

Turbo87: @bartocc because the `yarn.lock` file is not used when mirage users install it. they will get the latest dependency versions that the constraints in `package.json` allow and that is also what we should test against so that we know once a downstream dependency has broken compat.

bartocc: thx @Turbo87

PR #1136, opened 2017-07-12
Make JsonApiSerializer._getResourceObjectForModel getting proxied isPrimary property

lancedikson:
For the sake of temporary avoiding of the issue when ED doesn't recognize both `links` and `data` inside of `included` property (https://github.com/emberjs/data/issues/5062)

Related to #1120 and mentioned in #1108. It can be merged before the @kellyselden's PR for the purpose not to break tests.


samselikoff:
Closing because I believe #1120 worked itself out.

Issue #509, opened 2016-01-18
Memory leaks because of Pretender request tracking


simonihmig:
I am having some issues with the fact that pretender is keeping a record of all requests it intercepted, which is never cleared (https://github.com/pretenderjs/pretender/blob/master/pretender.js#L299). So all data coming through pretender is effectively locked from garbage collection.

From the view of pretender it is certainly "a feature, not a bug" (TM), as this is intended this way. And this is probably acceptable for short lived application runs for testing. But an important use case for mirage - at least from my point of view - is to be able to develop and play with your app without having to run a real API backend. So not freeing up those resources can pile up memory consumption considerably if the payload is big enough.

In my case it is about many passthrough requests loading a number of quite large assets (>1MB WebGL textures and 3D data). Obviously I don't need mirage/pretender for those, but as I need to mock my JSON API requests, I configured those asset requests as passthrough requests. As said those request payloads will never be garbage collected. No problem for production, not a big problem for desktops, but mobile clients will crash regularly because they run out of memory.

Probably a rather special case here, but still annoying, at least for me! :) Any ideas how to solve this?

bekzod:
is this a solution #510 ?

simonihmig:
Works for me. Just wondering if this may cause problems in the future, as it modifies private data of  the pretender object. Might be better to have a pretender option to opt out of request tracking?

bekzod:
there is no such api in pretender to `opt out tracking` at the time of writing, will try to submit PR with such functionality

simonihmig:
Yeah, I know. That was actually my question, if it would be better to get this functionally implemented in pretender itself instead of patch ing it...

bekzod:
it depends how fast `pretender` maintainers will respond and react to change, for now we can monkey patch it I think and remove it once `pretender` has it covered

bekzod:
added PR not sure if I did it right way as there was no way passing options variable to pretender before

samselikoff:
It looks like those methods just store the fake request object, and not the response. If I'm not mistaken the actual responses from your backend should not be stored anywhere within Pretender or Mirage.

Could you check and see if you have the same memory issues when running your app without Mirage (and thus Pretender) enabled? There have been some known memory leaks within Mirage, but they've typically cropped up when running tests, rather than during development.

simonihmig:
[...]

PR #1137, opened 2017-07-18
bump pretender and disable request tracking in pretender

bekzod:


addon/server.js
```diff
@@ -62,7 +62,7 @@ function createPretender(server) {
          mirage/config.js file. Did you forget to add your namespace?`
       );
     };
-  });
+  }, { trackRequests: false });
```
kellyselden: What is the affect of this on users? Also, is this testable?

package.json
```diff
@@ -80,4 +80,4 @@
   "ember-addon": {
     "configPath": "tests/dummy/config"
   }
-}
+}
```
kellyselden: Can you revert this please, as npm will add this back the next time it gets modified.

addon/server.js
bekzod: no effect to use I think, originally solves this https://github.com/samselikoff/ember-cli-mirage/issues/509

samselikoff:
@Turbo87 @kellyselden thank yall so much for your recent help with these PRs, can't tell you how nice it is to find this stuff getting merged without any intervention on my part! ❤️

PR #1138, opened 2017-07-19
Unlock and upgrade ember-get-config

dfreeman:
`ember-get-config` issued a noisy warning when used with engines in `0.2.1`, and it's locked to that version here.


mnoble01:
@samselikoff Any chance we could get a release with this change?

PR #1153, opened 2017-07-30
fix(base-route-handler): add polymorphism support to _getAttrsForRequest

anulman:
_could not copy Twiddle; was noted as private / secret_

Assume a model `Image` with polymorphic relation `imageable`, and `User` + `Post` which may several relations to `image`:

```javascript
// mirage/models/image.js

export default Model.extend({
  imageable: belongsTo({ polymorphic: true }),
});
```

```javascript
// mirage/models/user.js

export default Model.extend({
  // ...
  image: belongsTo('image'),
});
```
```javascript
// mirage/models/post.js

export default Model.extend({
  // ...
  images: hasMany('image'),
});
```

The base route handler `mirage.post('/images')` at master was throwing 500s, as it cannot find the inverse of the polymorphic rel.

Sleuthing showed that [_getAttrsForRequest's JSON-API relationships loop](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/route-handlers/base.js#L57-L66) sets only `rel.id` vs. `rel.type`. Changing this only affected tests re: exact serialization expected from `_getAttrsForRequest`, so updated those.

In a real app, I observed failures at the ORM level, as mirage could not parse the new rel objects passed for non-polymorphic relationships. This did not cause test failure. Updated the orm objects accordingly; erred on the side of conservatism in case `foreignKeys.id` was unavailable.

This PR resolves the issue (#201created 👍), but seems to propagate further issues with a real-life ORM in spite of the passing tests.

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -457,10 +457,10 @@ class Model {
         });
         found = _compact(found);
       } else {
-        found = this._schema.db[toCollectionName(association.modelName)].find(foreignKeys);
+        found = this._schema.db[toCollectionName(association.modelName)].find(foreignKeys.id || foreignKeys);
       }
 
-      let foreignKeyLabel = association.isPolymorphic ? foreignKeys.map(fk => `${fk.type}:${fk.id}`).join(',') : foreignKeys;
+      let foreignKeyLabel = association.isPolymorphic ? foreignKeys.map(fk => `${fk.type}:${fk.id}`).join(',') : (foreignKeys.id || foreignKeys);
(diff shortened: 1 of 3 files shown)


anulman:
@samselikoff curious what you think, or if you have recommendations re: best course of action?

lukemelia:
I ran into this same issue.

n8n8baby:
So, not sure why this is rotting on the vine, but I just noticed it after submitting a PR implementing a slightly different solution.
https://github.com/samselikoff/ember-cli-mirage/pull/1297

As long as it works, I don't really care, maybe a combination is best. Either way, please feel free to review and comment.

samselikoff:
Sorry for the delay, I'm trying to triage issues and hopefully will come back to this soon

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1337 supercedes

PR #1156, opened 2017-08-02
WIP: fix speed issue, cloning problem

kellyselden:
cc @dingoeatingfuzz

This branch was very successful in my app at speeding things up. It is lacking tests, and I'm not at a point where I can add them. Anyone is free to take this over and bring it to completion.

If this PR is just noise, feel free to close.


samselikoff:
Believe we've had some speed improvements since, will close for now. Thanks :)

PR #1157, opened 2017-08-04
Fix serialization of sideloaded associations with embedded associations

npafundi:
Consider the following models, and note the serialization method:
```
// User (serializer sideloads posts)
export default Model.extend({
  posts: hasMany('post')
});

// Post (serializer embeds comments)
export default Model.extend({
  authors: hasMany('user')
  comments: hasMany('comments')
});

// Comment
export default Model.extend({
  post: belongsTo('post')
});
```

Assuming we have populated our database and we serialize a user, the
resulting serialization may look like this:
```
{
  user: {
    id: '1',
    name: 'John',
    postIds: ['1', '2']
  },
  posts: [
    { id: '1', title: 'Lorem', comments: [{ id: '1', text: 'first' }] },
    { id: '2', title: 'Ipsum', comments: [] }
  ]
}
```

Currently, there is a bug in Mirage which causes an error to be thrown
when attempting this kind of serialization:
`TypeError: Cannot read property 'filter' of undefined`

This commit fixes the bug, and adds multiples tests to ensure proper
serialization.

### DIFF
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -113,7 +113,7 @@ class Serializer {
     }
 
     if (this.embed) {
-      return [hash];
+      return [hash, []];
 
     } else {
       let addToIncludes = _(serializer.getKeysForIncluded())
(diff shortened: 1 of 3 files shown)


npafundi:
Hey @samselikoff, just curious if there is any movement on this or #1160. I'm sure you're busy, but let me know if I can answer any questions on these two!

PR #1160, opened 2017-08-15
Improve detection/determination of inverse associations

npafundi:
`inverseFor` currently throws an error when there are multiple potential
inverse associations defined, unless they are all explicit.

It is possible to have multiple explicit and implicit inverses on the
same model, and still detect the correct associations.

Consider the following example:
```
// User
posts: belongsTo('post', { inverse: 'authors' })

// Post
authors: hasMany('user', { inverse: 'posts' }),
editor: belongsTo('user'),
```

`user.posts` is explicitly defined as the inverse of `post.authors` (and
vice-versa), so we know these are meant to be inverse associations.

This commit updates `inverseFor` to first choose the explicit inverse
(if provided), before defaulting to implicit inverses. It will also
throw an error if multiple explicit inverses are defined with the same
key, or if multiple implicit inverses are defined.

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -233,26 +233,31 @@ class Model {
   inverseFor(association) {
     let associations = this.schema.associationsFor(this.modelName);
     let modelName = association.ownerModelName;
+    let candidates = _values(associations).filter(association => association.modelName === modelName);
 
-    let theInverse = _values(associations)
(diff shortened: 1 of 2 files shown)


samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1286 closes

PR #1173, opened 2017-09-04
Omghax fix 1112

samselikoff:


lukemelia:
Thanks for working on this @omghax and @samselikoff. After upgrading to this commit, my polymorphic belongsTo relationship is no longer being serialized in the relationships node of a response the way it was prior to this change. I assume that is not intentional?

samselikoff:
@lukemelia no, most likely not... can you post a screenshot of the relevant models & response perhaps?

lukemelia:
@samselikoff. Sure, here you go. Let me know if you need me to synthesize to a failing test.

Models on the left, serializers on the right:

image

Response before updating:

image

Response after updating (note that in each item under `data`, the `relationships.list-parent` node is missing):

image

The relevant mirage/config.js code:

image

PR #1174, opened 2017-09-04
Fixes #1175 - Error when embedding included polymorphic hasMany items

### LINKED ISSUE #1154: Polymorphic relationships
## Description
For polymorphic relationships the foreign keys on the `hasMany` side end up containing
```
{ type, id }, id
```
which results in an error when trying to lookup the models.

### Twiddle
https://ember-twiddle.com/570ae58cf87abee21b8b59a0c868feb0

### Affected versions
- 0.3.0 - 0.3.2 fail with #1057 
- 0.3.3 - 0.3.4 fail with
> Mirage: Your GET handler for the url /timelines threw an error: Cannot read property ‘replace’ of undefined

as `type` is undefined here
```
          if (association.isPolymorphic) {
            found = foreignKeys.map(function (_ref2) {
              var type = _ref2.type;
              var id = _ref2.id;

              return _this5._schema.db[(0, _emberCliMirageUtilsNormalizeName.toCollectionName)(type)].find(id);
            });
            found = (0, _lodashCompact['default'])(found);
          }
```

### Not affected versions
- 0.2.9

### LINKED ISSUE #1175: hasMany polymorphic relationship error
I had originally believed the issue I was getting was the same as #1154, but that's the same error message coming from a different location.

When using polymorphic relationships and having a serializer which has both `embed: true` and an `include` for the polymorphic `hasMany` field, mirage gets an error coming from the `camelize` call here:

```
// addon/serializer-registry.js
serializerFor(type, { explicit = false } = {}) {
  let SerializerForResponse = this._serializerMap && (this._serializerMap[camelize(type)]);
  // ...
}
```

The `type` is coming through `undefined` and `camelize` attempts to call `replace` on it, which causes an error.

This ultimately tracked back to the code trying to read the value of `type` from `modelName` on a `PolymorphicCollection`, which doesn't have that property.

Apologies for not including a reproduction twiddle but the boilerplate seems broken (get an error about it being a private gist), and attempting to use mirage 0.3.4 in ember-twiddle currently seems to get a build error for the addon.  There is already a PR in to fix this though (#1174), with a reproduction case added as a test.

jon-hall:
Allows embedding of included polymorphic hasMany items.

### DIFF
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -103,13 +103,32 @@ class Serializer {
   }
 
   getHashForResource(resource, removeForeignKeys = false, didSerialize = {}, lookupSerializer = false) {
-    let hash;
-    let serializer = lookupSerializer ? this.serializerFor(resource.modelName) : this; // this is used for embedded responses
+    let hash,
+      serializer;
+
+    if (!lookupSerializer) {
+      serializer = this; // this is used for embedded responses
+    }
+
+    // PolymorphicCollection lacks a modelName, but is dealt with in the map
+    // by looking up the serializer on a per-model basis
+    if (lookupSerializer && resource.modelName) {
+      serializer = this.serializerFor(resource.modelName);
+    }
 
     if (this.isModel(resource)) {
       hash = serializer._hashForModel(resource, removeForeignKeys, didSerialize);
     } else {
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you and sorry for the delay!

PR #1176, opened 2017-09-15
Remove mention of bower in install guide

SebastianPlace:
As far as I'm aware, bower isn't required to run this addon anymore.


locks:
Thanks :)

PR #1181, opened 2017-09-26
Add support for arrays in the x-www-form-urlencoded form

marxsk:
If multiple options are selected than encoded form contains field[]=value1&field[]=value2&...
Previously, the last value overwrote the older content, now the output is "field: [value1, value2]"

It makes sense to use 'decodedKey' also for the rest of the keys but it may introduce incompatibility so I did not attempt to do it. Alternatively, it is possible to just remove last four bytes from the original key and left rest untouched.


samselikoff:

PR #1184, opened 2017-10-02
serializer blueprint should extend the application serializer

### LINKED ISSUE #1165: mirage-serializer blueprint should extend application serializer
When running `ember generate mirage-serializer foo` I would expect the serializer to extend the Application serializer like mentioned in the docs. It may cause confusion that the blueprint doesn't follow this advice.

ChrisBarthol:
Fixes #1165

### DIFF
--- a/blueprints/mirage-serializer/files/__root__/serializers/__name__.js
+++ b/blueprints/mirage-serializer/files/__root__/serializers/__name__.js
@@ -1,4 +1,4 @@
-import { JSONAPISerializer } from 'ember-cli-mirage';
+import ApplicationSerializer from './application';
 
-export default JSONAPISerializer.extend({
+export default ApplicationSerializer.extend({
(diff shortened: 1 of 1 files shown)


blueprints/mirage-serializer/files/__root__/serializers/__name__.js
```diff
@@ -1,4 +1,4 @@
-import { JSONAPISerializer } from 'ember-cli-mirage';
+import BaseSerializer from './application';
```
locks: I think we should go with `ApplicationSerializer` for consistency.

ChrisBarthol:
@locks I went with Base Serializer to match what was in the current docs.  I do think it should be `ApplicationSerializer` and updated this PR to reflect that.  I also opened a PR #1186 to update the docs as well.

PR #1185, opened 2017-10-02
update error message for 3.x version

wismer:
This is just cosmetic. Though the docs are largely unchanged, users could navigate to the link thinking they're looking at 0.3.x docs when it's 0.2.x.

Apologies if there was another PR addressing this. I didn't see it if when I tried searching for it.


samselikoff:
thank you!

PR #1187, opened 2017-10-03
Base Class.extend() on native es class syntax

cowboyd:
The extend mechanism, which we inherited from Ember pre-dates the advent of native ES class syntax. Among other things, this meant manually constructing the prototype chain when expressing the inheritance.

This change lets the `class` syntax do the heavy lifting of prototype wiring while making the code more clear as to its intent.

As a bonus, instances created with this class show up in development tools as `class Child` instead of

```js
function () {
  return parent.apply(this, arguments);
}
```

Still not ideal, but better than the alternative.

Finally, the `__super__` property was removed since it's of dubious value and did not cause any tests to fail.

### DIFF
--- a/addon/utils/extend.js
+++ b/addon/utils/extend.js
@@ -1,44 +1,29 @@
 import _assign from 'lodash/assign';
 import _has from 'lodash/has';
 
-export default function(protoProps, staticProps) {
-  let parent = this;
-  let child;
+export default function extend(protoProps, staticProps) {
+  let Parent = this;
 
-  // The constructor function for the new subclass is either defined by you
-  // (the "constructor" property in your `extend` definition), or defaulted
-  // by us to simply call the parent's constructor.
-  if (protoProps && _has(protoProps, 'constructor')) {
-    child = protoProps.constructor;
-  } else {
-    child = function() {
-      return parent.apply(this, arguments);
-    };
+  class Child extends Parent {
+    constructor(...args) {
+      super(...args);
+      // The constructor function for the new subclass is optionally defined by you
+      // in your `extend` definition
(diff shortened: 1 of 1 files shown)


samselikoff:
I would love to use ES class systems, the current extend module is not actually from Ember, it's copied from Backbone's source.

At the origin of Mirage we had classes, but I was unwilling to adopt them because we had no class fields. I wanted Mirage's API to "feel like" Ember so it relies heavily on these sorts of things:

```js

export default Model.extend({
  users: hasMany()
});
```

I didn't want (and still don't want) folks to have to do something like

```js
export default Model.extend({
  constructor() {
    super();
    this.users = hasMany();
  }
});
```

However it is my understanding that class fields gives us something like this

```
export default class extends Model {
  users = hasMany();
}
```

which is more tolerable. Is that true? (What will Ember Data's API look like once Ember moves over to ES classes?)

samselikoff:
Also `super` is a much-wanted feature in things like route handlers, so this would be a benefit from moving to ES classes.

cowboyd:
>However it is my understanding that class fields gives us something like this
>
>export default class extends Model {
>  users = hasMany();
>}
>which is more tolerable. Is that true?

Yup, though it's currently stage-2.

This change won't make a difference for end users, but it did make both understanding and debugging a lot clearer.

rwjblue:
Yep, sorry @cowboyd for co-opting your PR. 🌪

samselikoff:
@rwjblue this change led to this bug: https://github.com/samselikoff/ember-cli-mirage/issues/1211

Any ideas there on how to fix? It has something to do with Babel and the changes from this PR.

PR #1188, opened 2017-10-09
Support `Class.extend()` in ES6 environments

dfreeman:
If a dev disables support for IE in their `targets.js`, Babel will leave ES6 classes untranspiled. This is great for debugging, but unfortunately the `extend` implementation in Mirage blows up when trying to invoke the base constructor for a native class:

image

Babel has a `classCallCheck` that it inserts in transpiled constructors to try to capture this behavior, but it's not full fidelity, which is why this only shows up with native classes.

The fix I'm proposing here uses class syntax to set up the inheritance instead of wiring everything together by hand, which avoids the issue since it can then call actual `super` in the constructor. This is _almost_ a drop-in replacement (and the test suite still all passes), but unfortunately there are some subtle differences in observable behavior:
 - custom `constructor()` methods now have the super constructor invoked before they're run—this is unavoidable since you have to call `super` before you're allowed to touch `this` in a constructor
 - static properties from the parent class are now available on the child via the prototype chain rather than being copied over directly (that's just how ES6 static inheritance works, though we _could_ work around it)

The second bullet point doesn't seem like a big deal, but the first might be significant enough to block a change like this from landing until the next breaking release—not sure. Either way, I wanted to at least raise the issue since I came across it, and I suspect others will as folks start to play more with the targets stuff in development.

### DIFF
--- a/addon/utils/extend.js
+++ b/addon/utils/extend.js
@@ -3,38 +3,20 @@ import _has from 'lodash/has';
 
 export default function(protoProps, staticProps) {
   let parent = this;
-  let child;
-
-  // The constructor function for the new subclass is either defined by you
-  // (the "constructor" property in your `extend` definition), or defaulted
-  // by us to simply call the parent's constructor.
-  if (protoProps && _has(protoProps, 'constructor')) {
-    child = protoProps.constructor;
-  } else {
-    child = function() {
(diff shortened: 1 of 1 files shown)


dfreeman:
D'oh. I only checked issues, not open PRs. Just saw #1187.

PR #1191, opened 2017-10-15
Bump ember-cli, ember-cli-babel

mixonic:
Bump to the latest version of Ember-CLI which includes an ember-cli-babel bump to support the latest of the new modules API.

The small whitespace changes and formatting changes are from `ember init`, should make future diffs smaller.


rwjblue:
Looks like most of the failures are ESLint related.

mixonic:
@rwjblue yeah I'm pushing a new commit now that looks better.

I did need to bump Ember-Data to avoid an "unmet peer dependency" error for ember-inflector. I tried bumping to 2.16 but that yielded a number of errors. This commit only bumps to 2.13 and tests pass locally.

rwjblue:
FWIW, the number of builds is pretty incredible. I created https://github.com/samselikoff/ember-cli-mirage/issues/1192 to try to provide guidance on a path forward...

Issue #1099, opened 2017-06-04
Bug - attempting to use ID 1 but its already used


samselikoff:
```
  afterCreate(clip, server) {
    clip.encodes = [
      server.create('encode', 'mp4'),
      server.create('encode', 'webm')
    ];
    clip.save();
  },

...
clip.encodes.destroy();

...

      clip.update({
        encodes: [
          server.create('encode', 'longMp4'),
          server.create('encode', 'longWebm')
        ]
      });
```

geekygrappler:
Hey @samselikoff I couldn't reproduce this: https://github.com/geekygrappler/ember-cli-mirage/commit/44162b8a8d634827cc2a656bf53551870727fa0d

Created new children with correct IDs in my test, didn't try and reuse id `1`. Any further steps to reproduce?

erichonkanen:
Just ran into this, not sure if it's ember-data or mirage but I create a new record with a specific id, then I delete it, then I recreate the record with that specific id and I get ```Attempting to use the ID current-user|test-user-0, but it's already been used```

samselikoff:
@erichonkanen That's on Mirage's side - it's how the default [identity manager](http://www.ember-cli-mirage.com/docs/advanced/identity-managers) works (check out [the set method](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/identity-manager.js#L33)).

What's your use case?

PR #1193, opened 2017-10-17
Reuse ID after record is removed

thorsteinsson:
Be able to reuse an ID if the record has been removed.

Might be related to #1099.

### DIFF
--- a/addon/db-collection.js
+++ b/addon/db-collection.js
@@ -221,12 +221,16 @@ class DbCollection {
     } else if (typeof target === 'number' || typeof target === 'string') {
       let record = this._findRecord(target);
       let index = this._records.indexOf(record);
+
+      this.identityManager.remove(record.id);
       this._records.splice(index, 1);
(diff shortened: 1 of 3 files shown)


ondrejsevcik:
I don't think this is a good idea. This may lead to bugs in your tests. Mirage tries to be "database-like" storage and databases usually doesn't reuse ids.

thorsteinsson:
It only reuses the ID if you specify it when you create the record otherwise it will get a new ID. That means that this will not break any existing functionality, only add the ability to use removed IDs by specifying the ID in your create call.

I noticed this while upgrading from 0.1 to 0.3 and later found out that I can use a custom identity manager to solve this. I would not encourage reusing IDs but I'm questioning if Mirage should have that restriction.

samselikoff:
This should be possible by making a custom IdentityManager https://miragejs.com/docs/advanced/mocking-guids

PR #1201, opened 2017-10-27
Namespaced model and subdirectory support.

patsy-issa:
The current implementation of ember-data model detection does not
account for nested models i.e: `models/namespaced/user.js` resulting in `models/user.js` and `models/namespaced/user.js` being registered as the same and throwing an error when you try to create `namespaced/user` in scenarios.


Patsy-issa:
Relationship lookup needs to be changed as well, working on it.

Patsy-issa:
Small update: have to drop work on this for the time being, will try to pick the work back up as soon as I have some free time.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1785 closes

Issue #983, opened 2017-01-12
support ember-network fetch mocking


izelnakri:
This might not be common but I still believe an essential feature. Mirage/pretender should support fetch() requests. 

I have one component that doest fetch request to a model endpoint however the endpoint is dynamic sometimes it is the same as the ```person``` model endpoint

I have this:
```js
// in mirage/config.js
this.get('/people', ({people}, request) => {
    const params = request.queryParams;
    if (params.query) {
      return peopleSearch(people.all());
    } 
     
    return people.all();
});
```

I have this call in a components:
```js
// in components/foo.js

export default Component.extend({
    searchPath: computed('attribute', function() {
        const entityType = this.get('entityType');

        if (entityType === 'person') {
            return '/people';
        } else {
            return '/application';
        }
    },
    keyUp() {
        let query = $(event.target).val();

        if (query.length > 0) {
            // IMPORTANT PART:
            fetch(config.apiHost + `${this.get('searchPath')}?query=${query}`)
            // I do more with it 
        }
    }
});
```

this fetch makes a request to the same endpoint as Ember 'person' model adapter. It would be great if we can make mirage support fetch calls. Currently I have to mock this with something like: https://github.com/wheresrhys/fetch-mock and duplicate the ember mirage code/logic.

P.S I will always use fetch instead of $.get() because that is what fastboot supports.

morhook:
I think this ticket is related with how pretender could do it https://github.com/pretenderjs/pretender/issues/60

samselikoff:
After some quick reading it looks like @stefanpenner recommends `ember-fetch`, could somebody verify this?

If so I think that should be the solution, `ember-fetch` polyfills `fetch` but never falls back to default (as of today), so Pretender should always work.

izelnakri:
Yes, this is indeed the case, I think ```ember-network``` will get replaced by ```ember-fetch```. ```ember-network``` also has caveats regarding its then/catch promise handling.

lypborges:
I have a similar use case here, it's a dashboard component that uses `ember-fetch` to load data. Is there any way to help with this or should we wait for pretender to be updated?

0xadada:
so i'm working on an app where we've
1) removed jQuery as a dependency
2) migrated to `ember-fetch` as a replacement for `$.ajax` / `XMLHTTPRequest`
  2a) Configured `fastboot` to use `node-fetch`
[...]

PR #1206, opened 2017-11-01
refactor(jQuery): Remove $.ajax, use async/await, move custom paths t…

geekygrappler:
…o their own test

- I wanted to remove $.ajax, so replaced it with fetch
- Used async/await hopefully makes the tests easier to follow
- We were testing custom paths using `posts` model in every test, moved testing of this behaviour into it's own test and removed from all other tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.0",
     "ember-fastboot-addon-tests": "^0.4.0",
+    "ember-fetch": "^3.4.3",
     "ember-load-initializers": "^1.0.0",
     "ember-resolver": "^4.0.0",
     "ember-sinon": "0.5.1",
--- a/tests/integration/server/resource-shorthand-test.js
+++ b/tests/integration/server/resource-shorthand-test.js
@@ -1,6 +1,7 @@
 import {module, test} from 'qunit';
 import { Model, ActiveModelSerializer } from 'ember-cli-mirage';
 import Server from 'ember-cli-mirage/server';
+import fetch from 'fetch';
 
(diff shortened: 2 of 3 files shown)


geekygrappler:
It was bugging me that we had $.ajax in there and were not using async/await, so I fixed it for this file. It does introduce `ember-fetch` though, maybe you don't want that dependency @samselikoff?

rwjblue:
This seems good to me, but generally speaking the only reason `ember-fetch` works is because it still uses `XMLHttpRequest` under the hook (e.g. it is not a "true" polyfill for native `fetch`). We need to create a pretender adapter for `fetch` (I think @cibernox had a spike of this at one point).

I'm _slightly_ concerned about landing these changes considering that I am _also_ pushing to use native `fetch` if the host apps targets support it.

Some background links:

* https://github.com/pretenderjs/pretender/issues/60
* https://github.com/ember-cli/ember-fetch/issues/45
* https://github.com/ember-cli/ember-fetch/pull/63
* https://github.com/samselikoff/ember-cli-mirage/issues/983

cibernox:
I did. I can't locate it, but it's nothing I can't reproduce within a couple hours. The web community in general, not only the Ember community, needs a fetch stub that works.

geekygrappler:
Thanks for taking a look @rwjblue.

> I am also pushing to use native fetch if the host apps targets support it.

Does that require some work on Pretender first to intercept native `fetch` requests?

From reading the links, I guess my only concern now would be that `ember-fetch` stops using `XMLHttpRequest`, and uses native `fetch` which would mean these tests break... I think.

In any case I think I'll split this into 3 different PRs. I think we can get the async/await changes and the moving of the custom paths merged fairly easily. The `fetch` stuff can wait.

rwjblue:
Sounds good, thanks for working on this!

Issue #1207, opened 2017-11-02
Records do not get updated if the parent is not changed


mupkoo:
I tried to dig deeper into this and it looks like if you try to update a record without changing the parent, it gets saved twice. This is what happens.

```javascript
let schema = new Schema(db, {
  author: Model.extend({
    posts: hasMany()
  }),
  post: Model.extend({
    author: belongsTo()
  })
});

let author = schema.create('author');
let post = schema.create('post', { authorId: author.id });

post.update({ title: 'Updated title' });
```

Once you do this, `post.save` gets called for the first time. This triggers `author.save()`, which eventually triggers `post.save` again, but this time, the post has the old attributes, so re-updates with the old attributes, as `author.posts` contains clones of the `post` records.

mupkoo:
This commit is causing the problem -> https://github.com/samselikoff/ember-cli-mirage/commit/85a6ac052f38ffd80f41b6d8e87332a680d33740

mupkoo:
I tried to dig deeper into this and it looks like if you try to update a record without changing the parent, it gets saved twice. This is what happens.

```javascript
let schema = new Schema(db, {
  author: Model.extend({
    posts: hasMany()
  }),
  post: Model.extend({
    author: belongsTo()
  })
});

let author = schema.create('author');
let post = schema.create('post', { authorId: author.id });

post.update({ title: 'Updated title' });
```

Once you do this, `post.save` gets called for the first time. This triggers `author.save()`, which eventually triggers `post.save` again, but this time, the post has the old attributes, so the updates does not happen.

mupkoo:
If I remove the check introduced by https://github.com/samselikoff/ember-cli-mirage/commit/85a6ac052f38ffd80f41b6d8e87332a680d33740, only a single `save` is called, and it is the `post.save`.

demsullivan:
I'm seeing this too, but it's not 100% of the time. Instead of updating title, I'm updating the relationship, e.g.:

```
let schema = new Schema(db, {
  author: Model.extend({
    posts: hasMany()
  }),
  post: Model.extend({
    author: belongsTo()
  })
});

let author = schema.create('author');
let secondAuthor = schema.create('author');
let post = schema.create('post', { authorId: author.id });

post.update({ authorId: secondAuthor.id });
```

I've traced it to https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L91-L95

[...]

Issue #1061, opened 2017-04-08
One-to-One relationship causes stack size exceeded


timhaines:
Here's a reproduction in a [simple ember app](https://github.com/timhaines/mirage-recursive-bug)

There are two models - organization and subscription, that have a 1-1 belongsTo relationship. When Mirage tries to load the organization model as a result of the schema.organizations.all() call, the maximum stack size exceeded error occurs.

When I last dug into this, it appeared the problem was related to the model get and set methods in belongsTo.js.  In my basic understanding, when the model is being created, within _setupRelationships, setting the model calls `inversesAlreadyAssociated`, which results in a call to getting the model, which in turn calls `find` on the model in this line:
`model = association.schema[toCollectionName(association.modelName)].find(foreignKeyId);`

timhaines:
I've created a twitter here: https://ember-twiddle.com/072de99410f21ff42db8264bd9665ca3?openFiles=mirage.factories.user.js%2C

timhaines:
I've created a Twiddle that shows the error in the javascript console here: https://ember-twiddle.com/59cc935d15d813e36e8d03a7145649e6?openFiles=mirage.factories.user.js%2C

samselikoff:
We tracked it down to [these lines](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L81-L84), looks like it's a bug with the `Model.update` method.

`_definePlainAttribute` is being called for every key in the object passed to `update`, but it should be skipped if a key is a relationship (which would be known ahead of time, since relationships are defined on the model). So this code just need to have an `if` statement to skip `definePlainAttribute` for relationship keys.

Would love for someone to PR this change with a test, should be relatively straightforward!

PR #1208, opened 2017-11-02
Update without changing the parent

mupkoo:
This is just a failing test, related to #1207 

Looks like the problem is not with the none numeric IDs, but with updates that do not involve updating the parent.

### DIFF
--- a/tests/integration/orm/belongs-to/1-basic/update-test.js
+++ b/tests/integration/orm/belongs-to/1-basic/update-test.js
@@ -1,0 +1,44 @@
+import Helper from './_helper';
+import { Model, belongsTo, hasMany } from 'ember-cli-mirage';
+import Schema from 'ember-cli-mirage/orm/schema';
+import { module, test } from 'qunit';
+
+module('Integration | ORM | Belongs To | Basic | update', function(hooks) {
+  hooks.beforeEach(function() {
+    this.helper = new Helper();
+
+    this.helper.schema = new Schema(this.helper.db, {
+      author: Model.extend({
+        posts: hasMany()
+      }),
+      post: Model.extend({
+        author: belongsTo()
+      })
+    });
+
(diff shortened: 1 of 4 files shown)


tests/integration/orm/belongs-to/1-basic/update-test.js
```diff
@@ -0,0 +1,44 @@
...
+import Schema from 'ember-cli-mirage/orm/schema';
+import { module, test } from 'qunit';
+
+module('Integration | ORM | Belongs To | Basic | update', function(hooks) {
+  hooks.beforeEach(function() {
+    this.helper = new Helper();
+
+    this.helper.schema = new Schema(this.helper.db, {
+      author: Model.extend({
+        posts: hasMany()
```
mupkoo: If I comment out this part, the test passes.

mupkoo:
If I remove the changes introduced by this commit - https://github.com/samselikoff/ember-cli-mirage/commit/85a6ac052f38ffd80f41b6d8e87332a680d33740, the tests pass as well.

mupkoo:
I added a test case for #1061 and a fix for the problems that this is causing.

Let me know what do you think. 

PS I wonder how nobody complained about the update not working...

mupkoo:
@samselikoff let me know if you can help some more with this 👍

kennethdavidbuck:
@mupkoo thank you for this. I was running into what I thought were timing issues as well and this update fixes things for me. @samselikoff curious to see if there are plans to get this PR into master/next release?

samselikoff:
I lost track of all the threads, but just checked this out, added only the tests, and they don't fail anymore. I believe we fixed this bug elsewhere (and when I ran the changes included in this PR they broke some new tests.

So, I'm going to assume the bug is fixed & I'll close this for now! Feel free to ping if still having trouble 😄

PR #1212, opened 2017-11-07
refactor: move custom path behaviour into one test

geekygrappler:
We were testing custom path behaviour in every test here. I've moved the tests for custom path behaviour into one test. The third part of #1206.


rwjblue:
Thank you!

Issue #1163, opened 2017-08-22
discoverEmberDataModels and relationships


caseywatts:
I was just trying out the feature `discoverEmberDataModels`, and I love it 🎉 

Maybe it doesn't cover this use case yet though:

In my `plan` model, I have a relationship like this:
`  submitter: DS.belongsTo('user'),`

and I get the error
`
Error: Mirage: You're trying to create a repricing model and you passed in a model:user(1) under the submitter key, but you haven't defined that key as an association on your model.Mirage: Error: You're trying to create a repricing model and you passed in a model:user(1) under the submitter key, but you haven't defined that key as an association on your model.
`

cc @offirgolan

scottkidder:
I think I ran into this when I still had a `mirage/models/plan.js` (in your case).  Does that file exist?

caseywatts:
`mirage/models/plan.js` doesn't exist for us~

We have only these two mirage model files:
- the one that I mentioned above
- one for this similarly-shaped situation - we have a `mirage/models/addon.js` just so we can have this line: `  documentations: DS.belongsTo('documentation'),`

caseywatts:
In ember-data model definitions, the key isn't necessarily the model type - it's sorta an alias (at least from this perspective of model definitions). The string passed to `belongsTo` is more reliably the model type. For example with `submitter: DS.belongsTo('user')`, submitter is the ~alias and `user` is the model type.

I expect this situation is here because we're doing an `Object.keys()` to iterate through relationships (but I'm having trouble reading the code).

Somewhere around here:
https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/ember-data.js#L72

offirgolan:
The Object.keys is not iterating through relationships. It is iterating through all the app's defined models where the key is the model's name and the value is the DS.Model class.

If you look a few lines below, you will notice we are using ember data's public api `eachRelationship` to iterate over each of the model's relationships.

caseywatts:
@offirgolan that tip helped me read the source code so much, thanks! :)

I've got it all working now - it does totally cover this use case already :)

I traced my issue back to a mirage serializer I had, which had an `includes: ['user']`.

PR #1214, opened 2017-11-08
Ember data auto-detection - model name mismatch test

caseywatts:
I was concerned this new feature didn't cover [this edge case](https://github.com/samselikoff/ember-cli-mirage/issues/1163#issuecomment-342935214) - but it does :)

While I was in there, I made these two tweaks:

- add tests that make it clearer that this edge case is covered.
- a minor readability refactor


geekygrappler:
@caseywatts unfortunately adding two fields with the same `belongsTo` breaks that Model for other tests:
`Error: Assertion Failed: You defined the 'address' relationship on (subclass of DS.Model), but multiple possible inverse relationships of type (subclass of DS.Model) were found on (subclass of DS.Model). Look at http://emberjs.com/guides/models/defining-models/#toc_explicit-inverses for how to explicitly specify inverses`

Maybe set up a new Model in the app that isn't used elsewhere to run these assertions? Or define the inverses?

caseywatts:
oop good catch - done! ✨

samselikoff:
cheers! 🍻

PR #1215, opened 2017-11-09
Promise ajax

geekygrappler:
Wraps Ajax with a promise... Paves the way for `async/await`... paves the way to drop in fetch in the future maybe.

### DIFF
--- a/tests/helpers/promise-ajax.js
+++ b/tests/helpers/promise-ajax.js
@@ -1,0 +1,9 @@
+import { Promise } from 'rsvp';
+
+export default (options) => {
+  return new Promise((resolve, reject) => {
+    $.ajax(options)
+      .done((data, status, xhr) => resolve({data, status, xhr}))
+      .fail((xhr, status, error) => reject({xhr, status, error}));
+  });
+};
--- a/tests/integration/http-verbs-test.js
+++ b/tests/integration/http-verbs-test.js
@@ -1,6 +1,7 @@
 import {module, test} from 'qunit';
 import Server from 'ember-cli-mirage/server';
 import { Model } from 'ember-cli-mirage';
+import promiseAjax from '../helpers/promise-ajax';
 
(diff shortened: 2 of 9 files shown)


.eslintrc.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
     'plugin:ember-suave/recommended'
   ],
   env: {
-    browser: true
+    browser: true,
+    es6: true
```
geekygrappler: So there's no eslint error for `Promise`

addon/utils/promise-ajax.js
```diff
@@ -0,0 +1,7 @@
+export default (options) => {
+  return new Promise((resolve, reject) => {
```
rwjblue: I don’t think we can rely on native Promise (unless we don’t support things like older IE and whatnot). We might need to add `import { Promise } from 'rsvp’`...

addon/utils/promise-ajax.js
```diff
@@ -0,0 +1,7 @@
+export default (options) => {
```
rwjblue: If this file is only used in our own tests, can you move it to tests/helpers/?

geekygrappler: :+1:

addon/utils/promise-ajax.js
geekygrappler: Can rely on native Promise if we're only using this as a test helper? Or will people run the tests against IE? Not something I really know about.

jherdman: I'll raise my hand here. I build my tests against IE11.

geekygrappler: Thanks @jherdman. I'll use rsvp.

geekygrappler:
Should be good to go now @rwjblue.

rwjblue:
Thank you!

Issue #1200, opened 2017-10-26
Track requests with Pretender in test mode


DingoEatingFuzz:
By default Pretender tracks requests made during the lifespan of the Pretender server. This is a really handy way to spy on requests and assert behaviors in tests.

A simple example:

https://github.com/hashicorp/nomad/blob/master/ui/tests/acceptance/namespaces-test.js#L21-L28

A more complex example:

https://github.com/hashicorp/nomad/blob/master/ui/tests/acceptance/token-test.js#L45-L56

However, in #1137 `trackRequests` was hardcoded to be false.

I think the best way forward is to track requests in tests by default, don't track requests in dev by default, and expose a property like `server.logging` to control whether or not requests should be tracked.

geekygrappler:
Fixed in #1217

PR #1217, opened 2017-11-13
make pretender request tracking configurable

patrickjholloway:
Track requests with Pretender in test mode [#1200](https://github.com/samselikoff/ember-cli-mirage/issues/1200)

Our app depends on pretender tracking for assertions in our tests about outgoing requests. It would be great if this were configurable.

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -61,7 +61,7 @@ function createPretender(server) {
          mirage/config.js file. Did you forget to add your namespace?`
       );
     };
-  }, { trackRequests: false });
+  }, { trackRequests: server.shouldTrackRequests() });
 }
 
 const defaultRouteOptions = {
@@ -154,6 +154,7 @@ export default class Server {
     this.timing = this.timing || config.timing || 400;
     this.namespace = this.namespace || config.namespace || '';
     this.urlPrefix = this.urlPrefix || config.urlPrefix || '';
+    this.trackRequests = config.trackRequests;
 
     this._defineRouteHandlerHelpers();
 
@@ -181,6 +182,9 @@ export default class Server {
     let hasFactories = this._hasModulesOfType(config, 'factories');
     let hasDefaultScenario = config.scenarios && config.scenarios.hasOwnProperty('default');
 
+    let didOverridePretenderConfig = (config.trackRequests !== undefined) && this.pretender;
+    assert(!didOverridePretenderConfig,
+      'You cannot modify Pretender\'s request tracking once the server is created');
     this.pretender = this.pretender || createPretender(this);
 
     if (config.baseConfig) {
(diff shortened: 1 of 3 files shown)


addon/server.js
```diff
@@ -154,6 +154,7 @@ export default class Server {
     this.timing = this.timing || config.timing || 400;
     this.namespace = this.namespace || config.namespace || '';
     this.urlPrefix = this.urlPrefix || config.urlPrefix || '';
+    this.trackRequests = this.trackRequests || config.trackRequests || (this.environment === 'test');
```
geekygrappler: Maybe default to false rather than true in all tests. It's probably set to false originally for some reason.

geekygrappler:
Is it possible to add a test for this? Maybe here: https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/server-config-test.js.

patrickjholloway:
@geekygrappler updated PR per comments

geekygrappler:
Seems to be an ESlint error
`not ok 2156 Chrome 62.0 - ESLint | app: changing the trackRequests configuration of the server throws an error` it wasn't clear to me what the problem was from CI

patrickjholloway:
@geekygrappler sorry about that, should be all set now. I didn't have the latest from master running tests locally, reproduced lint test failure after rebasing.

geekygrappler:
@samselikoff Originally hardcoded `trackingRequests` for pretender to false due to memory leaks #509. This fix leaves it defaulted to false, but gives users the option to turn it on if they need it. What do you think?

@patrickjholloway could also add a note about this option here: https://github.com/samselikoff/ember-cli-mirage/blob/gh-pages/docs/v0.3.x/configuration.md 
(I'm also happy to do that if/when it gets merged)

samselikoff:
Can we not make it possible to set this after server instantiation? Thinking folks might want to turn it on in a test (after server was created) and then assert against the requests.

patrickjholloway:
@samselikoff trackRequests has to be set when pretender is created. From what I can tell that would mean rebuilding all the handlers on pretender between each test and some considerable refactoring. Right now Pretender only gets created when startMirage gets called in the initializer.

geekygrappler:
Coincidentally at work we're also using `trackRequests`. Trying to upgrade to `0.4` and this is blocking us. I've not looked into a way around it yet. But I'm now more in favour of letting those people who don't mind/care about the memory issues switching it on if they want. (I was in favour of it before I discovered my need for it as well 😉)

patrickjholloway:
I think if we are still concerned about the memory issues we could upgrade to the latest pretender. https://github.com/pretenderjs/pretender/commits/v1.6.1 It seems like they have addressed some, if not all, of the memory issues.

I would be happy to make a separate PR for that.

geekygrappler:
Hey @patrickjholloway I can now merge on this project. I'd like to get the pretender upgrade in first, are you happy to submit that PR?

PR #1221, opened 2017-11-19
refactor: use async/await in tests

geekygrappler:
Last part of #1206.


geekygrappler:
I should probably say to save having to read #1206. This simply replaces `.then` callbacks with `async/await`.

geekygrappler:
Hey @rwjblue I'm going to merge this at the weekend, unless you have any comments/objections.

rwjblue:
👍 seems good to me!

Issue #1223, opened 2017-11-24
Cannot read property 'modelName' of null


Shravan-Joopally:
ember-cli-mirage: 0.4.0

Getting "Cannot read property 'modelName' of null" error when serializer included associated resource is null.

I think there is null check missing

geekygrappler:
Hi @Shravan-Joopally, the fix looks good. 

Could you please explain to me in what situation an included associated resource can/will be null? Not a trick question, I just want to understand the situation to make sure having null on the associate resource is a correct state that we should allow.

ShravanJoopally-xo:
Hi its failing when I have embedded records as part of another model, and it is nul.
Please find below example
here I have 2 models model1 and model2. model1 get request will have model2 embedded record, so i specified embed true and added model2 as included in model2 serializer.
And model2 can be null.
 
model1.js
```
export default Model.extend({
   model2: belongsTo()
 });
```

model2.js
```
export default Model.extend({
 });

```
serializers/model1.js
```
import Serializer from './application';

 export default Serializer.extend({
  embed: true,
  include: ['model2']
 });
```

fixtures/model1.js
```
export default [{
  model2: 1
}, {
model2: null
}]
```

its throwing error when i access get /model1
its failing to seralize the second object in fixture as included resource is null

geekygrappler:
OK great @Shravan-Joopally. I understand now. Would you be able to write a test case for this?

ShravanJoopally-xo:
@geekygrappler sure will do

PR #1224, opened 2017-11-24
fixes issue #1223 added null check

Shravan-Joopally:
added null check.

Its throwing error when included property is null


samselikoff:
Thanks! Could we please get a test along with this change?

samselikoff:
Closing for now, believe 1289 fixed this

PR #1229, opened 2017-11-29
fix[request-tracking]: request tracking defaults to false

geekygrappler:
Sorry @samselikoff, #1217, changed the default for request tracking to be true, I made a mistake when reviewing it. I noticed when I upgraded our app to 0.4.1 so I have fixed it.


addon/server.js
```diff
@@ -244,7 +244,7 @@ export default class Server {
    * @public
    */
   shouldTrackRequests() {
-    return typeof this.trackRequests !== 'undefined' ? this.trackRequests : false;
+    return this.trackRequests ? this.trackRequests : false;
```
rwjblue: Isn't this essentially the same as either `return !!this.trackRequests` or `return Boolean(this.trackRequests)`?  I think either of those are a bit more idiomatic in this case...

geekygrappler: Agreed, either of those are better. I'll update.

PR #1232, opened 2017-12-01
Support coalesce option properly according to JSON:API spec

### LINKED ISSUE #1202: `coalesce` shorthand option does not comply with JSON:API / ember-data
The `coalesce` option of shorthand GET requests, introduced in #123 by @cibernox, expects query params in the form `?ids=1,3`. This however does not comply with the JSON:API [recommendation](http://jsonapi.org/format/#fetching-filtering) and the default serialization of the ember-data JSONAPI adapter's [`coalesceFindRequests`](https://www.emberjs.com/api/ember-data/2.16/classes/DS.JSONAPIAdapter/properties/coalesceFindRequests?anchor=coalesceFindRequests), which looks like `?filter[id]=1,3`.

It would be nice to have this match out of the box!

Changing that would probably not be very hard, but introduce a breaking change. Not sure how to solve this?
* do the change and accept a breaking change
* introduce another option like `coalesceWithFilter` (or whatever)
* use the filter-based serialization only when the `JSONAPIAdapter` is used for the given model. Would still be kind of a breaking change, although I am unsure if anybody was able to use the existing way with an JSON:API compliant API?

simonihmig:
Introduces a new `getCoalescedIds` method for serializers, to encapsulate knowledge on how to deserialize coalesced Ids from the given request in that layer, so we can support the proper way for each serializer independently.

Fixes #1202 

cc @cibernox

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -214,8 +214,15 @@ const JSONAPISerializer = Serializer.extend({
 
   typeKeyForModel(model) {
     return dasherize(pluralize(model.modelName));
-  }
+  },
 
+  getCoalescedIds(request) {
+    let ids = request.queryParams && request.queryParams['filter[id]'];
+    if (typeof ids === 'string') {
+      return ids.split(',');
+    }
+    return ids;
+  }
 });
 
 JSONAPISerializer.prototype.alwaysIncludeLinkageData = false;
--- a/addon/route-handlers/shorthands/get.js
+++ b/addon/route-handlers/shorthands/get.js
@@ -29,11 +29,13 @@ export default class GetShorthandRouteHandler extends BaseShorthandRouteHandler 
       } else {
         return model;
       }
-    } else if (this.options.coalesce && request.queryParams && request.queryParams.ids) {
-      return modelClass.find(request.queryParams.ids);
-    } else {
-      return modelClass.all();
+    } else if (this.options.coalesce) {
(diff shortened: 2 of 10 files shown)


simonihmig:
Tbh, I was not sure how  the `ActiveModelAdapter` was supposed to support this, if at all. But it seems to extend from `RestAdapter` (speaking of the ember-data adapter here), and one acceptance test (closed-friends) was relying on the same behaviour, so I made the `ActiveModelSerializer` behave the same as the `RestSerializer` here...

simonihmig:
the one failing test is unrelated, can someone restart that? The Chrome issue Travis had yesterday should have been fixed by now, dunno what's up there...

addon/serializers/json-api-serializer.js
```diff
@@ -214,8 +214,15 @@ const JSONAPISerializer = Serializer.extend({
 
   typeKeyForModel(model) {
     return dasherize(pluralize(model.modelName));
-  }
+  },
 
+  getCoalescedIds(request) {
+    let ids = request.queryParams && request.queryParams['filter[id]'];
+    if (typeof ids === 'string') {
+      return ids.split(',').map((id) => parseInt(id, 10));
```
lukemelia: We shouldn't assume integer IDs

simonihmig: @lukemelia Agree, good catch! Removed that!

simonihmig:
Anything missing here?

cibernox:
FWIW, this looks good to me.

samselikoff:
Very nice - thank you!!

PR #1235, opened 2017-12-06
Performance: Don't make copies of record arrays internally

DingoEatingFuzz:
Right now, whenever accessing a records property on the `db` object, a copy of all records is made. 

This has a purpose, to avoid the possibility of a user accidentally mutating the db state of Mirage, but it comes at a great cost.

```js
const records = server.db.myModel; // Makes a copy of the myModel records array
records.length = 0; // Destroys records

// Still fine, since the previous line was operating on a copy
const records2 = server.db.myModel; 
```

The cost is that every operation gets slower as the records array for a type grows. More seriously, the `createList` method has quadratic complexity.

```js
// Makes a copy of [ 1 ], [ 1, 2 ], ... [ 1, ..., 499 ]
const newRecords = server.createList('myModel', 500); 
```

This PR introduces a new private accessor for db collections that does not make a record copy. Mirage source is trusted to not accidentally mutate internal db state.

Since the private accessor doesn't make a copy, the performance of `createList` is now linear. Here is the before and after of the perf tests:

**Before**
screen shot 2017-12-06 at 3 49 20 pm

**After**
screen shot 2017-12-06 at 3 32 36 pm

### DIFF
--- a/addon/db.js
+++ b/addon/db.js
@@ -55,9 +55,29 @@ class Db {
       let IdentityManager = this.identityManagerFor(name);
       let newCollection = new DbCollection(name, initialData, IdentityManager);
 
+      // Public API has a convenient array interface. It comes at the cost of
+      // returning a copy of all records to avoid accidental mutations.
       Object.defineProperty(this, name, {
         get() {
           let recordsCopy = newCollection.all();
+
+          ['insert', 'find', 'findBy', 'where', 'update', 'remove', 'firstOrCreate']
+            .forEach(function(method) {
+              recordsCopy[method] = function() {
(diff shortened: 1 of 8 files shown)


DingoEatingFuzz:
Some real world impact:

**Before** 3:26
screen shot 2018-01-17 at 2 58 17 pm

**After** 2:12
screen shot 2018-01-17 at 3 19 17 pm

There is some assertion variability due to looping over data generated from factories, but the difference in assertions doesn't account for the difference in time.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1254 closes

Issue #1210, opened 2017-11-06
Consider a Noop Initializer when mirage is not enabled in a build


kennethdavidbuck:
Hey, first off, thank you for all the great work! 

I just came across an issue where one of my apps initializers had to run before the ember-cli-mirage initializer (to seed momentjs with a predefined "now"). Doing so created a scenario where everything would be fine in both the test and development builds (all tests green), but the apps production build would fail due to a missing initializer. I did some investigating and I believe it happens for the following reason:

In development/test builds ember will catch missing initializers that are referenced in an initializers "before" and "after" keys, however because mirage is excluded when its enabled flag is set to false, it is able to pass the assertions and then it disappear in a production build.

To get around the issue I am currently conditionally setting the before key in my apps initializer based on the ember-cli-mirage enabled configuration value. While this works, it seems like it would be better if consumers didn't have to think about this detail. One way to handle this could involve a noop initializer for the cases where mirage is excluded from the builds.

Please let me know what you think

Thanks!

rwjblue:
Agreed. We should always have an initializer, it should just do nothing when not included...

kennethdavidbuck:
@rwjblue thank you for the quick reply. If there is no immediate rush, then I would be up for attempting a PR for this. It looks like the initializer is already set up to handle the case where mirage is disabled, so I imagine there just needs to be a higher degree of specificity when excluding files from the build tree?

rwjblue:
Seems reasonable, though I haven't recently reviewed the build configuration here so am not 100% certain.

kennethdavidbuck:
Ok, I will take a look at it and see how far I can get. Thanks again.

kennethdavidbuck:
@rwjblue I made a first attempt at addressing this in #1236 and was wondering if you wouldn't mind taking a look at it / giving me some feedback?

PR #1236, opened 2017-12-07
Add noop initializer when mirage is excluded from build

kennethdavidbuck:
This is my first attempt at addressing #1210.

@rwjblue would you be up for taking an initial peak at what I have so far? I have taken a pretty simple approach to start with here. This is also the first time that I have attempted to write a test(s) to cover the state of a broccoli tree. Any / insight feedback would be appreciated. Thank you in advance.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,8 @@ const path = require('path');
 const mergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
 const map = require('broccoli-stew').map;
+const rm = require('broccoli-stew').rm;
+const replace = require('broccoli-replace');
 
 module.exports = {
   name: 'ember-cli-mirage',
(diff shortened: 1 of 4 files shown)


samselikoff:
This looks great!

Since the current initializer only runs `startMirage` if `shouldUseMirage` is true, do we need to replace it with the noop? (It essentially is a noop). That keeps all the initializer logic within the initializer, and lets you remove the `replace` logic from `index.js`.

kennethdavidbuck:
@samselikoff , thanks! The reason why I am replacing the entire initializer is because it imports code that doesn't exist in production builds (since everything gets pruned) which, if I remember correctly, leads to missing import errors etc. The idea was to strip all references. I am completely open to trying any alternatives/suggestions you might have for that. Thanks again!

samselikoff:
Ah, right you are. Let's go with this approach for now. Thanks!

kennethdavidbuck:
My pleasure, and thank you!

HodofHod:
This change causes a massive performance hit (~15s) to my warm builds even when `excludeFilesFromBuild` is true. The logs indicate that returning a tree from `treeFor` causes `lodash` (and other dependencies) to be funneled into the build, even though they don't end up in the vendor.js. Besides for that, SimpleReplace (from `broccoli-string-replace`) takes about ~7s (running on multiple trees).

This is on ember-cli@3.3.0, development build.

```
Build successful (44671ms)

Slowest Nodes (totalTime => 5% )              | Total (avg)
----------------------------------------------+---------------------
Babel: lodash (1)                             | 9446ms
broccoli-persistent-filter:SimpleReplace (5)  | 6764ms (1352 ms)
Babel: my-app (5)                             | 3899ms (779 ms)
```


Using 
```js
addons: {
  blacklist: [
    EmberApp.env() === 'test' ? "" : "ember-cli-mirage"
  ],
}
```
seems to be the way to do it.

samselikoff:
Yeesh. Opened an issue to track.

samselikoff:
@HodofHod https://github.com/samselikoff/ember-cli-mirage/pull/1410 should fix it! Let me know if you run into anything else related to this 👍

HodofHod:
@samselikoff That's amazing, thank you!

PR #1237, opened 2017-12-07
fix auto-discovery of nested ember-data models

ro0gr:
Currently while discovering `ember-data` models mirage registers found model under its leaf name. For example if you have a `pet/dog` ember-data model it will be accessible as `schema['dog']`. This causes some exceptions while I try to use `ember-cli-mirage`.

After this change you can access nested model like `schema['pet/dog']`.


ro0gr:
@offirgolan I would appreciate your 👀 here when you have time

ro0gr:
@samselikoff what can I do to get it merged?

samselikoff:
Thank you!

ro0gr:
🎉

Issue #1204, opened 2017-10-30
0.4.0 does not include relationships


kielni:
Upgrading from 0.3.3 to 0.4.0 broke my acceptance tests that use objects with `belongsTo` relationships. My tests fail when trying to navigate through the `belongsTo` because the relationships were not included.  For example, with a `bill` model that has `meter: belongsTo('meter')`, `bill.meter.id` is undefined, because the `meter` relationship was not included in the response.

I tried Mirage Boilerplate Twiddle but got an alert that "The gist is missing or secret"

My `bill` model looks like this:
```
  meter: DS.belongsTo('meter'),
  initial: DS.attr('date),
```

My real API and Mirage 0.3.3 ([_getResourceObjectForModel](https://github.com/samselikoff/ember-cli-mirage/blob/ee0efc89615c0ae0553a5407286fcde4f17d7382/addon/serializers/json-api-serializer.js#L167)) return data like this:

```
{
    "type": "bill-summary",
    "attributes": {
        "initial": "2014-07-02"
    },
    "id": "4170595602188",
    "relationships": {
        "meter": {
            "data": {
                "type": "meter",
                "id": "1971049867684"
            }
        }
    }
}
```

With 0.4.0, there are no relationships included because [_maybeAddRelationshipsToResourceObjectForModel](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializers/json-api-serializer.js#L181) includes them only if `hasLinksForRelationship`, `alwaysIncludeLinkageData`, or `this._relationshipIsIncluded`  Data returned by Mirage 0.4.0:

```
{
    "type":"bill-summary",
    "id":"1",
    "attributes": {
        "initial": "2014-07-02"
    }
}
```

My real API does not add links or include relationships for this model, and `alwaysIncludeLinkageData`is [defaults to false](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializers/json-api-serializer.js#L221)

I think this changed here: https://github.com/samselikoff/ember-cli-mirage/commit/ab87ac40747d1f46e8b5d6fa22bcfa133271a07f  I expect can work around this by setting `alwaysIncludeLinkageData` on every model with `belongsTo` relationships, but I'd rather that my Mirage API responses match my real API responses as much as possible.

riklaunim:
I have it too. When I create a "foo" that has FK to "bar" then with 0.4.0 foo.bar will be undefined and tests will fail.
```
[...]

PR #1252, opened 2018-01-18
Include information about linkage change in changelog

TayHobbs:
I spent a large amount of time to figure out why my update from 0.3.x to 0.4.x caused breaking changes in my application's tests. The changelog is currently misleading, implying that no change is necessary for the update. This PR adresses #1204 and uses wording from that issue (credit to @seansellek for the wording).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ Changes:
 
 Update notes:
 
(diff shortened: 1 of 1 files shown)


TayHobbs:
@samselikoff any thoughts on this? I also do not know why the build would be failing.

samselikoff:
I added the info + a blog post, thanks for contributing!

PR #1259, opened 2018-02-06
Remove sinon as a dependency since its no longer used

sivakumar-kailasam:


geekygrappler:
Hey, I think you need to remove `bower install` from here and this should be good to go.
https://github.com/samselikoff/ember-cli-mirage/blob/master/.travis.yml#L50

sivakumar-kailasam:
@geekygrappler could you restart the failed build? seems like an intermediate issue

sivakumar-kailasam:
@geekygrappler Force pushed the commit again to make it pass. Good to go now.

PR #1261, opened 2018-02-15
Reword manual Mirage, with short notes on endpoint config

xcambar:
Following my question on Slack, this PR comes with small changes in the documentation on how **and when** to start Mirage manually.

The added documentation could have been much extended, but I'm waiting to hear your feedback on the necessity to expand further.


samselikoff:
Thank you, and sorry for the delay!

We should rework a lot of the testing setup guides now with the new-style tests.

PR #1263, opened 2018-02-16
Fix RFC232/268 tests

### LINKED ISSUE #1257: Does not play nicely with rfc232/rfc268 tests
The new test framework ([rfc232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)/[rfc268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md)) runs initializers differently than the older `moduleFor*` tests, and is problematic for how `ember-cli-mirage` sets itself up. Specifically, it can cause the multiple pretender instances warning, and can leak state between tests.

## Old tests
In the old tests, each individual test would have its own mirage server instance.

### integration & unit
In old-style integration/unit tests, the test framework would not run any initializers, so the test itself would have to invoke the initializer and shut down the server [manually](http://www.ember-cli-mirage.com/docs/v0.3.x/manually-starting-mirage/).

### acceptance
In old-style acceptance tests, initializers would run before each test, so `ember-cli-mirage`'s initializer would run before each test, and its blueprint would cause the server to be shut down [after each test](https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/ember-cli-mirage/index.js#L53). 

## New tests
There are no longer any relevant-to-this-issue distinctions between the various types of tests -- they all behave the same with respect to initializers. That behavior is that, while instance initializers run before each test, initializers just run before the first test in the entire test run. `destroy-app.js` is no longer used, so `ember-cli-mirage`'s current blueprint has no effect on these tests. This causes a couple of problems:

1. If you have a mix of old and new-style tests, you'll probably get a whole bunch of multiple pretender instance warnings since the first new-style test that runs will create a mirage server instance and never shut it down, so every old-style test that sets its own up (either via the initializer called from a `moduleForAcceptance` test or manually in a unit/integration test) will create a "second" pretender instance.
2. If all of your tests are new-style tests, changes to the mirage database and routes dynamically created in one test will persist into other tests because it's one long-lived instance that is never shut down.

## Possible solutions
1. Change the server from being a standalone object to a singleton (`service:mirage`?) registered in, and created out of, the container. This will ensure that it is shut down when the application is shut down. Then change the initializer that creates it to be an instance initializer. That way it will be created before each test, and shut down after each test like it used to be. This would have the side benefit of allowing users like myself that aren't fond of globals to access the server by just looking it up in the container (but it could still set itself as a global to preserve compatibility).
2. Implement some mechanism for resetting the mirage server to its initial state and implement an instance initializer that does that.
3. Explore other ways of running code after each test to shut down/reset the mirage server. This could be blueprint code that detects QUnit vs Mocha and adds global after-test behavior to them, or some other mechanism that I'm not aware of, or that would have to be implemented in `@ember/test-helpers`. This also might involve changing the existing initializer to an instance initializer.

My strong preference is (1), but I don't know the history of how the mirage setup code came to be how it is today, so maybe there are reasons to not make it a container object?

bendemboski:
## PR summary

The end result of this PR is:

* RFC232/RFC268 tests are now supported properly
* The `server` global is no longer necessary in RFC232/RFC268 tests, so at some point it can be deprecated

This also involved updating to `ember-cli@2.18.2` (and some associated linting fixes), which is in its own commit.

This is all backwards-compatible.

Fixes #1257 

## How it works in different types of tests

* **Old-style non-acceptance tests**: mirage will not automatically start, and can still be manually started using the `startMirage()` import from `<app>/initializers/ember-cli-mirage`.
* **OId-style acceptance tests**: mirage will still be started before each test by `<app>/initializers/ember-cli-mirage` and shut down by the code that `ember-cli-mirage`'s blueprint adds to `destroyApp()`.
* **RFC232 and RFC268 tests**: `<app>/initializers/ember-cli-mirage` will detect that it's an RFC232/RFC268 test and not start the mirage server automatically in the way that is problematic for the new tests, and since the new tests don't use `destroyApp()`, it's not at play either. So there are two options:
  - In each test that needs it, the user can call `setupMirageTest()`, which will register `beforeEach`/`afterEach` hooks to set up and shut down the mirage server and make it accessible via `this.server` (and also the `server` global).
  - The user can set `autostart: true` in their `ember-cli-mirage` config object in `config/environment` which will cause an instance initializer that runs before each test (**every** RFC232/RFC268 test, not just "acceptance" (application) tests) to start mirage and set it as `this.server` on the test context (and also the `server` global), and also register and create a singleton that, when destroyed during app shutdown, will shut down the mirage server.

## Upgrade path

This is **not** a breaking change -- existing tests will run just as they are once this PR is merged and released. Furthermore, new and old tests can live side by side and work properly with mirage if the right setup is done, so users are not forced to migrate their whole suite all at once. As they migrate tests, they have two options:

* As non-acceptance tests are migrated, delete the manual starting/stopping of mirage in `beforeEach`/`afterEach` and replace it with a call to `setupMirageTest()` and continue using `this.server`. As acceptance tests are migrated, add a call to `setupMirageTest()` and optionally switch from using the `server` global to using `this.server`.
* Set `autostart: true` in `config/environment`, and then as non-acceptance tests are migrated, just delete the manual starting/stopping of the mirage server and continue using `this.server`. As acceptance tests are migrated, no changes are necessary, but users can optionally switch from using the `server` global to using `this.server`.


bendemboski:
The fastboot tests are failing and I'm a complete fastboot n00b. Halp!

bendemboski:
Hurray! All tests passing! Thanks for pointing me in the right debugging direction @simonihmig

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016
+Copyright (c) 2018
```
cibernox: I believe that copyright years are not supposed to be updated, and if they do it should be a year range.

addon-test-support/setup-mirage-test.js
```diff
@@ -0,0 +1,24 @@
+import startMirage from 'ember-cli-mirage/start-mirage';
+
+//
+// Used to set up mirage for a test. Must be called after one of the
+// `ember-qunit` `setup*Test()` methods. It starts the server and sets
+// `this.server` to point to it, and shuts the server down when the test
+// finishes.
+//
+export default function setupMirageTest(hooks) {
```
dfreeman: Do folks have any strong opinions about `setupMirageTest(hooks)` vs just `setupMirage(hooks)`?

Based on the way e.g. `setupRenderingTest(hooks)` implicitly sets up the test context so you don't _also_ need to call `setupTest(hooks)`, I think I would expect a function called `setupMirageTest` to do the same.

For me, `setupMirage` on its own doesn't carry that same implication, so might be clearer.

addon/instance-initializers/ember-cli-mirage-autostart.js
```diff
@@ -0,0 +1,46 @@
...
+      shutdown.set('testContext', testContext);
+    }
+  }
+}
+
+export default {
+  initialize
+};
+
+const MirageShutdown = EmberObject.extend({
```
rwjblue: I totally realize that this is fine, but it hurts my head to essentially rely on hoisting here. Can we move this above?

addon/start-mirage.js
```diff
@@ -0,0 +1,52 @@
+import Ember from 'ember';
+import readModules from 'ember-cli-mirage/utils/read-modules';
+import Server from 'ember-cli-mirage/server';
+import _assign from 'lodash/assign';
+
+const {
+  getWithDefault
+} = Ember;
```
rwjblue: Can we use `import { getWithDefault } from '@ember/object';`?

app/initializers/ember-cli-mirage.js
```diff
@@ -1,37 +1,45 @@
...
 import baseConfig, { testConfig } from '../mirage/config';
-import Server from 'ember-cli-mirage/server';
-import _assign from 'lodash/assign';
-
-const {
-  getWithDefault
-} = Ember;
-
+import getRfc232TestContext from 'ember-cli-mirage/get-rfc232-test-context';
+import newStartMirage from 'ember-cli-mirage/start-mirage';
```
rwjblue: I'd vaguely prefer a different variable name here (I'm always leery of using "new" or "legacy" in variable names because they loose meaning over time), though I don't have a better suggestion...

addon-test-support/setup-mirage-test.js
cibernox: If I had to vote, I'd also prefer `setupMirageTest`

bendemboski: You get a vote! And you get a vote! Everybody gets a vote!

bartocc: 👍  for setupMirageTest if I get a vote as well ;-)

LICENSE.md
bendemboski: Oh yeah, good catch. An artifact of `ember-cli-update`.

addon-test-support/setup-mirage-test.js
```diff
@@ -0,0 +1,24 @@
...
+        + ' setupRenderingTest() or setupApplicationTest() methods before'
+        + ' calling setupMirageTest()');
+    }
+
+    this.server = startMirage(this.owner);
+  });
+
+  hooks.afterEach(function() {
+    this.server.shutdown();
+    delete this.server;
```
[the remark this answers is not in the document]
bendemboski: Yeah, I mainly put that in to make testing of the addon easier (e.g. [here](https://github.com/samselikoff/ember-cli-mirage/pull/1263/files#diff-54a9281995063e12fb3000bd3db0675dR62)).

addon/start-mirage.js
bendemboski: Sure...I guess I left it this way because it's what the modules API codemod produces. Maybe a bug with the codemod?

app/initializers/ember-cli-mirage.js
bendemboski: Yeah, I totally agree -- over in my `ember-cli-page-object` PR, I wrote (mostly in comments) "new test API" and then went back and changed it to "rfc232/rfc268 test API" like a thousand times 😄 

Maybe `realStartMirage` or `startMirageImpl`?

addon-test-support/setup-mirage-test.js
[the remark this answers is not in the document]
bendemboski: You mean just `setupMirageTest(hooks = self)`?

bendemboski: After thinking about this some more, I think I also prefer `setupMirage` over `setupMirageTest` since it distinguishes it from the methods that set up the test (context) itself, and also the word `test` is totally redundant since this method is only available in tests (imported from `addon-test-support/`), so I don't see that it adds anything.

Does anybody have an argument for `setupMirageTest` other than vague personal preference?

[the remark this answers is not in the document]
rwjblue: Yes, precisely

addon/start-mirage.js
rwjblue: Ya, possibly. Might need to make sure it has the newest mapping’s....

samselikoff:
Thank you so much for this, and sorry about the delay!

buschtoens:
For the Googlers: This is how `setupMirageTest` is imported.

```js
import setupMirageTest from 'ember-cli-mirage/test-support/setup-mirage';
```

balinterdi:
Thank you, Jan, that's quite useful.

I have a question related to your testing code: what is the use of the following custom hook you set up?

```js
hooks.beforeEach(function() {
    this.actions = {};
    this.send = (actionName, ...args) => this.actions[actionName].apply(this, args);
  });
```

buschtoens:
The test above is taken from some legacy tests in [ember-light-table](https://github.com/offirgolan/ember-light-table/blob/e6f027eefbc5451d8ad330eeba28c17f539a387f/tests/integration/components/light-table-test.js#L18-L21) that we've converted to the new testing API.

Thank you for pointing out that oddity. This "hack" is (or was?) required to make the `(action "someAction")` helper work with named actions, as used in the ["extra data and tableActions" test](https://github.com/offirgolan/ember-light-table/blob/e6f027eefbc5451d8ad330eeba28c17f539a387f/tests/integration/components/light-table-test.js#L222-L262):

```js
    this.actions.someAction = () => {
      assert.ok(true, 'table action is passed');
    };

    await render(hbs `
      {{#light-table table
        extra=(hash someData="someValue")
        tableActions=(hash
          someAction=(action "someAction")
        )
        as |t|
      }}
        {{t.head}}
        {{t.body}}
      {{/light-table}}
    `);
```

This is unnecessarily complicated and should be replaced by what is done above in the ["scrolled to bottom" test](https://github.com/offirgolan/ember-light-table/blob/e6f027eefbc5451d8ad330eeba28c17f539a387f/tests/integration/components/light-table-test.js#L30-L55):

```js
    this.set('onScrolledToBottom', () => {
      assert.ok(true);
    });

    await render(hbs `
      {{#light-table table height='40vh' as |t|}}
        {{t.head fixed=true}}
        {{t.body onScrolledToBottom=(action onScrolledToBottom)}}
      {{/light-table}}
    `);
```

simonihmig:
FWIW, `ember-qunit-codemod` does exactly this transformation, when you use `(action "someAction")`: https://github.com/rwjblue/ember-qunit-codemod/blob/afceb5ba1b8cf603463621f65220f40b22027181/__testfixtures__/ember-qunit-codemod/on.output.js#L11. So probably not that "hacky" actually :)

But as @buschtoens pointed out, this can be simplified with passing a function directly...

balinterdi:
@simonihmig So the hacky solution is applied for old-style (non-closure) actions?

simonihmig:
> So the hacky solution is applied for old-style (non-closure) actions?

Yeah, I think so. 

Tbh, I sometime get confused about the nomenclature... `action=(action "someAction")` is still a closure action, isn't it? As the action helper returns a plain function (rather than those really old string actions like `action="someAction"`). 

Anyway, as `(action "someAction")` will try to lookup "someAction" on the actions hash, the codemod will add this `this.actions = {};` to make it work, I believe...

balinterdi:
> Tbh, I sometime get confused about the nomenclature... action=(action "someAction") is still a closure action, isn't it? As the action helper returns a plain function (rather than those really old string actions like action="someAction").

You're not alone :), and yes, it's a closure action. If I'm unsure I look up [this excellent post](https://embermap.com/notes/26-a-note-on-actions).

PR #1264, opened 2018-02-21
Add options verb mocking

asantos00:
Hi guys!

Good job with this plugin, we at Uniplaces use it in all our projects and it makes the development super easy, it is completely essential now!

# Context
This adds the possibility to mock `option` requests. I followed what you guys said here (https://github.com/samselikoff/ember-cli-mirage/issues/804). Hope it is alright!

- I've renamed the internal `options` (configuration options) as it is limiting us (we couldn't define the `options` method). I would love to hear from you guys if you think this can break APIs. 

- From my understanding, we shouldn't be changing options from outside of the server and I obviously renamed the inside calls to `this.options` solving the above problem. 

If it does break, it would be cool to discuss a better way of doing this :) (and create tests to confirm people don't rename `this.options` without tests breaking.

Thanks!

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -149,7 +149,7 @@ export default class Server {
       'You cannot modify Mirage\'s environment once the server is created');
     this.environment = config.environment || 'development';
 
-    this._options = config;
+    this._config = config;
 
     this.timing = this.timing || config.timing || 400;
     this.namespace = this.namespace || config.namespace || '';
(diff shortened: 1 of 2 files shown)


samselikoff:
looks good! How about instead of `_options` we just call it `this.config`?

samselikoff:
Whoops the method is `config`. How about `_config` for now.

asantos00:
Yeah, i've also thought about that! Going to change it, thanks for the fast response ;)

asantos00:
Done @samselikoff 🙏

samselikoff:
sweet!

PR #1269, opened 2018-02-23
Allow nested mirage modules

ro0gr:
When I've tried to put a factory to the nested directory inside of `mirage/factories/` directory I've got an error. Is there any reasons for this? With this change seems like everything works.

### DIFF
--- a/addon/utils/read-modules.js
+++ b/addon/utils/read-modules.js
@@ -4,7 +4,7 @@
 
 import { assert } from '@ember/debug';
 import _camelCase from 'lodash/camelCase';
-import { pluralize } from 'ember-cli-mirage/utils/inflector';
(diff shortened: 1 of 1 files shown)


addon/utils/read-modules.js
```diff
@@ -51,7 +50,7 @@ export default function(prefix) {
 
     let data = module.default;
 
-    modulesMap[moduleType][_camelCase(moduleKey)] = data;
+    modulesMap[moduleType][camelize(moduleKey)] = data;
```
ro0gr: this is needed to match the name while we lookup factories https://github.com/samselikoff/ember-cli-mirage/blob/f0e5c30add1117b673014d4918abfa087b273019/addon/server.js#L338-L339

ro0gr:
superseded by #1287

PR #1272, opened 2018-02-27
support include as array

sbatson5:
# What's in this PR?

Ran into this issue earlier with params such as `include[]=users`.  The current JSONAPISerializer assumes that the `include` query param will be a string, but it could be passed as an array.

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -62,7 +62,8 @@ const JSONAPISerializer = Serializer.extend({
     let relationshipPaths;
 
     if (this.hasQueryParamIncludes()) {
-      relationshipPaths = this.request.queryParams.include.split(',');
+      let { include } = this.request.queryParams;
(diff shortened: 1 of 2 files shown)


samselikoff:
Thanks for the PR! 

JSONAPISerializer is meant to work with JSON:API spec-compliant backends - I've never come across the form of `includes` being passed as an array. Would you mind sharing where you're seeing this?

sbatson5:
@samselikoff thank you for the review. I wasn't aware that json-api spec requires `include` to be a comma separate list http://jsonapi.org/format/#fetching-includes:
>The value of the include parameter MUST be a comma-separated (U+002C COMMA, “,”) list of relationship paths. A relationship path is a dot-separated (U+002E FULL-STOP, “.”) list of relationship names.

The Rails API we work on has a `Devise` controller that expects `include` as an array and is how we pass our data from our front-end.   I suppose this is more a question for the spec than mirage. Happy to close the issue if no one else is running into this.

Issue #1274, opened 2018-03-02
Serializer produces wrong association key format for polymorphic has-many relations


kaeufl:
There are two issues related to how polymorphic has-many relations are handled by the base Serializer:

Given the `user` model
```javascript
Model.extend({
    things: hasMany({ polymorphic: true })
})
```
1. If `serializeIds` is set to `'always'`,
the expected json (in accordance with what ember-data expects, as far as I know) would be 
```json
{
  "user": {
    "id": "1",
    "thingIds": [{
        "id": "1",
        "type": "picture"
      }]
  }
}
```

while the actual result is
```json
{
  "user": {
    "id": "1",
    "thingIds": ["1"]
  }
}
```
That is, the type information is missing from the polymorphic id.

2. If on the other hand `serializeIds` is set to `'included'`, and the 'things' relation is requested to be included in the payload, then to my understanding the expected json should be
```json
{
  "pictures": [
    {
      "id": "1",
      "title": "Lorem ipsum"
    }
  ],
  "user": {
    "id": "1",
    "name": "Ned",
    "thingIds":  [
          { "id": "1", "type": "picture" }
     ]
  }
}
```

while the actual result is
```json
{
  "pictures": [
    {
      "id": "1",
      "title": "Lorem ipsum"
    }
  ],
  "user": {
    "id": "1",
    "things": [
          { "id": "1", "type": "picture" }
     ]
  }
}
```

That is, the key is called "things" rather than "thingIds" in the serialization of the user model, although this time the type info is correctly reflected.

These issues can easily be worked around by overriding the serialize method in a custom serializer, but nevertheless, I feel that the current default behaviour is somewhat odd.

geekygrappler:
Where are you getting the expected responses from? The PR looks OK, but I couldn't find what ED would expect here?

geekygrappler:
Could you show me the setup for how you'd use `serializeIds` in an app because I can't get it to work

kaeufl:
1. Regarding the format of the polymorphic id:
[...]

PR #1275, opened 2018-03-02
Fix format of association ids for polymorphic has-many relations in base serializer (#1274)

kaeufl:
Failing test for #1274 and attempt at fixing.

### DIFF
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -271,9 +271,23 @@ class Serializer {
   _maybeAddAssociationIds(model, attrs) {
     let newHash = _assign({}, attrs);
 
-    if (this.serializeIds === 'always') {
-      model.associationKeys.forEach((key) => {
-        let association = model[key];
+    let associationKeys = this._associationKeysToInclude(model);
+
+    associationKeys.forEach((key) => {
+      let association = model[key];
+
+      if (model.associationFor(key).isPolymorphic) {
+        if (this.isCollection(association)) {
+          let formattedKey = this.keyForRelationshipIds(key);
+          newHash[formattedKey] = model[`${singularize(key)}Ids`];
+        } else {
+          let formattedTypeKey = this.keyForPolymorphicForeignKeyType(key);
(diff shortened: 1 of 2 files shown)


addon/serializer.js
```diff
@@ -271,52 +271,43 @@ class Serializer {
...
+      }
+    });
 
-        if (model.associationFor(key).isPolymorphic) {
-          if (association instanceof PolymorphicCollection) {
-            let formattedKey = this.keyForRelationship(key);
-
-            newHash[formattedKey] = model[`${singularize(key)}Ids`];
-          } else if (association instanceof Collection) {
-            let formattedKey = this.keyForRelationshipIds(key);
```
kaeufl: Could this code be reached? All tests stayed green, when I removed this case.

geekygrappler: I'm really uncomfortable deleting this much code, even if the tests go green.

kaeufl: Keeping aside the problem that the branch for polymorphic relations seemed to be missing completely from the `this.serializeIds === 'always'`case, I also couldn't really see why the cases `this.serializeIds === 'always'` and `this.serializeIds === 'included'` should be handled differently at all, when it comes to how the ids are being serialized.

As far as the deleted code is concerned, I didn't really remove that much logic. Lines 277-283 of the original code were duplicates of lines 306-313. The actual bits of logic that were altered are:
* the branch for polymorphic relations is now also present in the `this.serializeIds === 'always' case
* in the _polymorphic_ case `PolymorphicCollection` and `Collection` are treated the same (this is admittedly the bit I'm most uncertain of, too)
 * instead of calling the `keyForRelationship` hook (line 291 of old code), the `keyForRelationshipIds` hook is now called. For the reasoning behind this, please see the second point here https://github.com/samselikoff/ember-cli-mirage/issues/1274#issuecomment-371300993

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1348 closes

PR #1281, opened 2018-03-14
POC: Fix teardown when pending requests

wagenet:
Since QUnit callbacks are async, this leaves a window where an AJAX request could happen _after_ mirage is torn down but before the application is. This can cause errors. To fix, we make sure that things have settled before we tear down the server.

### DIFF
--- a/addon-test-support/setup-mirage.js
+++ b/addon-test-support/setup-mirage.js
@@ -1,4 +1,5 @@
 import startMirage from 'ember-cli-mirage/start-mirage';
+import { settled } from '@ember/test-helpers;
 
 //
 // Used to set up mirage for a test. Must be called after one of the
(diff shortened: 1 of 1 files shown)


addon-test-support/setup-mirage.js
```diff
@@ -1,4 +1,5 @@
 import startMirage from 'ember-cli-mirage/start-mirage';
+import { settled } from '@ember/test-helpers;
```
ro0gr: missing closing apostrophe

ro0gr:
I've just ran to this issue. It took me few hours to understand the reason..

wagenet:
@samselikoff does this look correct to you? If so, I can clean up the PR.

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1349 supersedes

Thanks Peter!

PR #1282, opened 2018-03-14
fix(rest-serializer): return correctly serialized objects 

geekygrappler:
Part of #1274.

When using **ember-data's** `RESTSerializer`, it expects associated objects rather than ids, i.e.:
```
{
  "user": {
    "id": "1",
    "things": [{
        "id": "1",
        "type": "picture"
      }]
  }
}
```
What will not work with `RESTSerializer` is the following:
```
{
  "user": {
    "id": "1",
    "thingIds": [{
        "id": "1",
        "type": "picture"
      }]
  }
}
```
OR
```
{
  "user": {
    "id": "1",
    "thingIds": ["1"]
  }
}
```
I couldn't really find good docs on what ember data expects when you are sending associations. But I have this [repo](https://github.com/geekygrappler/mirage-example/tree/isssue-1274-rest-serializer), when you visit http://localhost:4200/users/1 it'll throw an error. When using **Mirage's** `RestSerializer` with `serializeIds: 'always'` the response is:
```
{
  "user": {
    "id": "1",
    "things": ["1"]
  }
}
```

And if I make the changes in this PR, then ember data doesn't throw an error and loads the associated models.

I think the main problem here, is that **ember-data's** `RESTSerializer` doesn't seem to respond to serialized Ids for associations in responses i.e. `thingIds: [ids]`, or it wants serialized objects for with ids for associations in repsonses i.e. `things: [objects]`.

So while this fix works, I'm not sure `serializeIds: 'always'` is the correct option for telling **mirage's** `RestSerializer` to include the associated objects in the response.

### DIFF
--- a/addon/serializers/rest-serializer.js
+++ b/addon/serializers/rest-serializer.js
@@ -1,5 +1,8 @@
 import ActiveModelSerializer from './active-model-serializer';
 import { camelize, singularize, pluralize } from '../utils/inflector';
+import _assign from 'lodash/assign';
+import Collection from '../orm/collection';
+import PolymorphicCollection from '../orm/polymorphic-collection';
 
 export default ActiveModelSerializer.extend({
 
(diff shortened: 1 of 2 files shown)


addon/serializers/rest-serializer.js
```diff
@@ -29,6 +32,64 @@ export default ActiveModelSerializer.extend({
...
+  _maybeAddAssociationIds(model, attrs) {
+    let newHash = _assign({}, attrs);
+
+    if (this.serializeIds === 'always') {
+      model.associationKeys.forEach((key) => {
+        let association = model[key];
+        if (this.isCollection(association)) {
+          let formattedKey = this.keyForRelationshipIds(key);
+          newHash[formattedKey] = model[key].models.map((m) => {
+            return { id: m.id, type: m.modelName };
```
geekygrappler: This is copy and paste of the function here: https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializer.js#L279 with only this line changed. I didn't really know the best way to just change this line.

tests/integration/serializers/rest-serializer/serialize-ids-test.js
```diff
@@ -0,0 +1,54 @@
...
+      id: 1,
+      name: 'Link'
+    });
+    wordSmith.createBlogPost();
+    wordSmith.createBlogPost();
+    wordSmith.createSpecialPost();
+
+    let result = registry.serialize(wordSmith);
+
+    assert.deepEqual(result, {
```
geekygrappler: The test is kind of pointless because I'm just saying this is what Ember Data expects, but no way to confirm this...

geekygrappler:
Hey @samselikoff, is there anyone you would recommend to check over what I've done here? It's not too complicated I just want to make sure I'm not making a mistake because I'm not too familiar with Ember Data & Rest serializer.

kaeufl:
Hey @geekygrappler, while your PR would solve the first problem mentioned in #1274 for our app (since we are indeed using the RESTSerializer), I still think it should rather be fixed in the base class itself. In my opinion an implementation of a serializer that omits the type information when serializing the ids of a _polymorphic_ relationship just doesn't make any sense. What's a client supposed to do with a response like
```
{
  "user": {
    "id": "1",
    "things": ["1"]
  }
}
```
when thing could be either a `picture` or a `book` (for example)? How should he know what type of related resource to load?

samselikoff:
@geekygrappler sorry about the delay. Sifting through a backlog of issues but will try to make it back to this soon

samselikoff:
@geekygrappler Was this PR specifically about Polymorphic relationships, or just any hasMany? The test case seems to set up a normal hasMany.

samselikoff:
@geekygrappler Can you take a look at the test added in https://github.com/samselikoff/ember-cli-mirage/pull/1350?

I believe some other changes I recently made fixed this behavior.

samselikoff:
Went ahead and merged https://github.com/samselikoff/ember-cli-mirage/pull/1350. Closing for now, thanks for your help & let me know if this is still an issue!

PR #1291, opened 2018-04-01
Revert initializer name

bendemboski:
I didn't mean to change this, and it breaks apps that have initializers that reference this one in their `before` or `after` fields.


samselikoff:
👍  thanks ben!

PR #1296, opened 2018-04-10
[DOC] quickstart s/schema/schema.db/g

givanse:


samselikoff:
This is actually supposed to be demonstrating Mirage's ORM (accessed via `schema.authors.[method]` which is the primary way to interact with Mirage's database. Maybe the lead-up paragraph is confusing?

PR #1297, opened 2018-04-10
 fix: _getAttrsForRequest now takes associations argument, to allow it to resolve polymorphic relationship patch/post

n8n8baby:


samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1337 supercedes

PR #1305, opened 2018-04-18
Allow for the same inverse association key to exists many times

jrjohnson:
While this solution works I don't really know enough about the mirage internals to know if it is the right fix. 

Hopefully a test is worth a thousands words here, I'm finding this issue difficult to explain. This should be supported:
```
user: Model.extend({
  authoredPosts: hasMany('post', { inverse: 'authors' }),
  authoredBooks: hasMany('book', { inverse: 'authors' })
}),
post: Model.extend({
  authors: hasMany('user', { inverse: 'authoredPosts' })
}),
book: Model.extend({
  authors: hasMany('user', { inverse: 'authoredBooks' })
})
```

The `user` model here inverses two other models with the same property.

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -271,7 +271,7 @@ class Model {
     let associations = this.schema.associationsFor(this.modelName);
     let matchingExplicitInverses = Object.keys(associations).filter(key => {
       let candidate = associations[key];
-      let modelMatches = association.modelName === candidate.ownerModelName;
+      let modelMatches = association.ownerModelName === candidate.modelName;
       let inverseKeyMatches = association.key === candidate.opts.inverse;
 
       return modelMatches && inverseKeyMatches;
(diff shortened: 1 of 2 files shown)


addon/orm/model.js
```diff
@@ -271,7 +271,7 @@ class Model {
     let associations = this.schema.associationsFor(this.modelName);
     let matchingExplicitInverses = Object.keys(associations).filter(key => {
       let candidate = associations[key];
-      let modelMatches = association.modelName === candidate.ownerModelName;
+      let modelMatches = association.modelName === candidate.ownerModelName && association.ownerModelName === candidate.modelName;
```
samselikoff: Miraculously, the existing code is simply wrong. Not sure how it managed to survive this long.

We're making sure there's only one association on `user` that matches `post.authors`, so we want to check all the associations on `user` (the candidates) whose _modelName_ matches the given association's _ownerModelName_ (and whose inverse key matches).

The code should be

```js
let modelMatches = association.ownerModelName === candidate.modelName;
```

jrjohnson: Cool, it felt like that might be true to me, but I was completely unable to keep all the association and model names straight. I must have stared at my debugger for 15 minutes just looking.  
Anyway, I just pushed an amended commit with only the correct check.

samselikoff: haha I know... very confusing part of the codebase. Always open to improvements if you have any ideas!

Thanks!

PR #1309, opened 2018-04-26
return pretender handler

kellyselden:
This will enable you to do:

```js
  this.post('token-auth', () => ({ 'jwt': 'xxx' }))._name = 'post:token-auth';
```

```js
export function errorOnAuth() {
  let original = server.pretender.handlers.find(h => h._name === 'post:token-auth');
  server.post('token-auth', { errors: [{ detail: 'There was an error' }] }, 500);
  return function reset() {
    server.post('token-auth', original);
  };
}
```

unless there is already a way to do this?

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -485,7 +485,7 @@ export default class Server {
     [['get'], ['post'], ['put'], ['delete', 'del'], ['patch'], ['head'], ['options']].forEach(([verb, alias]) => {
       this[verb] = (path, ...args) => {
         let [ rawHandler, customizedCode, options ] = extractRouteArguments(args);
-        this._registerRouteHandler(verb, path, rawHandler, customizedCode, options);
+        return this._registerRouteHandler(verb, path, rawHandler, customizedCode, options);
       };
 
       if (alias) {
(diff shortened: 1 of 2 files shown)


samselikoff:
no this is great! Would love to add more sugar over this pattern in the future too 👍  Something like `server.once.get`

samselikoff:
@denzo check it out:

```js
test('it can recover from an error', async function(assert) {
  // For a server error 
  let originalHandler = server.pretender.handlers.find(h => h._name === 'post:users');
  server.post('/users', { errors: [ 'Uh-oh, there was an error' ] }, 500);

  await visit('/');
  await click('[data-test-id="add-user"');
  await fillIn('[data-test-id="name"]', 'Sam');
  await click('[data-test-id="save"');

  assert.dom('[data-test-id="message"]').hasText('There was a problem');

  // Restore the handler
  server.post('/users', originalHandler);
  
  await click('[data-test-id="save"');
  
  assert.dom('[data-test-id="message"]').hasText('Success! User created!');
});
```

Leooo:
for some reason I can't access any originalHandler on my side @samselikoff : `h._name` is undefined, I see a `h.name` which always returns "". Any idea which way to go from there? USing ember-cli-mirage 4.14, pretender 2.1.1

Leooo:
Any test of this functionality available?

kellyselden:
@Leooo You have to set _name yourself, like in my first example.

Leooo:
thanks @kellyselden

PR #1311, opened 2018-04-28
Enhance log messages

gcollazo:
* Default message stays the same
* Groups logs using `console.groupCollapsed`
* Adds request to logged data
* Adds "raw" request to logged data

Demo: https://www.youtube.com/watch?v=qV_HlRnvKZk

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -38,9 +38,18 @@ function createPretender(server) {
 
     this.handledRequest = function(verb, path, request) {
       if (server.shouldLog()) {
-        console.log(`Mirage: [${request.status}] ${verb.toUpperCase()} ${request.url}`);
-        let { responseText } = request;
+        console.groupCollapsed(
+          `Mirage: [${request.status}] ${verb.toUpperCase()} ${request.url}`
+        );
+        let { requestBody, responseText } = request;
+        let loggedRequest;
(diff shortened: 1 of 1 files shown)


jherdman:
This is awesome.

addon/server.js
```diff
@@ -38,17 +38,30 @@ function createPretender(server) {
 
     this.handledRequest = function(verb, path, request) {
       if (server.shouldLog()) {
-        console.log(`Mirage: [${request.status}] ${verb.toUpperCase()} ${request.url}`);
-        let { responseText } = request;
-        let loggedResponse;
+        console.groupCollapsed(
```
jherdman: Can you confirm that this works on mobile Safari and mobile Chrome? Compatibility [isn't clear](https://developer.mozilla.org/en-US/docs/Web/API/Console/groupCollapsed), and given that one may need to run tests against these targets it'd be nice to not have surprise fails here.

gcollazo:
Thanks for the feedback @jherdman. Did some testing and it works on all the tested devices. Do you think we need to test on more devices?

**ios7**
ios7

**ios11**
ios11

**android4.4**
android4 4

**android7**
android7

samselikoff:
Very cool @gcollazo – thanks!!

Issue #1298, opened 2018-04-13
Factories using `association()` create extra models when setting the associations as attributes on create


happycollision:
It is my understanding (especially if your inspiration for your factories was FactoryBot/Girl) that when you pass in associations to override on model creation—e.g. `server.create('child', {mother})`—the model being created should not have its factory *also* create the association being passed in. It ought to just apply the passed association.

If you create a `Mother` and then create a `Child` where you pass the mother in as an attribute on the `create` method, a new `Mother` should not be created.

This is easier to explain with the code below:

If your factory for a `Child` model looks like this

```javascript
export default Factory.extend({
  name() {
    return faker.name.firstName();
  },
  
  mother: association(),
});
```

and you run the following code in your default scenario

```javascript
export default function(server) {
  function logMothers() {
    const numMothers = server.db.mothers.length
    console.log(`We should have 1 mother. Actual number: ${numMothers}`)
  }
  
  const mother = server.create('mother');
  logMothers();
  const child1 = server.create('child', {mother});
  logMothers();
  const child2 = server.create('child', {mother});
  logMothers();
  
  console.log('Mother\'s children (by ID): ' + mother.childIds.join(', '))
}
```

your console will produce the following:

```txt
We should have 1 mother. Actual number: 1
We should have 1 mother. Actual number: 2
We should have 1 mother. Actual number: 3
Mother's children (by ID): 1, 2
```

With complex scenarios whose factories make use of `association()`, there is a TON of extra data created.

Open the console at the following twiddle and see for yourself.
https://ember-twiddle.com/c9493551fda94449ce67bc005936bf1b?fileTreeShown=false&openFiles=mirage.scenarios.default.js%2C

happycollision:
I have just discovered that if you use the `create` method from the schema, no extra models are created. So if I changed the code above...

```javascript
  const mother = server.create('mother');
  logMothers();
  const child1 = server.schema.children.create({mother});
  logMothers();
  const child2 = server.create('child', {mother});
  logMothers();
```

The console would read:
```txt
We should have 1 mother. Actual number: 1
We should have 1 mother. Actual number: 1
We should have 1 mother. Actual number: 2
```

happycollision:
Nevermind. Not a workaround. That version of `create` does not use the factory at all.

happycollision:
My new workaround is to define the following function which I can import into a factory definition file:

```javascript
export function ensureBelongsTo(currentObject, server, property, type) {
[...]

PR #1315, opened 2018-05-06
Don't create extra models in the db when passing association overrides to `create`

happycollision:
This addresses #1298. Hope this change works for you.


samselikoff:
Thank you!

PR #1323, opened 2018-06-05
Relationship path normalization

### LINKED ISSUE #1322: JSON API Serializer assumes `include`d relationships are always dasherized
We have an API that uses underscores to separate multi-word model names (e.g. `blog_post` instead of `blog-post`) in its responses.

Accordingly, it also wants the `?include=` query param to use the same format for relationship names (e.g. `GET /comments/1?include=blog_post`).

I therefore am wanting to configure Mirage to respond to a request like so:
```js
// GET /comments/1?include=blog_post

{
  data: {
    attributes: {},
    id: "1",
    type: "comments"
    relationships: {
      blog_post: { data: { id: "1", type: "blog-posts" } }
    }
  },
  included: [{
    id: "1",
    attributes: {},
    type: "blog-posts"
  }]
}
```

~~(I would probably also ideally want to alter the `type` values in the payload to be singularized+underscored, but that's unrelated to this specific issue.)~~ Edit: This can be changed using `typeKeyForModel`.

If I define a `keyForRelationship` method on the serializer that returns "blog_post" for the relationship key "blogPost", then I can get partway there: The `included` section of the mirage response is correct, but the `relationships` attribute is not present in the `data` section.

This appears to be due to how Mirage creates its internal graph (in `_addPrimaryModelToRequestedIncludesGraph`) and how it later reads that graph (in `_relationshipIsIncludedForModel`).
When creating the graph from the `queryParamIncludes`, it uses the `relationshipKey` unmodified from the query param:

https://github.com/samselikoff/ember-cli-mirage/blob/3b4509c5fc93942e227f5d0f52d7cf549f133a8d/addon/serializers/json-api-serializer.js#L280

In my example case the value of `relationshipKey` is "blog_post".

When using the graph to populate the response `relationships` field in `_relationshipIsIncludedForModel`, it looks for the dasherized model name in the graph's `included` part:

https://github.com/samselikoff/ember-cli-mirage/blob/3b4509c5fc93942e227f5d0f52d7cf549f133a8d/addon/serializers/json-api-serializer.js#L217

In the case of my example, this check fails (it is looking for `graph.relationships['blog-post']` but the data is actually in `graph.relationships['blog_post']`), so the related blog post isn't part of `data.relationships`.

However, the `getIncludesForResourceAndPath` does end up adding the related blog post to `included` because it searches for the camelized relationship path here:

https://github.com/samselikoff/ember-cli-mirage/blob/3b4509c5fc93942e227f5d0f52d7cf549f133a8d/addon/serializers/json-api-serializer.js#L91-L97

On the last line, the values from my example would be: `let relationship = comment['blogPost]'`.

I think this would be fixed if Mirage's graph always dasherizes relationship path names before putting them into the graph. I have a PR I will submit that I think will address this.

bantic:
I believe this fixes #1322. This PR adds a commit for the failing test and a separate commit to fix it.

I added the test to the `integration/server/regressions` directory because that seemed the best place for it, but I'm not sure if it really is a regression per se. I'd be happy to move it somewhere more appropriate.

### DIFF
--- a/tests/integration/server/regressions/1322-relationship-path-normalization-test.js
+++ b/tests/integration/server/regressions/1322-relationship-path-normalization-test.js
@@ -1,0 +1,73 @@
+import { module, test } from 'qunit';
+import { Model, hasMany, belongsTo, JSONAPISerializer } from 'ember-cli-mirage';
+import Server from 'ember-cli-mirage/server';
+import promiseAjax from 'dummy/tests/helpers/promise-ajax';
+
+module('Integration | Server | Regressions | 1322 Relationship Path Normalization Test', function(hooks) {
+  hooks.beforeEach(function() {
+    this.server = new Server({
+      environment: 'test',
+      models: {
+        happyUser: Model.extend({
+          happyLicenses: hasMany()
+        }),
+        happyLicense: Model.extend({
+          happyUser: belongsTo()
+        })
+      },
+      serializers: {
+        application: JSONAPISerializer.extend({
+          keyForRelationship(relationshipName) {
+            if (relationshipName === 'happyUser') {
+              return 'happy_user';
+            } else {
+              return JSONAPISerializer.prototype.keyForRelationship.apply(this, arguments);
+            }
+          }
+        })
+      },
+      baseConfig() {
+        this.resource('happy-licenses');
+      }
+    });
+  });
+
+  hooks.afterEach(function() {
+    this.server.shutdown();
+  });
+
+  test('it works', async function(assert) {
+    let user1 = this.server.create('happy-user');
+    this.server.create('happy-license', { happyUser: user1 });
+
+    assert.expect(2);
+
+    let response = await promiseAjax({
+      method: 'GET',
+      url: '/happy-licenses/1?include=happy_user'
+    });
+    let json = response.data;
+
+    assert.deepEqual(json.data, {
+      attributes: {},
+      id: "1",
+      relationships: {
+        "happy_user": {
+          data: {
+            id: user1.id,
+            type: "happy-users"
+          }
+        }
+      },
+      type: "happy-licenses"
+    });
+    assert.deepEqual(json.included, [
+      {
+        id: user1.id,
+        type: 'happy-users',
+        attributes: {}
+      }
+    ]);
+  });
+
+});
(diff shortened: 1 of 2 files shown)


tests/integration/server/regressions/1322-relationship-path-normalization-test.js
```diff
@@ -0,0 +1,73 @@
...
+
+  hooks.afterEach(function() {
+    this.server.shutdown();
+  });
+
+  test('it works', async function(assert) {
+    let user1 = this.server.create('happy-user');
+    this.server.create('happy-license', { happyUser: user1 });
+
+    assert.expect(2);
```
ro0gr: I believe it doesn't add a value to the test since the only way the assertions wouldn't be executed is a hanging promiseAjax which would lead to the test timeout anyways.

samselikoff:
Thank you!

agrobbin:
@samselikoff @ro0gr I think there might be another place where `dasherize` needs to be called:

https://github.com/samselikoff/ember-cli-mirage/blob/cb8aa34bfd74cfd606a80411d24471bf4ad00feb/addon/serializers/json-api-serializer.js#L317-L337

Specifically, line 332 needs to have its `relationshipName` wrapped in a `dasherize` call.

samselikoff:
@agrobbin Checked the test suite and it seemed the name was dasherized by the time it got to this function call. Was this causing any bugs for you?

I think the real solution here is to button up our naming conventions, perhaps with typescript.

jembezmamy:
@samselikoff Actually @agrobbin is right, this code is causing bugs for me when I do a nested camel-cased include:

- `GET /users/1?include=profilePhotos` works.
- `GET /comments/1?include=user.profilePhotos` doesn't work, user and photos are included as should be, but user is missing relationships definition.

When I paused script execution at the line 332 and dasherized `relationshipName`, it worked. I can see that `relationshipNames` was passed here via `_addModelToRequestedIncludesGraph` and `_addResourceToRequestedIncludesGraph` from the function below, where it is simply created by splitting query param without any transformation (line 268):

https://github.com/samselikoff/ember-cli-mirage/blob/387b7bac6d735f386c9fce63cadba06b59954f79/addon/serializers/json-api-serializer.js#L259-L292

@samselikoff I can write a failing test for this case if you'd find it useful.

samselikoff:
@jembezmamy would be very helpful!

I thought the convention was `user.profile-photos` but we should probably make it work either way

jembezmamy:
@samselikoff Yeah, that's true about the convention... but my API coworker wants everything to be camel cased (same story as in #1322).

PR #1326, opened 2018-06-07
Add syntheticAddrs namespace to Serializer

### LINKED ISSUE #1325: synthetic/derived attributes
This was discussed [in slack](https://embercommunity.slack.com/archives/C0TG21RPW/p1528209958000657), and it was suggested to add a tracking issue here to keep track of it.

Sometimes an API will send attributes that are composed of others. An example would be an API that doesn't conform to JSON API and sends a `parent_id` attribute instead of putting that linkage in the `relationships` hash.

Right now Mirage has the `attrs` property that can be set on serializes to whitelist the attributes that will be included in serialization, and one can also override the `serialize` method to alter the json payload, but there isn't an easy way to identify `attrs` that don't map directly to properties on the Mirage model instance.

One suggestion is to add a `syntheticAttrs` namespace to the serializer:
```javascript
// app/models/comment.js
  post: DS.belongsTo('post')
})

// mirage/serializers/comment.js
export default JSONApiSerializer.extend({
  syntheticAttrs: {
    postId(comment) {
      return comment.postId;
    }
  }
})
```

Other suggested namespaces were `derivedAttributes`, `syntheticAttributes`, `derivedAttrs`.

bantic:
This allows defining attributes that will be serialized for a model that don't map one-to-one to a model's defined attributes.

Example:
```javascript
// mirage/serializers/comment.js
export default JSONApiSerializer.extend({
  syntheticAttrs: {
    postId(model) {
      return model.postId;
    }
  }
});
```

Fixes #1325

One thought that came to mind as I put this together:
  * If both `attrs` and `syntheticAttrs` are defined for a serializer, both will be included in the serialization. This could be confusing if users understand `attrs` to mean "whitelist only these attrs and no other attributes will ever be in the serialization". I added a test to cover this scenario.

### DIFF
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -278,6 +278,12 @@ class Serializer {
     // Remove fks
     model.fks.forEach(key => delete attrs[key]);
 
+    if (this.syntheticAttrs) {
+      Object.keys(this.syntheticAttrs).forEach(syntheticAttrName => {
+        attrs[syntheticAttrName] = this.syntheticAttrs[syntheticAttrName](model);
+      });
+    }
+
     return this._formatAttributeKeys(attrs);
   }
 
--- a/tests/integration/serializers/base/synthetic-attrs-test.js
+++ b/tests/integration/serializers/base/synthetic-attrs-test.js
@@ -1,0 +1,96 @@
+import SerializerRegistry from 'ember-cli-mirage/serializer-registry';
+import Serializer from 'ember-cli-mirage/serializer';
+import schemaHelper from '../schema-helper';
+import { module, test } from 'qunit';
+
+module('Integration | Serializers | Base | Synthetic Attrs', function(hooks) {
+  hooks.beforeEach(function() {
+    this.schema = schemaHelper.setup();
+    this.registry = new SerializerRegistry(this.schema, {
+      wordSmith: Serializer.extend({
+        syntheticAttrs: {
+          foo() {
+            return 'bar';
+          },
+          foo2(model) {
+            return `${model.name}-foo2`;
+          }
+        }
+      })
+    });
+  });
+
+  hooks.afterEach(function() {
+    this.schema.db.emptyData();
+  });
+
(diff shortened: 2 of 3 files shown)


samselikoff:
Awesome PR ❤️❤️

Thinking out loud: wonder if `attrs` should actually be the complete whitelist of serialized attributes? So if `foo` is a `synthetic` attr but it's not specified in `attrs`, it should not be included?

Basically I think the intention behind `attrs` is what you said: "whitelist only those attrs and no other attributes will ever be in serializer." What do you think?

(As an aside – I'm not sure how many folks use `attrs`).

bantic:
I think that makes sense, too — so you can define synthetic attrs and they get added, along with the model's attrs, to the total possible `attrs` that would be serialized. Then at serialization time the serializer will inspect *its* `attrs` array and, if it exists, only choose those attributes (synthetic or "real") to serialize.

I like that approach! If you agree I can quickly refactor this PR to do so.

samselikoff:
Yes exactly right 👍

bantic:
I realized that this might make things a little more complicated. Right now in the "all attributes" (nothing whitelisted) case, the serializer `assign`s all of `model.attrs` into the `attrs` hash, and as a second step goes through and removes all attrs that match any of the model's `fks`. 

I think it makes the most sense for synthetic attrs to happen *after* those `fks` are removed, because it's possible one of the synthetic attrs is the same name as an `fk` (in fact, this is the use case above, with `postId`.
Is that your understanding too?

samselikoff:
@bantic Yep, that makes sense. FKs are removed since they're not attributes, they're "special" and different serializers handle them differently.

Definitely agree that synthetic attrs _are_ attrs and should be added at the end, after the model/ORM stuff is done.

samselikoff:

I definitely think something like this belongs in Mirage. I think we should borrow the Resource concept from JSONAPIResources, and use that as an abstraction layer over the db.

PR #1328, opened 2018-06-07
Throw when `include`ing a value that doesn't match an associationKey

### LINKED ISSUE #1327: JSONAPISerializer: If a queryParam include value is an attr name, unexpected error
I noticed that if the value of a query param `include` is a model's attr name, the error `Cannot read property 'replace' of undefined` is eventually thrown.

This happens because Mirage inadvertently ends up with a `relationship` variable that actually holds a property of the model rather than an actual relationship.

For example, if a particular `post` model has a `title` "Lorem Ipsum", then making the request `GET /posts/:id?include=title` will be processed by Mirage's `_addPrimaryModelToRequestedIncludesGraph` method and around these lines the variable `relationship` will equal "Lorem Ipsum":

https://github.com/samselikoff/ember-cli-mirage/blob/3b4509c5fc93942e227f5d0f52d7cf549f133a8d/addon/serializers/json-api-serializer.js#L267-L270

The `relationship` variable is passed to `_addResourceToRequestedIncludesGraph` which attempts to call `dasherize(pluralize(resource.modelName))`:
https://github.com/samselikoff/ember-cli-mirage/blob/3b4509c5fc93942e227f5d0f52d7cf549f133a8d/addon/serializers/json-api-serializer.js#L289-L292

Because `resource` is actually the string "Lorem Ipsum" and the `modelName` property is `undefined`, this results in the error mentioned above.

Would it make more sense to throw or warn in this case?

bantic:
This prevents downstream errors when a property of the model is being
passed around that Mirage thinks is an association but is in reality a
primitive (such as a string) property of the model.

Fixes #1327

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -2,6 +2,7 @@ import Serializer from '../serializer';
 import { dasherize, pluralize, camelize } from '../utils/inflector';
 import _get from 'lodash/get';
 import _ from 'lodash';
+import assert from 'ember-cli-mirage/assert';
 
 const JSONAPISerializer = Serializer.extend({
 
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you!

PR #1331, opened 2018-06-18
DEPENDENCY: updated to broccoli-string-replace

toranb:
The [broccoli-replace](https://github.com/outatime/broccoli-replace) library is both vulnerable and unmaintained so this PR migrates over to [broccoli-string-replace](https://github.com/rwjblue/broccoli-string-replace)

Here is the `npm audit` report you get today if you run that command w/ npm 6 after installing this addon. I've spent the weekend removing all references to `broccoli-replace` in other addons to help unlock a perfect 0 vulnerability audit and this will help. Thanks for reviewing!

audit

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ const mergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
 const map = require('broccoli-stew').map;
 const rm = require('broccoli-stew').rm;
-const replace = require('broccoli-replace');
+const replace = require('broccoli-string-replace');
 
 module.exports = {
   name: 'ember-cli-mirage',
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "broccoli-funnel": "^1.0.2",
     "broccoli-merge-trees": "^1.1.0",
-    "broccoli-replace": "^0.12.0",
+    "broccoli-string-replace": "^0.1.2",
     "broccoli-stew": "^1.5.0",
     "chalk": "^1.1.1",
     "ember-cli-babel": "^6.8.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -182,14 +182,6 @@ aot-test-generators@^0.1.0:
   dependencies:
     jsesc "^2.5.0"
 
-applause@1.2.2:
-  version "1.2.2"
-  resolved "https://registry.yarnpkg.com/applause/-/applause-1.2.2.tgz#a8468579e81f67397bb5634c29953bedcd0f56c0"
-  dependencies:
-    cson-parser "^1.1.0"
-    js-yaml "^3.3.0"
-    lodash "^3.10.0"
-
 aproba@^1.0.3:
   version "1.2.0"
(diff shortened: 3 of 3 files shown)


samselikoff:
Thanks!

It looks like we have a node test failure... anyone have any ideas why without diving in?

toranb:
@samselikoff  so looking at some source quick it [appears](https://github.com/outaTiME/applause/blob/master/src/applause.js#L177) that the original lib used `applause` to return an object like this ...

```js
  content = content.replace(match, replacement);
  // ... 
  return {
    content: content,
    detail: detail,
    count: total_count
};
```

Then from within the [broccoli-replace](https://github.com/outaTiME/broccoli-replace/blob/master/index.js#L44) library it would pull this off like so

```js
ReplaceFilter.prototype.processString = function(string) {
  var res = this.applause.replace(string);
  var result = res.content;
  var count = res.count;
  if (count === 0) {
    // no replacements
    return string;
  }
  return result;
};
```

From the [simpler broccoli](https://github.com/rwjblue/broccoli-string-replace/blob/master/index.js#L61) library from @rwjblue it would appear much simpler to follow

```js
SimpleReplace.prototype.processString = function (str) {
  for (var i = 0, l = this.patterns.length; i < l; i++) {
    var pattern = this.patterns[i];

    str = str.replace(pattern.match, pattern.replacement);
  }

  return str;
};
```

Looking at the test now and I don't see anything obvious without firing up a inspect debugging session. I'll dive back into this later dudes ;) sorry to punt this PR over and fail to solve this myself Sam

toranb:
screen shot 2018-06-21 at 7 18 42 pm

@rwjblue looks like yours returns an array w/ something called `Minimatch` inside of it. I'm still able to view into that array for what looks like the same assertion output but I'd like to get your take on this when you've got a minute

rwjblue:
Ya, ultimately these tests are a tad too implementation specific IMHO. The changes in this PR seem to function correctly, but the tests in question are failing because of internal details on how the actual broccoli trees are constructed.

My suggestion would be to rewrite [this test file](https://github.com/samselikoff/ember-cli-mirage/blob/master/node-tests/addon-tree-test.js#L26) such that it is using [broccoli-test-helper](https://github.com/broccolijs/broccoli-test-helper) (which has a really nice example with Mocha in its README).

samselikoff:
No it's ok! You can leave it open and others could help with updating the node tests, or you could open an issue for it

toranb:
@samselikoff cool - if anyone has a moment to jump in and rewrite those node tests this should be solid otherwise. I've updated to the `broccoli-replace-string` lib on all of my other addons and as usual :) Robert did a great job so you shouldn't see any impact from the switch

samselikoff:
@rwjblue I've picked this up. How might I assert against which files Mirage includes/excludes in the built `output`? I managed to see the `files.assets` and my dummy.js in there, but are there some examples of asserting against output? It's all concat'd by the time I'm looking at it.

samselikoff:
Perhaps I should be writing tests a bit lower level, against Mirage's broccoli code, rather than using EmberAddon() each time

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1361 supersedes.

Thanks Toran!

PR #1332, opened 2018-06-21
Support loading other scenarios in serve mode

scottkidder:
Took a shot at implementing my feature request [here](https://ember-cli-mirage.canny.io/feature-requests/p/support-for-loading-other-scenarios-in-ember-serve) for myself.

Feedback very welcome and I'm in no real rush to get this merged, I can add documentation as well if the approach is validated.

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -193,6 +193,15 @@ export default class Server {
     }
 
     let hasFactories = this._hasModulesOfType(config, 'factories');
+
+    let hasConfiguredScenario = !!config.scenario;
+    let configuredScenarioPresent = config.scenarios && config.scenarios.hasOwnProperty(config.scenario);
+
+    if (hasConfiguredScenario) {
+      assert(configuredScenarioPresent,
+        `You configured a scenario '${config.scenario}' but it was not found`);
+    }
+
     let hasDefaultScenario = config.scenarios && config.scenarios.hasOwnProperty('default');
 
     let didOverridePretenderConfig = (config.trackRequests !== undefined) && this.pretender;
@@ -214,6 +223,9 @@ export default class Server {
 
     if (this.isTest() && hasFactories) {
       this.loadFactories(config.factories);
+    } else if (!this.isTest() && hasConfiguredScenario) {
+      this.loadFactories(config.factories);
+      config.scenarios[config.scenario](this);
     } else if (!this.isTest() && hasDefaultScenario) {
       this.loadFactories(config.factories);
       config.scenarios.default(this);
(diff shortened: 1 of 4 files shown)


samselikoff:
Cool!

What's the dev experience like here, if you change the `environment` var will the server autoreload?

In our app this is what our `scenarios/default.js` looks like:

```js
import createAcceptanceTesting from './create-acceptance-testing';
import createEmberConcurrency from './create-ember-concurrency';
import createContextualComponents from './create-contextual-components';
import createEmberMapPodcast from './create-ember-map-podcast';
import createMirage from './create-mirage';
import createRefactorings from './create-refactorings';
import createBookshelves from './create-bookshelves';
import createNotes from './create-notes';
import userStates from './user-states';

// Choose one
const state =
  // 'anonymous'
  // 'unsubscribed'
  'subscribed'
;

  window.server = server;

  createAcceptanceTesting(server);
  createEmberConcurrency(server);
  createContextualComponents(server);
  createMirage(server);
  createEmberMapPodcast(server);

  createRefactorings(server);

  createBookshelves(server);

  createNotes(server);

  server.createList('plan', 3);
  server.create('plan', { price: 290000, interval: 'yearly' });

  userStates[state](server);
}
```

So we just change that `state` variable on line 12 if we want different "scenarios".

Originally my thought was the scenarios should be a first-class thing in Mirage but now I guess I'm not as convinced, since folks can just use functions/modules to split up their seeds logic. If we have to change a variable anyway and rebuild the app to see a new scenario, I don't see much difference in what we're doing above and something like an ENV variable (as in your PR). But I still might be persuaded, if only to have "one way" of separating these things.

However I am interested in any possible APIs we could use to "switch" scenarios without having to rebuild... so if we had multiple scenario files, how could we signal to Mirage to load different ones? Something like a query param (but that has issues), or even `localStorage`... `localStorage.mirage.scenario = 2`, then reload the browser. That way you could load a new scenario without rebuilding the app. 

Not sure – what do you think? What are some of your pain points that are hard to solve with JS modules/imports/functions?

scottkidder:
Thanks Sam, the example you've posted helps a lot and I see how that approach basically checks all the boxes.

I was also thinking of scenarios as first class citizens. Especially in the context of another feature I'm very interested in, sharing mirage factories, serializers, and scenarios from addons.

I'll move forward on our side with your suggestions above and see if I run into anything.

Thanks again!

samselikoff:
Closing for now, let me know how the above goes 👍 

Happy to revisit this feature!

PR #1334, opened 2018-06-27
Added `responseContentType` configuration option

piotrekbator:
Mirage's `Response` class internally assumes a default `application/json` content type.

This pull request adds an option to globally set a different `content-type` response header. This is essential as per JSON:API specification (http://jsonapi.org/format/) responses must have a `application/vnd.api+json` content type. Without this option it was impossible to use Mirage with Orbit.js which requires this header.

Example of usage:

````js
// mirage/config.js
this.responseContentType = 'application/vnd.api+json';
````

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "ember-cli-babel": "^6.8.2",
     "ember-cli-node-assets": "^0.2.2",
     "ember-get-config": "^0.2.2",
-    "ember-inflector": "^2.0.0",
+    "ember-inflector": "^2.0.0 || ^3.0.0",
     "ember-lodash": "^4.17.3",
     "fake-xml-http-request": "^2.0.0",
     "faker": "^3.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1384,7 +1384,7 @@ broccoli-babel-transpiler@^6.0.0, broccoli-babel-transpiler@^6.1.2:
     rsvp "^3.5.0"
     workerpool "^2.3.0"
 
-broccoli-babel-transpiler@^6.4.5:
+broccoli-babel-transpiler@^6.4.5, broccoli-babel-transpiler@^6.5.0:
   version "6.5.0"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by samselikoff** [comment]
Right now you can customize the headers if your route handler returns an instance of Response

```
return new Response(
  200,
  {
    'Content-Type': 'application/vnd.api+json'
  },
  this.serialize(model)
);
```

but I agree this is super annoying to do globally!

Why don't we instead add an option `this.headers`, where folks can specify the default headers used on every request?

```
// mirage/config.js
this.headers = {
  "Content-Type": "application/vnd.api+json"
}
```

**Comment by dfriquet** [comment]
Hello! I can see this PR is marked as `merged` but without any changes. I can't find anything related to this issue in the code, and I'm being annoyed by this "forced" `application/json` content-type – especially when using the `JSONAPISerializer`.

Could you share the status of this Feature/Enhancement with us?

**Comment by samselikoff** [comment]
@dfriquet I'm very confused. I also can't track down why this is marked as merged... I think @jembezmamy must have rebased and force-pushed with a commit that made it into master, overriding the work from before.

In any case we should open this feature request as a new issue rather than a pull request. Would you mind, with some details about your use case?

**Comment by piotrekbator** [comment]
@samselikoff @dfriquet it's really weird, indeed. This PR had a simple implementation that only allowed setting the `responseContentType`. @samselikoff suggested allowing global configuration of all response header (not just content type) and that would be better, but I had no time to expand on the idea.

I might have some time later this week to make a new PR. And I'll try talking with @jembezmamy about what might have happened with this PR.

**Comment by samselikoff** [comment]
For sure, sounds good.

You know I almost think we are going to add some sort of nested/module/beforeEach/afterEach API for things like this. We've already talked about it for things auth.

Interestingly `namespace` and `timing` and other route-specific options already work like this:

```js
this.namespace = 'api'; // applies to everything below
this.get('/users');
this.get('/posts');

this.namespace = 'api/admin'; // resets namespace for everything below
this.get('videos');
```

This is pretty confusing and actually really just an accident of the design (the fact that route handlers are created as `config.js` is read). I think it'd be great to have some sort of module/nesting ability here to set options for groups of routes. Will have to simmer on it. Or if anyone has any ideas.

PR #1335, opened 2018-06-27
Fix exists-sync deprecation by replacing with fs.existsSync

scottkidder:
Fixes this deprecation warning
```
warning ember-cli-mirage > exists-sync@0.0.3: Please replace with usage of fs.existsSync
```


samselikoff:
Thank you!

PR #1336, opened 2018-06-29
fix: serialize belongsTo key in camel-case

hugurp:
Mirage seems to use and expect camelCased keys for its relations, such as belongsTo.

Some application provide snake_cased keys, so in the case of a key being composed of multiple words, `snake_case_key`, Mirage will generate an invalid `snake_case_keyId` while it expects in fact `snakeCaseKeyId`.

This seems to fix the issue for the application I'm working on. I've investigated without success on a more appropriate place.

Am i missing or misunderstanding anything?

I am not sure if this solution should be enforces on other part of the code, and what test to write.

I'll hapilly implement any suggestion.

Cheers.

### DIFF
--- a/addon/serializer.js
+++ b/addon/serializer.js
@@ -298,7 +298,7 @@ class Serializer {
           newHash[formattedKey] = model[key].models.map((m) => m.id);
         } else if (association) {
           let formattedKey = this.keyForForeignKey(key);
-          newHash[formattedKey] = model[`${key}Id`];
+          newHash[formattedKey] = model[`${camelize(key)}Id`];
         }
       });
     } else if (this.serializeIds === 'included') {
@@ -318,8 +318,8 @@ class Serializer {
             let formattedTypeKey = this.keyForPolymorphicForeignKeyType(key);
             let formattedIdKey = this.keyForPolymorphicForeignKeyId(key);
 
-            newHash[formattedTypeKey] = model[`${key}Id`].type;
-            newHash[formattedIdKey] = model[`${key}Id`].id;
+            newHash[formattedTypeKey] = model[`${camelize(key)}Id`].type;
+            newHash[formattedIdKey] = model[`${camelize(key)}Id`].id;
           }
         } else {
           if (this.isCollection(association)) {
@@ -329,7 +329,7 @@ class Serializer {
           } else if (association) {
             let formattedKey = this.keyForForeignKey(key);
 
-            newHash[formattedKey] = model[`${key}Id`];
+            newHash[formattedKey] = model[`${camelize(key)}Id`];
           }
         }
       });


samselikoff:
Hm - so, Mirage does expect camelCased keys for its relationships, internally.

The reason for this is that Mirage's ORM relies on being able to know about foreign keys to look up relationships. If you think about Mirage's ORM & database, it's just a place where your data is stored, but not necessarily how that data will be transported over the wire back to your Ember app.

If your API uses snake_case over the wire, that should be handled in the serializer layer. So it's like this:

```
[ Server/DB/ORM (all camelCase, every app) ] =>  [ Serializer (maybe turn to snake_case) ]  =>  [ Ember ]
```

In practice this looks like customizing the `keyFor*` methods on whichever [serializer](http://www.ember-cli-mirage.com/docs/v0.4.x/serializers/) you're using. Take a look at the [ActiveModelSerializer](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/serializers/active-model-serializer.js) for an example.

Feel free to ask any questions here or jump into #ec-mirage in Slack if you're having trouble with that. But I don't believe this should need a code change! (The code change in your PR is for code "internal" to Mirage's data layer.)

xcambar:
> [ Server/DB/ORM (all camelCase, every app) ] =>  [ Serializer (maybe turn to snake_case) ]

By that, do you suggest that everything **in Mirage** but the serializer should use camelCase, including fixtures, models, factories...?

hugurp:
Hello Sam, thank you for your answer. I understand what you present.

One thing left that I do not get, is where/how I should fix the serializing behavior of the belongsTo relationships of my models.

In a dummy app similar to mine, there are 3 models, of which the main is the only one where the relationships are defined.
```
export default Model.extend({
  owner: belongsTo('owner'),
  tennis_court: belongsTo('tennis-court')
});
```

In mirage, I not use mirage models and rely on ember's definitions.

Here is an overview of the tow mirage serializers I use.
```
// ./mirage/serializers/house.js
import ApplicationSerializer from './application';

export default ApplicationSerializer.extend({
  include: ['owner', 'tennis_court']
});
```

```
// ./mirage/serializers/application.js
import { underscore } from '@ember/string';
import { RestSerializer } from 'ember-cli-mirage';
import { pluralize } from 'ember-inflector';

export default RestSerializer.extend({
  keyForModel(type) {
    return underscore(type);
  },
  keyForAttribute(attr) {
    return underscore(attr);
  },
  keyForRelationship(type) {
    return underscore(pluralize(type));
  },
  keyForEmbeddedRelationship(attributeName) {
    return underscore(attributeName);
  },
  keyForRelationshipIds(type) {
    return `${underscore(type)}_ids`;
  },
  keyForForeignKey(relationshipName) {
    return `${underscore(relationshipName)}_id`;
  },
  serialize(object) {
    let json = RestSerializer.prototype.serialize.apply(this, arguments);
    if (this.isCollection(object)) {
      json = json[pluralize(underscore(object.modelName))];
      return object.models.map((model)=> {
        return json.find((element)=> element.id === model.id);
      });
    }
    return json[object.modelName];
  }
});
```

With this configuration, the belongTo relationship for `owner` works correctly, because inside mirage (at the lines I "corrected" in this [PR file changes](https://github.com/samselikoff/ember-cli-mirage/pull/1336/files#diff-21df86ecb5f80627fb6dc6057c305faaR301)), the pseudo-code partial evaluation of `newHash[formattedKey] = model[`${camelize(key)}Id`];` is `newHash['owner_id'] = model['ownerId'];`.

With the relationship `tennis_court` (that has 2 words), the same line gives `newHash['tennis_court_id'] = model['tennis_courtId'];`.

In the first case (client model), because there is only one word in the relationship, it serializes well; as in the second with two words, the serialization fails (maybe because of the added `Id` in camel-case).

I fail to understand where I should specify/serialize my model relationship key. I would say that the best solution would be to have all my models and ember serializers use camel case instead of snake case property as the result of serialization, but it is not possible at this point in time.

I am not sure if I exposed the situation in a clear fashion, let me know if there is something I can explain, and/or if I am missing a point in the definition and serialization on the mirage side.

Thanks in advance.

samselikoff:
> By that, do you suggest that everything in Mirage but the serializer should use camelCase, including fixtures, models, factories...?

Yes. In older versions of Mirage the keys that were used in factories/models/etc. were also used in the actual JSON payload, but once we added Serializers that became their responsibility.

The idea is to use camelCase everywhere so "Mirage code" feels like standard JS code, where camelCase is the convention.

xcambar:
Food for thoughts:
Considering the autogenerated Mirage models (from ED), if (for any reason) the attributes in ED are not camelCased, the autogenerated Mirage models must be transformed.

PR #1337, opened 2018-06-29
Improve polymorphism support

samselikoff:
- Adds polymorphism support to shorthands
- Adds polymorphism support to normalizedRequestAttrs

Co-authored-by: Aidan Nulman <anulman@gmail.com>


jherdman:
Hey Sam. I've long had the hunch that some of our apps at Precision Nutrition tend to be flexing this feature of Ember Data, and Mirage, more thoroughly than most. I'd like to offer a bit of our time to help verify this didn't break anything. Would this be of interest to you? If so, what sort of things can I try out to sniff out the changes in this PR?

samselikoff:
Hey James! That'd be great 👍 

This doesn't affect the ORM layer – polymorphic support seems to be solid there.

If you'd like some history for this check https://github.com/samselikoff/ember-cli-mirage/pull/1153

The problem was with the shorthands, specifically `post` and `patch`. They weren't correctly taking the request payload and normalizing it into the form that Mirage's ORM expects.

The polymorphic tests added in this PR for the shorthands are failing in master. So, they were technically broken.

The way most folks were getting around this was to drop down and write a custom function handler. Here's one we have in EmberMap:

```js
this.post('/collection-items', function(schema, request) {

  attrs.itemizableId = JSON.parse(request.requestBody).data.relationships.itemizable.data;

  return schema.collectionItems.create(attrs)
});
```

We have this because `itemizableId` was basically getting turned into something like `1`, when Mirage needs it to look like `{ type, id }`, since its polymorphic.

With this PR, the data returned by `normalizedRequestAttrs` (which is both public and used by the shorthands) will be correct for polymorphic associations.

So, if you'd like to test this, I'd say first look to see if `normalizedRequestAttrs` is being used anywhere in your Mirage config, as the data returned from that will change for polymorphic associations. And second, to look at polymorphic associations that are hitting post/put shorthands.

Let me know if you have any questions or want to pair, and let me know how the testing goes! Happy to leave this open & unmerged until you get a chance to look.

jherdman:
I checked our code, and we unfortunately don't flex this edge case (this kind of surprised me). That said, I ran this branch against our test suite and all was golden!

samselikoff:
Thank you sir!

PR #1338, opened 2018-06-29
Improved BaseRouteHandler#getModelClassFromPath

bryanaka:
Right now, when you provide a path like `/assets/:objectId` and use `this.normalizedRequestAttrs()` in the route handler, it is doing naïve matching on `:id` to break out of the reverse path iterator when looking for the name to "classify" for the model. 

There is no restriction that I am aware of that says route path variables should be `:id`. This change will allow for named route path variables other than :id when using a path and handler that utilizes `BaseRouteHandler#getModelClassFromPath` in some form.


samselikoff:
Thank you sir!

bryanaka:
Thanks for all your hard work on Mirage! Happy to contribute. 😁

PR #1346, opened 2018-07-16
fix: associationFor should handle dasherized relationshipName

jelhan:
Relationship names are dasherized in most cases (e.g. `product-owner`) but properties of model are camelCase (e.g. `productOwner`). `associationFor()` returned `undefined` in that cases and therefore
`association.isPolymorphic` throws.

This was introduced by a19814654721bbce9131454ad72b4d931de9339a


jelhan:
Closed in favor of #1353, which comes with tests.

PR #1350, opened 2018-07-22
Add test to ensure correct behavior for polymorphic has many on RESTS…

samselikoff:
…erializer


samselikoff:
Supersedes https://github.com/samselikoff/ember-cli-mirage/pull/1282

PR #1352, opened 2018-07-27
Check if there is a server before shutting it down

mupkoo:
In one of our test cases, the `afterEach` cases is called, when there is no server defined, which results in an error and in the case of mocha this cause the whole test suite to fail. Although this is most probably a problem there, I think that it is worth having this simple check here as well.


mupkoo:
Looks like Travis is failing, because of yarn

PR #1353, opened 2018-07-27
Camelize the relationship name before getting the association

mupkoo:
The default behaviour of the JSON:API serializer is to dasherize the relationship names, so having a multiple word relationship name, will result in a block like `... "post-author": { data: null }...`.

Without this patch the `_getAttrsForRequest` fails, because it cannot find the correct association (_Cannot read isPolymorphic of undefined_)


mupkoo:
Looks like there is a PR that solves the same problem, but uses different approach
https://github.com/samselikoff/ember-cli-mirage/pull/1346

samselikoff:
Woah, thanks! Little embarrassed this wasn't covered by tests... might be good to go through the test cases and switch everything to multi-word examples.

PR #1355, opened 2018-07-28
Enable Ember addons to register `mirage-plugins`

anulman:
## Background

@samselikoff: per our Slack chat, this PR enables Ember addons to register themselves as `mirage-plugins` and provide a tree to merge down into the compiled application's mirage files.

The approach taken satisfies the following design criteria:

- **Minimalist API:** addon developers must only declare a keyword in their `package.json`, plus a `treeForMirage` attr / hook in their `index.js`;
- **No extra files:** any files included by the `treeForMirage` obey `e-cli-mirage`'s inclusion directives, so they do not show up in un-enabled prod builds;
- **No boilerplate:** addon developers can ship familiar models, serializers, factories, scenarios—even a default `config.js`!—in their `treeForMirage`s, which the host app can consume without configuration;
- **Easy to override:** addon `treeForMirage`s are merged with `{ overwrite: true }`, which lets host app files overwrite similarly-named addon files.

## Usage

To write your own plugin, first add `mirage-plugin` to the `keywords` array in your `package.json`:
```js
// my-addon/package.json
{
  ...,
  "keywords": [
    "ember-addon",
    "mirage-plugin"
  ],
  ...
}
```

Now add a `treeForMirage` to your `index.js`:
```js
// my-addon/index.js
module.exports = {
  // ...,
  treeForMirage: 'mirage',

  // OR
  treeForMirage() {
    return validateStuff() && 'other-mirage'
  },

  // OR
  treeForMirage() {
    let addonDir = path.dirname(this.constructor._meta_.modulePath);

    return new Funnel(path.join(addonDir, 'other-other-mirage'), {
      destDir: 'mirage'
    });
  },

  // ...
}
```

## Questions
- How to test this?
- How to teach this?

### DIFF
--- a/index.js
+++ b/index.js
@@ -177,8 +177,8 @@ function loadMiragePluginTrees(context) {
   let addons = context.project.addons || [];
 
   return addons
-    .filter((addon) => addon.pkg.keywords.includes('mirage-plugin')
-      && ['function', 'string'].includes(typeof addon.treeForMirage))
+    .filter((addon) => includes(addon.pkg.keywords, 'mirage-plugin')
+      && includes(['function', 'string'], typeof addon.treeForMirage))
     .map((addon) => {
       let treeForMirage = addon.treeForMirage;
 
@@ -198,4 +198,20 @@ function loadMiragePluginTrees(context) {
       return treeForMirage;
     })
     .filter((tree) => tree && tree.constructor.name === 'Funnel');
+}
+
+function includes(array, key) {
+  return Array.prototype.includes
+    ? Array.prototype.includes.call(array, key)
+    : _includes(array, key);
+}
+
+function _includes(array, key) {
+  for (let value of array) {
+    if (value === key) {
+      return true;
+    }
+  }
+
+  return false;
 }
--- a/index.js
+++ b/index.js
@@ -144,7 +144,7 @@ module.exports = {
   },
 
   _shouldIncludeFiles() {
-    if (process.env.EMBER_CLI_FASTBOOT) {
+    if (process.env.EMBER_CLI_FASTBOOT || typeof FastBoot !== 'undefined') {
       return false;
     }
 
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 'use strict';
+const fs = require('fs');
 const path = require('path');
 const mergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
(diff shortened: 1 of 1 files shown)


samselikoff:
Really awesome work man 👍  The community is gonna love you for this!

Ed uses a bunch of apps in his [ember-auto-import addon](https://github.com/ef4/ember-auto-import/tree/master/test-apps) to test different scenarios. I think we'll go the same route.

I'm working on switching over `node-tests/addon-tree-test.js` to this style so we'll have an example there soon, or if you want you can just start tinkering!

anulman:
Yay! Glad this will be useful for others :)

I took a look through the ember-auto-import test scripts and grok the approach. Are you suggesting we:

- Migrate the addon-tree-test.js into a default test-app; and
- Create n test apps, with node-tests similar to your addon-tree-test?

If so, we’ll probably want to route `yarn test` through a similar parallelization script; will need guidance there as I’m not clear which package.json test scripts are in use / how this will affect Travis.

Is there an issue, or other log of your work on the addon-tree-test switch I can use to spool up on intent? Else I’d prefer to wait until a pattern is in place, if it’s all the same to you.

samselikoff:
Yep, basically. I might have time today to move `addon-tree-test` into that newer style (though ideally I'd like to hear Robert's take on [this comment](https://discuss.emberjs.com/t/best-way-to-test-what-files-my-addon-is-providing-to-host-app/15199/6?u=samselikoff) first.)

I was going to keep it simple and use a single app, but thinking about features like this one and others, it might make more sense to have `n` test apps.

I'll work on this soon and get back to you :)

index.js
```diff
@@ -156,3 +172,46 @@ function npmAsset(filePath) {
...
+  let addons = context.project.addons || [];
+
+  return addons
+    .filter((addon) => includes(addon.pkg.keywords, 'mirage-plugin')
+      && includes(['function', 'string'], typeof addon.treeForMirage))
+    .map((addon) => {
+      let treeForMirage = addon.treeForMirage;
+
+      if (typeof treeForMirage === 'function') {
+        treeForMirage = addon.treeForMirage(addon);
```
anulman: TODO: this should be `addon.treeForMirage.call(addon)`

samselikoff:
Just merged https://github.com/samselikoff/ember-cli-mirage/pull/1359

I don't have any example of addons wired up yet to use all this, but following https://github.com/ef4/ember-auto-import/ we should be able to get it working. Ed has some more info on the [Running tests](https://github.com/ef4/ember-auto-import/#running-tests) section of his readme, and in `/test-apps` you'll see some sample addons and apps wired up there.

anulman:
Sounds good Sam. I've got a site to publish; hoping to pick this back up Friday / this weekend. That okay?

samselikoff:
Of course man, no rush

theseyi:
Any updates on this? Basically, I have an app with a monorepo architecture. Would like to keep mirage factories and models in the respective / related ember-addons, but still be able to run acceptance tests using those factories and models from the host application. Is there another approach to this problem? @samselikoff @anulman

samselikoff:

(I feel like this is a problem that other addons need an answer for so it'd be great if we could solve this outside of Mirage!)

PR #1358, opened 2018-07-28
fix: server.create() should apply factory even if used with pluralized names

### LINKED ISSUE #1357: factories are ignored if record is created using pluralized model name
`server.create()` supports creating record with singular and pluralized model names. But if pluralized model name is used, a factory for that record is ignored.

Assume having a model `post` with a factory defined.
```
server.create('post'); // factory is used
server.create('posts'); // factory is not used
```

Reproduction: https://ember-twiddle.com/c70f9095edc4d2608de07ea048e2dee6?openFiles=mirage.scenarios.default.js%2C

Please also note that a working twiddle does not require that much code as current boilerplate twiddle included in `ISSUE_TEMPLATE.md`. Maybe that one should be reduced to make it easier for new contributors.

jelhan:
Fixes #1357

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -2,7 +2,7 @@
 
 import { Promise } from 'rsvp';
 
-import { pluralize, camelize } from './utils/inflector';
+import { singularize, pluralize, camelize } from './utils/inflector';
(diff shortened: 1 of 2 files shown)


jelhan:
As discussed in #1357 passing pluralized model names to `server.create()` or `server.createList()` should not be supported.

PR #1374, opened 2018-08-14
Adds assert for association definition

Duder-onomy:
Hey! This repo is awesome. I have honestly never seen a test 'sweet' as expansive as this one.

I came across an issue in my code where I had not explicitly defined an association that I was posting via the relationships attr. (json:api) 

This resulted in a javascript exception on [this line](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/route-handlers/base.js#L66): around not being able to access the property `isPolymorphic of undefined`. 

This PR adds a simple assert in that code to help people fix this problem if they hit it. 

* Adds assert that ensures associations are defined when being included in json:api post requests

I tried to match the language style from other asserts.

### DIFF
--- a/addon/route-handlers/base.js
+++ b/addon/route-handlers/base.js
@@ -63,6 +63,11 @@ export default class BaseRouteHandler {
         let association = modelClass.associationFor(camelize(relationshipName));
         let valueForRelationship;
 
+        assert(
+          association,
+          `You're passing the relationship '${relationshipName}' to the '${modelName}' model via a ${request.method} to '${request.url}', but you did not define the '${relationshipName}' association on the '${modelName}' model. http://www.ember-cli-mirage.com/docs/v0.4.x/models/#associations`
+        );
+
         if (association.isPolymorphic) {
           valueForRelationship = relationship.data;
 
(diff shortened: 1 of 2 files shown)


samselikoff:
Excellent! And thanks for the kind words!

Great addition – can we get a test in here too?

Here's the test module for a simple function route handler: https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/route-handlers/function-handler-test.js.

You can see [here](https://github.com/samselikoff/ember-cli-mirage/blob/2ef1b592edd5e30db2b41d9a6103b69f1a3c5ea9/tests/integration/route-handlers/function-handler-test.js#L155) how to use `assert.throws` to write a test that expects an error to throw.

If you wanted, you could make a new file under `tests/integration/route-handlers` called `assertions-test` so you can set up a new server on its own.

Thanks again and let me know if you need any help! ✨

Duder-onomy:
@samselikoff No problem! I just pushed a test I added. 

I tried to match the style I found in other tests.

Let mw know if there is anything else I can do! Happy to help.

samselikoff:
This is awesome, thank you so much!

Issue #1168, opened 2017-08-29
Associations not properly updated when Mirage is minified and uglified


DLiblik:
In the `model` class within Mirage, `save` makes a call to `_disassociateFromOldInverses()`.  This code is written to depend on a string constructor name for the association, as can be seen here in [model.js](https://github.com/samselikoff/ember-cli-mirage/blob/85a6ac052f38ffd80f41b6d8e87332a680d33740/addon/orm/model.js#L513).

The issue here is that in a prod build, minification comes up with wonderfully efficient names for parameters (like `t` in the `_savesBelongsToAssociations()` signature, which in turn becomes the constructor name) instead of 'HasMany` and as a result, associations are not being updated at all in this call.

Here is a screenshot in which a breakpoint shows the parameter `e` holding an object of `t` (this is the re-prettified minified and uglified (and hence borkified!) code):

![image](https://user-images.githubusercontent.com/4047997/29841941-4d98e188-8cd5-11e7-9660-a58cd6f7fd5a.png)

Mirage is not a production facility (typically) but thought it was worth raising.

DLiblik:
For completeness, I have confirmed that compiling for production with minification disabled for vendor files does suppress the issue.

acorncom:
This may have just been resolved. See https://github.com/samselikoff/ember-cli-mirage/issues/1092#event-1232447876

DLiblik:
That definitely looks like it should do it, but I won't have time to test until early next week (not sure if there are other areas of Mirage that depend on constructor names - didn't do a search).

I will post back once I have a chance to re-run the test here.

Thanks for the #1092 ref.

acorncom:
@Dliblik any updates here? I'll close for now but feel free to comment if we need to re-open ...

DLiblik:
@acorncom a quick test appears to be fine based on changes in 60918b1 so I agree with the closure - thanks

Issue #1092, opened 2017-05-22
Relationships not working in production


luxferresum:
its seems like in a production build relationships are not working as expected.

I have enabled mirage with:

    ENV['ember-cli-mirage'] = {
      enabled: true
    };

[Here](https://github.com/luxferresum/mirage-not-working-on-production) is a git repo to demonstrate the error. Compare the result of `ember s` with `ember s --prod`.

In production the relationship `bars` on the `foo` model is not loaded.

samselikoff:
Believe these lines https://github.com/luxferresum/mirage-not-working-on-production/blob/master/config/environment.js#L49-L51 need to be within your `if environment === production` block.

luxferresum:
That does not make any sense. This code will run as well, its not in any `if`. However I tried that as well.

samselikoff:
I found it! `inverse.constructor.name = 'HasMany'` was breaking in production due to minification.

60918b19c45b926ce3a9e9c26126e8ea8a0e3cfa closes.

PR #1376, opened 2018-08-22
Use instanceof and overrides to distinguish associations

jamescdavis:
Similarly to #1092 and #1168, https://github.com/samselikoff/ember-cli-mirage/blob/99c927009a539f7e1fbfb8f6e857372e2319856b/addon/orm/associations/association.js#L52-L54 and https://github.com/samselikoff/ember-cli-mirage/blob/99c927009a539f7e1fbfb8f6e857372e2319856b/addon/orm/associations/association.js#L56-L58 fail in production builds due to minification.

This PR replaces usages of `isHasMany` and `isBelongsTo` with `instanceof` and uses overrides within the associations classes themselves (for `identifier`).

### DIFF
--- a/addon/orm/associations/association.js
+++ b/addon/orm/associations/association.js
@@ -1,4 +1,4 @@
-import { camelize, dasherize } from 'ember-cli-mirage/utils/inflector';
+import { dasherize } from 'ember-cli-mirage/utils/inflector';
 
 export default class Association {
 
--- a/addon/orm/associations/association.js
+++ b/addon/orm/associations/association.js
@@ -1,4 +1,4 @@
-import { camelize, singularize, dasherize } from 'ember-cli-mirage/utils/inflector';
+import { camelize, dasherize } from 'ember-cli-mirage/utils/inflector';
 
 export default class Association {
 
@@ -49,15 +49,7 @@ export default class Association {
     return this.opts.polymorphic;
   }
 
-  get isHasMany() {
-    return this.constructor.name === 'HasMany';
-  }
-
-  get isBelongsTo() {
-    return this.constructor.name === 'BelongsTo';
-  }
-
   get identifier() {
-    return this.isHasMany ? `${camelize(singularize(this.key))}Ids` : `${camelize(this.key)}Id`;
+    return `${camelize(this.key)}Id`;
   }
 }
--- a/addon/orm/associations/association.js
+++ b/addon/orm/associations/association.js
@@ -50,6 +50,6 @@ export default class Association {
   }
 
   get identifier() {
-    return `${camelize(this.key)}Id`;
+    throw new Error('Subclasses of Association must implement a getter for identifier');
   }
 }
--- a/addon/orm/associations/belongs-to.js
+++ b/addon/orm/associations/belongs-to.js
@@ -14,6 +14,10 @@ import assert from 'ember-cli-mirage/assert';
  */
 export default class BelongsTo extends Association {
 
+  get identifier() {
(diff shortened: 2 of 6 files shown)


jamescdavis:
[Here's an example](https://jamescdavis.github.io/ember-osf-web/) of Mirage failing in production with:

[mirage/factories/contributor.js](https://github.com/CenterForOpenScience/ember-osf-web/blob/5e74a00d0a1303044536d77304b13e21db83c395/mirage/factories/contributor.js#L8)
[app/models/contributor.ts](https://github.com/CenterForOpenScience/ember-osf-web/blob/5e74a00d0a1303044536d77304b13e21db83c395/app/models/contributor.ts#L29)

I have confirmed this PR fixes the issue.

samselikoff:
Dang... can we keep the `isHasMany` and `isBelongsTo` getters but change the methods to check `this instanceof HasMany` instead of using `constructor`?

jamescdavis:
The problem is, in order to use `instanceof` *inside* `class Association`, you'd have to import the child classes, which leads to circular imports.

jamescdavis:
If you *really* wanted to keep the getters, you'd *could* do:
```js
class Association {
  get isHasMany() {
    return false;
  }
  get isBelongsTo() {
    return false;
  }
}

class HasMany extends Association {
  get isHasMany() {
    return true;
  }
}

class BelongsTo extends Association {
  get isBelongsTo() {
    return true;
  }
}
```
but that seems kinds gross, IMHO.

jamescdavis:
Also, `e-c-mirage` is already using `instanceof` instead of the getters in several places:

https://github.com/samselikoff/ember-cli-mirage/blob/4aeec648194474f6b0354de4bd1a49e04ea51765/addon/orm/model.js#L358

https://github.com/samselikoff/ember-cli-mirage/blob/4aeec648194474f6b0354de4bd1a49e04ea51765/addon/orm/model.js#L370

https://github.com/samselikoff/ember-cli-mirage/blob/4aeec648194474f6b0354de4bd1a49e04ea51765/addon/orm/model.js#L621

https://github.com/samselikoff/ember-cli-mirage/blob/4aeec648194474f6b0354de4bd1a49e04ea51765/addon/orm/model.js#L802

addon/orm/associations/association.js
```diff
@@ -49,15 +49,7 @@ export default class Association {
...
-    return this.constructor.name === 'HasMany';
-  }
-
-  get isBelongsTo() {
-    return this.constructor.name === 'BelongsTo';
-  }
-
   get identifier() {
-    return this.isHasMany ? `${camelize(singularize(this.key))}Ids` : `${camelize(this.key)}Id`;
+    return `${camelize(this.key)}Id`;
```
samselikoff: could we move this to `BelongsTo` and do something here like `throw 'Subclasses of Association must implement identifier'`?

samselikoff:
Sorry for letting this sit! I agree we should move forward with `instanceof`. Just one small comment then we can merge 👍

addon/orm/associations/association.js
```diff
@@ -1,4 +1,4 @@
-import { camelize, singularize, dasherize } from 'ember-cli-mirage/utils/inflector';
+import { camelize, dasherize } from 'ember-cli-mirage/utils/inflector';
```
samselikoff:
```suggestion
import { dasherize } from 'ember-cli-mirage/utils/inflector';
```

jamescdavis:
Thanks, @samselikoff!

PR #1380, opened 2018-09-05
Loosen ember-inflector constraint to include 3.0.0

lukemelia:


lukemelia:
Hmm, having some trouble consuming this ember-cli-mirage branch from my app. Yarn seems to only partially install ember-cli-mirage.

lukemelia:
Seems like it was a temporary registry issue. Working now.

lukemelia:
IDK, still having issues with this with older yarn versions. Maybe better just to set to "^3.0.0"?

samselikoff:
Ya let's go with that 👍

john-griffin:
This upgrade is needed for apps that use `ember-cli-dependency-lint` and `ember-data` `3.4.0`.

samselikoff:
@john-griffin did specifying the range fix it for you?

@lukemelia did you end up just going with `^3.0.0`?

lukemelia:
@samselikoff My conclusion is that the range should work fine. I was confused as a result of the addon not building properly when it is referenced as a github repository (i.e. my fork).

john-griffin:
Other addons I have seen usually put a space before and after the `||` so it would be 
`"^2.0.0 || ^3.0.0"`

For now I have forced `3.0.0` in the resolutions sections of `package.json` as a temporary work around.

jherdman:
Any news on this change?

PR #1388, opened 2018-09-30
Update yarn.lock

jherdman:
Yarn v1.10 is out, and it's adding "integrity" stanzas everywhere. Let's
update this to make things a little less messy for everyone.


samselikoff:
Thanks! Maybe they'll figure out this whole dependency thing one day...

PR #1390, opened 2018-10-01
Adding this in front of server calls

jrock2004:


samselikoff:
Thank you, and sorry for the confusion! This changed with the new-style application tests.

I'm in the process of updating Mirage's docs on an AddonDocs site and will have more sections dedicated to docs about testing.

PR #1392, opened 2018-10-09
Support Pretender 2

jherdman:
This change does not, like originally intended, make Pretender a peer
dependency. There's a lot of work and debate being done here, so in the
mean time let's unlock `fetch` support in Mirage.

Dependency changes
------------------

* pretender 1.6 => 2.1. I've locked at 2.1.0 until
  pretenderjs/pretender#235 is resolved. This will require code changes
  on our end, and thus I don't want this sliding from underneath us and
  causing unnecessary weeping and gnashing of teeth.
* Add xg-wang/whatwg-fetch. Required for the upgrade.
* fake-xml-http-request 1.4.0 => 2.0.0. Required for the upgrade.
* route-recognizer 0.2.3 => 0.3.4. Required for the upgrade.
* ember-cli-node-assets 0.1.4 => 0.2.2. Because why not! I'm kneed deep
  in this stuff, so I may as well keep this up to date.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,13 +11,23 @@ module.exports = {
 
   options: {
     nodeAssets: {
-      'route-recognizer': npmAsset({
-        path: 'dist/route-recognizer.js',
-        sourceMap: 'dist/route-recognizer.js.map'
+      '@xg-wang/whatwg-fetch': npmAsset({
+        import: ['dist/fetch.umd.js']
       }),
-      'fake-xml-http-request': npmAsset('fake_xml_http_request.js'),
-      'pretender': npmAsset('pretender.js'),
-      'faker': npmAsset('build/build/faker.js')
+      'route-recognizer': npmAsset({
+        srcDir: 'dist',
+        import: ['route-recognizer.js'],
+        vendor: ['route-recognizer.js.map']
+      }),
+      'fake-xml-http-request': npmAsset({
+        import: ['fake_xml_http_request.js']
+      }),
+      'pretender': npmAsset({
+        import: ['pretender.js']
+      }),
+      'faker': npmAsset({
+        import: ['build/build/faker.js']
+      })
     }
   },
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -144,15 +154,24 @@ module.exports = {
   }
 };
 
-function npmAsset(filePath) {
+function npmAsset(options = {}) {
```
jherdman: Refactoring here required to silence a source maps warning for route-recognizer.

jherdman:
I'm fairly sure this works. I still need to smoke test this on a few apps of mine.

package.json
```diff
@@ -34,21 +34,22 @@
     "prepare": "./scripts/link.sh"
   },
   "dependencies": {
+    "@xg-wang/whatwg-fetch": "^3.0.0",
     "broccoli-funnel": "^1.0.2",
     "broccoli-merge-trees": "^1.1.0",
-    "broccoli-string-replace": "^0.1.2",
     "broccoli-stew": "^1.5.0",
+    "broccoli-string-replace": "^0.1.2",
```
jherdman: This was moved around when installing the various dependencies.

jherdman:
Smoke test done, and I can report gold!

samselikoff:
Sick!!

Was busy earlier this week, I know you pinged me about helping out. Sorry for being radio silent, and thank you so much James!

mydea:
Any chance of getting a release with this change? Would be awesome!

bjornharrtell:
This seems to be released since quite some time but fetch via mirage passthrough still doesn't work for me, getting `Unexpected end of JSON input` when parsing the response.

bjornharrtell:
Got it to work after all because it seems my problem was caused by the issue fixed by https://github.com/pretenderjs/pretender/pull/280. Had to override pretender transitive dependency.

samselikoff:
@bjornharrtell yikes, sounds bad. So you're saying ember-cli-mirage wasn't intercepting fetch requests?

bjornharrtell:
@samselikoff yep, afraid so.

samselikoff:
Any chance you could provide a reproduction?

Issue #1391, opened 2018-10-03
Tests do not run under IE11


john-griffin:
I tried running the last few version of mirage's test suite in IE11 but get many failures around `deepEqual` calls such as this one https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/orm/has-many/3-reflexive/accessor-test.js#L18. 

Is it worth adding IE11 to CI also?

samselikoff:
Yes, sounds like a great solution. Is it possible with Travis?

samselikoff:
Looks like Ember uses Browserstack, and they offer free keys for open-source projects.

Because there is some setup involved, I'm going to deprioritize this and mark it as a feature request right now (instead of bug), since the request is really to be able to develop Mirage in IE11.

Browserstack is a great solution and the total work might be small, but I just want to focus on some other higher-priority things right now. If you (or someone else) wants to work on this, let me know here and I can help out by signing up for Browserstack and uploading the credentials. The next step would be to add it to part of our Travis build, most likely following Ember.js's example.

Another short-term solution would be to fix the tests that are failing in IE11. These `deepEqual` assertions were fickle elsewhere in the codebase so if you look in some of the later modules, you'll see that we can write these older assertions that look like this

```js
let tag = ...

tags.forEach((t, i) => {
  assert.deepEqual(tag.tags.models[i], t)
```

like this

```js
tags.forEach(tag => {
  assert.ok(tag.tags.includes(tag));
```

So, if you want to submit a PR with those assertions updated so things work in IE11, that would also be a great approach!

samselikoff:
https://github.com/samselikoff/ember-cli-mirage/pull/1407#issuecomment-440826553 closes

PR #1407, opened 2018-11-20
Fix tests in IE11

john-griffin:
Gets the tests passing in IE11. Based on feedback in #1391.

### DIFF
--- a/tests/integration/orm/mixed/2-many-to-one/accessor-test.js
+++ b/tests/integration/orm/mixed/2-many-to-one/accessor-test.js
@@ -14,7 +14,12 @@ module('Integration | ORM | Mixed | Many To One | accessor', function(hooks) {
     test(`the references of a ${state} are correct`, function(assert) {
       let [ post, user ] = this.helper[state]();
 
-      assert.deepEqual(post.user, user);
+      if (post.user) {
+        assert.ok(post.user.equals(user));
+      } else {
+        assert.equal(post.user, null);
+        assert.equal(user, null);
+      }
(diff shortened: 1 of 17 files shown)


tests/integration/orm/mixed/2-many-to-one/accessor-test.js
```diff
@@ -14,7 +14,6 @@ module('Integration | ORM | Mixed | Many To One | accessor', function(hooks) {
     test(`the references of a ${state} are correct`, function(assert) {
       let [ post, user ] = this.helper[state]();
 
-      assert.deepEqual(post.user, user);
```
john-griffin: A regular `equal` call had issues here too. Do you have any suggestions?

samselikoff: Hm try `assert.ok(post.user.equals(user))`

samselikoff:
nice work man!

tests/integration/orm/mixed/2-many-to-one/accessor-test.js
samselikoff: Ah I see. I might do something like this:
```
if (post.user) {
  assert.ok(post.user.equals(user));
} else {
  assert.equal(post.user, null);
  assert.equal(user, null);
}
```

john-griffin:
Thanks, that last change is in, can confirm all passing in IE11 now.

samselikoff:
Nice work, and thanks!

PR #1408, opened 2018-11-20
Update paths in configuration.md

MichalBryxi:
- Remove all the occurrences of `app/` as a prefix for example file paths because it's not the default now.


samselikoff:
Thank you! 🙏

PR #977, opened 2017-01-03
custom identity managers per application and/or model

jelhan:
Adds support for custom identity managers per application and/or per model as asked for in #960. PR is based on v0.3.0-beta.4.

PR #1438, opened 2018-11-28
add docs for identity managers

jelhan:
Documentation was missing since implementation done in #977.


jelhan:
@samselikoff Was quicker for me to do it in the new app. I'm not that used to jekyll. Feel free to backport if you think it's needed.

samselikoff:
oh, sweet. Thank you!!

Also that's one of the big reasons for Addon Docs (not having to learn another tool), so great to hear!

PR #1443, opened 2018-12-12
Allow property getter and setters to work with `.extend`

ming-codes:
This enables computed properties like this.

```js
export default Model.extend({
  account: belongsTo(),

  get parent() {
    return this.account;
  }
});
```

### DIFF
--- a/addon/utils/extend.js
+++ b/addon/utils/extend.js
@@ -22,7 +22,7 @@ export default function extend(protoProps, staticProps) {
   // Add prototype properties (instance properties) to the subclass,
   // if supplied.
   if (protoProps) {
-    _assign(Child.prototype, protoProps);
+    Object.defineProperties(Child.prototype, Object.getOwnPropertyDescriptors(protoProps));
   }
 
   return Child;


samselikoff:
interesting. (Thanks for the contrib!)

Could you explain the use case here? Does this get serialized & sent across the network?

(I'm a little hesitant because this might affect the mental model around what belongs on the model. Traditionally it has just been "database" columns, i.e. the persisted fields and the foreign keys. Also, I'd like to move away from the `extend` module and to ES6 classes soon, but that's a separate issue.)

ming-codes:
The use case is that I want to be able to define a property alias that references another property.

If you really want to emulate a relational database, getters could be a computer column while setters could be used to enforce validation constraints 😁

Of course, doing all this on plain es6 classes requires no changes.

samselikoff:
How do you use the alias? In your Mirage code, or does it get sent to your Ember app in any way?

ming-codes:
Both actually. So our product’s relationship model is like a tree with 5 layers of hasMany. They reference each other by model name, but they can  also reference each other by parent / children alias. We have a generic search endpoint that can return anything in this tree by name and the endpoint must also return reference to the model’s parent id. If putting this as part of the serializer, it’ll quite a big stack of `if else`. So I got this idea of putting an alias on the model.

samselikoff:
Closing this as stale. (We need an API for synthetic/derived attributes, I think it should be Resources.)

PR #1450, opened 2019-01-03
Support detecting models in MU structure

HodofHod:
Like it says on the tin, this will detect models that use the standard Module Unification app structure, specifically those in `/src/data/models/<name>/`.

This is working flawlessly in my app, but some things need discussion:

### Testing 
I'm not quite sure how to go about adding tests for this. @samselikoff thoughts? Another app in `/test-projects`?

### More robust detection
 Module Unification (I believe) actually allows users to specify other places to put modules, although for many reasons it's unlikely that that will be a common use case for ember-data models. 

   Nevertheless, it might be good to use a more robust method for detecting models than filtering `requirejs.entries` based on module paths. E.g.,
`appInstance.lookup('data-adapter:main').getModelTypes()`
  
   That seems solid and would even detect classic- and pod- models. I'm not sure how that would work with Mirage's lifecycle, since the model detection doesn't currently have access to the `appInstance`.

### DIFF
--- a/addon/ember-data.js
+++ b/addon/ember-data.js
@@ -29,6 +29,7 @@ export function getDsModels() {
   let moduleMap = requirejs.entries;
   let classicModelMatchRegex = new RegExp(`^${modulePrefix}/models/(.*)$`, 'i');
   let podModelMatchRegex = new RegExp(`^${podModulePrefix}/(.*)/model$`, 'i');
+  let muModelMatchRegex = new RegExp(`^${modulePrefix}/src/data/models/(.*)/model$`, 'i');
 
   DsModels = {};
 
@@ -38,7 +39,7 @@ export function getDsModels() {
 
   Object.keys(moduleMap)
     .forEach((path) => {
-      let matches = path.match(classicModelMatchRegex) || path.match(podModelMatchRegex);
+      let matches = path.match(classicModelMatchRegex) || path.match(podModelMatchRegex) || path.match(muModelMatchRegex);
       if (matches && matches[1]) {
         let modelName = matches[1];
 


samselikoff:
Given MU is on hold, should we close this PR and revisit when the MU design finalizes?

HodofHod:
Possibly. However, there are many applications that are already using the MU structure, so it wouldn't be valueless to still include this. 

At the very least, it makes sense to support configuring ec-mirage to find models in non-standard places.

dfreeman:
I'd definitely be wary of hard-coding anything MU-specific at this point, but we ran into an issue recently with engines where being able to configure a lookup path other than `env.modulePrefix` would have been useful.

To run integration tests for an engine, you have to set up the test module to use the engine's resolver instead of the dummy app's so that you can test things like components and services within the engine. Unfortunately this means Mirage tries to find its model config under the engine's module prefix, when I think it's preferable to keep them with the dummy app since you typically wouldn't want to publish them alongside the engine code.

I think ideally that scenario would 'just work' out of the box, but being able to configure the lookup would be a nice escape hatch in the meantime.

HodofHod:
@dfreeman I think you're probably right about not hard-coding MU, (though I don't think it'd be much worse than the currently hard-coded standard and pods structures).

I'd love to work on adding support for configuring mirage's detection. Ideally it'd be something more than just using `requirejs.entries`, like I mentioned in the opening comment here, but I don't know if Ember supports that well. 

Barring that, what do you (all) think of allowing the application to specify alternative regex patterns for filtering `requirejs.entries`?

dfreeman:
🤦‍♂ I should have read the PR more closely—the issue we ran into was actually with the detection of Mirage models, rather than Ember Data models. Regardless, being able to customize the regex(es) for both of those things feels like it would be useful.

samselikoff:
On the Ember Data side, is there any way to use Ember CLI's resolver to get the models map?

samselikoff:
Closing this as stale, especially given MU has been discarded!

PR #1454, opened 2019-01-07
Improve automatic response type setting

wagenet:
Simplifies checking for blank response to merely change to 200 when
the response type would have been a 204 and the response is not
undefined or an empty string. This also matches the heuristic in the
204-non-empty-payload warning.


samselikoff:
Makes sense. Thank you so much for your help! 🙏

PR #1457, opened 2019-01-17
Allow native async functions as route handlers

dfreeman:
I had a handler that looked something like this, and was baffled for a while this afternoon as to why the callback was never triggering even though I saw Mirage logging that it had handled the request:

```js
this.post('/some-path', async () => {
  // some promise-y stuff
});
```

It turns out, on platforms where `async` functions are supported natively, `Ember.typeOf` incorrectly [identifies them as objects](https://github.com/emberjs/ember.js/issues/15151). Ultimately it was decided [not to extend `Ember.typeOf` to handle this](https://github.com/emberjs/ember.js/pull/15165) in favor of native `typeof` instead.

### DIFF
--- a/addon/route-handler.js
+++ b/addon/route-handler.js
@@ -16,11 +16,11 @@ const DEFAULT_CODES = { get: 200, put: 204, post: 201, 'delete': 204 };
 function createHandler({ verb, schema, serializerOrRegistry, path, rawHandler, options }) {
   let handler;
   let args = [schema, serializerOrRegistry, rawHandler, path, options];
-  let type = typeOf(rawHandler);
+  let type = typeof rawHandler;
 
   if (type === 'function') {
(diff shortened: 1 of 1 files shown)


samselikoff:
Nice, thanks for the PR!

Any chance you could add a quick test, probably [here](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/route-handlers/function-handler/basic-test.js)? Just something simple like

```
test(`it works with async functions`, async function(assert) {
  assert.expect(1);

  this.server.get('/foo', async () => {
    return "bar";
  });

  let { data } = await promiseAjax({
    method: 'GET',
    url: '/foo'
  });

  assert.equal(data, "bar");
});
```

samselikoff:
Ah, right. I see.

I think if this is just too hard to test this change is fine as is!

PR #1471, opened 2019-01-29
Allow users to use their own version of Pretender

samselikoff:


samselikoff:
Closing in favor of simpler approach. Check the discussion here for reference:

https://github.com/ef4/ember-auto-import/issues/184

PR #1478, opened 2019-02-02
fix: shorthand link

Alonski:
There is also an issue with:
See the `serializer guide` to learn more.

Not sure where it should point to


samselikoff:
tyvm :D 

Going to be going through docs with a fine tooth comb soon. Appreciate the help!

PR #1480, opened 2019-02-05
Fix prefix example code

lindyhopchris:
Believe this needs to be `this.urlPrefix` instead of the existing `this.prefix`.


samselikoff:
ty sir!

PR #1481, opened 2019-02-05
Omit default Content-Type header for 204 responses

simonihmig:
As a 204 response has no response body, a Content-Type is basically meaningless. It doesn't seem to be a violation of any spec to have one AFAICT, but as it is optional and does not make sense here, it seems better to not add it automatically.

### DIFF
--- a/addon/response.js
+++ b/addon/response.js
@@ -44,7 +44,7 @@ export default class Response {
     }
 
     // Default "untyped" responses to application/json
-    if (!this.headers.hasOwnProperty('Content-Type')) {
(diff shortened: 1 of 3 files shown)


simonihmig:
Btw, got an error with orbit.js, as it tried to `JSON.parse()` the empty body, because it thinks there is content when `Content-Type` is set. See https://github.com/orbitjs/orbit/blob/master/packages/%40orbit/jsonapi/src/jsonapi-source.ts#L285-L295

samselikoff:
Makes sense. 

Turns out making a well-behaved server is tricky!

Thanks so much for the great PR 🤙

PR #1503, opened 2019-02-15
remove Pretender shim

givanse:
~- [ ] depends on the release of Pretender v3.0.0 (pending https://github.com/pretenderjs/pretender/pull/248)~
 - [x] https://github.com/pretenderjs/pretender/pull/255

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ember-get-config": "^0.2.2",
     "ember-inflector": "^2.0.0 || ^3.0.0",
     "lodash": "^4.17.11",
-    "pretender": "2.1.1"
+    "pretender": "3.0.1"
   },
   "devDependencies": {
     "@ember/jquery": "^0.6.0",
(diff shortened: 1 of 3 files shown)


samselikoff:
Oh yeah – we haven't been using `shim` since I added auto-import, correct?

givanse:
Correct, ember-auto-import has the magic to make the Pretender IIFE available as if it was an ES6 module. The shim wasn't necessary since then.

With this PR now Pretender can be imported as a native ES6 class, so auto-import is not needed anymore for this dependency (others still need it).

samselikoff:
Ah, very cool!

package.json
```diff
@@ -46,7 +46,7 @@
     "ember-get-config": "^0.2.2",
     "ember-inflector": "^2.0.0 || ^3.0.0",
     "lodash": "^4.17.11",
-    "pretender": "2.1.1"
+    "pretender": "3.0.1"
```
Turbo87: @samselikoff is there a reason why `pretender` is pinned to a specific version? I would like to take advantage of the bug fixes in v3.0.2, but due to the pinning I can't :-/

samselikoff:
@Turbo87 don't believe so, not that I can recall! We can definitely add a ^ there

samselikoff:
mind opening the PR?

PR #1559, opened 2019-03-25
Fix internal link to Schema docs

nino:
The current link leads to a "Not found" page. This PR changes the target to what I'm guessing is the intended location.


samselikoff:
🙏 Thank you sir!

PR #1561, opened 2019-03-26
Save over 500KB with one weird trick

ef4:
This PR cuts more than half a megabyte from any app that uses mirage.

The problem is that:

```js
import _ from 'lodash';
let item = _(value);
```

needs to put methods on `item` for every possible thing lodash supports. Most of which aren't needed.

This isn't one of the cases where a more aggressive module bundler could try to optimize them out. This style really just can't be statically optimized. It's a pity, because the chaining style is easier to read IMO.


samselikoff:
boss.

PR #1562, opened 2019-03-26
Remove duplicated dependencies

ef4:
These are dependencies of pretender. They stopped being necessary when mirage switched to loading pretender via ember-auto-import. When consumed that way, pretender already does the right thing and includes its own dependencies automatically.

Prior to this PR, all of these were being included twice.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,22 +8,6 @@ const writeFile = require('broccoli-file-creator');
 module.exports = {
   name: 'ember-cli-mirage',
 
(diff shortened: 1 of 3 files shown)


samselikoff:
Thank you very much!

How'd you discover this?

ef4:
I am testing embroider against ember-observer, which uses mirage. So I’ve
been going through the bundles to figure out where all the bytes come from.

Issue #1560, opened 2019-03-25
Missing documentation in serializer.js?


nino:
In [`addon/serializer.js`](https://github.com/samselikoff/ember-cli-mirage/blob/be23f538131228d1deeec1aedf9a2a51f0a7957f/addon/serializer.js#L902), the following methods don't have any documentation:

- `keyForForeignKey`
- `keyForPolymorphicForeignKeyId`
- `keyForPolymorphicForeignKeyType`

Since they are not marked `@private`, and don't have a leading `_`, I'm assuming they are meant to be public. Is that correct?

samselikoff:
Correct! Just laziness on my part... would love some help!

From the commit for `keyForForeignKey`:

```
- new serializer hook: `keyForForeignKey`
  - used for belongsTo relationships (keyForRelationshipIds)
  - TODO: this is an awful name, change it
```

samselikoff:
d34784201bfe1e5c2e7c331e7c5c54e26d8aa152 closes

dgaus:
keyForForeignKey is still undocumented

samselikoff:
I believe it's there on master's docs ya? https://www.ember-cli-mirage.com/versions/master/docs/api/modules/ember-cli-mirage/serializer~Serializer#keyForForeignKey

dgaus:
You're right, sorry, I missed that. Thanks!

>

PR #1568, opened 2019-03-29
Document keyForForeignKey method

nino:
Addresses part of #1560.


samselikoff:
Thank you!! Wow this is badly named 🙈

PR #1589, opened 2019-04-17
Tests coverage for Schema.first() and Schema.findBy()

### LINKED ISSUE #1567: Clear behavior of Schema.first() on empty collection and Schema.findBy() on empty result
It's currently not defined how `first()` and `findBy()` methods of a `Schema` behave if schema doesn't have any record (`first()`) or doesn't have any matching record (`findBy()`).

This case isn't mentioned in docs for either of them. It's not [covered by tests for `schema.first()`](https://github.com/samselikoff/ember-cli-mirage/blob/master/tests/integration/orm/first-test.js). I wasn't able to find any tests for `Schema.findBy()`.

There is a similar case of `DbCollection.findBy()`. It's docs also doesn't cover how it behaves if no record matches the criteria but it has test coverage for that scenario: https://github.com/samselikoff/ember-cli-mirage/blob/20d9d84841ebbeb7672a313282b8a03a0bfe1338/tests/unit/db-test.js#L276-L280

Based on that test coverage, which exists since implementation of that feature, I would expect that `Schema.findBy()` and `Schema.first()` also return `null` if no record is found. This should be covered by tests and mentioned in docs for all of them.

Oreoz:
Attempts to fix #1567.

Not quite sure if the tests are located where they should be. Happy to revisit with a little guidance.

I also removed `bower i` from CONTRIBUTING.md since when I copied the command it "errored" since there is no bower.json file anymore.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,20 +18,20 @@ To make more substantial changes, you'll want to be able to write locally. The d
     git clone git@github.com:[your-name]/ember-cli-mirage.git
     cd ember-cli-mirage
     ```
-    
+
 2. Checkout the `gh-pages` branch:
 
     ```
     git checkout gh-pages
     ```
-    
+
 3. Start the Jekyll build:
 
     ```
     bundle install
     bundle exec jekyll serve --watch --baseurl ''
     ```
-    
+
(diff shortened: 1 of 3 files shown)


samselikoff:
Thank you so much!

PR #1601, opened 2019-04-24
do not whitelist global server in blueprint

jelhan:
There isn't any need for a global `server` variable in new testing API. Allowing it in ESLint and JSHint configuration isn't necessary for new apps anymore.


samselikoff:
Lovely!

Issue #1429, opened 2018-11-21
Trait composition


samselikoff:
![image](https://user-images.githubusercontent.com/2922250/48871579-51141680-edb3-11e8-8538-f490122baf0f.png)

Oreoz:
@samselikoff Opened a PR that attempts to add this feature. Happy to discuss. :)

samselikoff:
Latest issue: https://github.com/miragejs/discuss/issues/24

PR #1604, opened 2019-04-26
Trait Composition

Oreoz:
My first crack at #1429. Comments welcomed -- not sure if I've covered everything. Happy to discuss and modify as needed.


samselikoff:
Very cool - nice work!

I wonder if we should think about args to traits before landing on an API for composition. Args is something folks have wanted for a while.

Something like

```js
server.create('post', { trait: 'withComments', args: { count: 3 }});
```

but a less crappy API.

I've wondered if we should take inspiration from [ExMachina](https://thoughtbot.com/blog/announcing-ex-machina) and try to reduce the number of concepts in Mirage in favor of JS primitives, assuming we can still land on an ergonomic API.

Oreoz:
If we're to add support for args to traits it makes total sense to hold off on this.

Did you want to move discussion about the new API to the issue instead of this PR? It would make it easier for everyone to follow along and chime in.

I have some work time I can devote to OSS and I want to give back to your project since it's been a great help in my team. So I could take a stab at a trait "overhaul" once we have a clearer idea of what we want as a final API.

samselikoff:
Closing this , I believe https://github.com/miragejs/ember-cli-mirage/issues/1422 takes precedence.

PR #1605, opened 2019-04-30
export setupMirage from test-support directly

jelhan:
Before:
```js
import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
```

After:
```js
import { setupMirage } from 'ember-cli-mirage/test-support';
```

It's less verbose. I also think it's more inline with rest of the ecosystem. Have a look at [Ember Power Select](https://ember-power-select.com/docs/test-helpers), [Ember Simple Auth](https://github.com/simplabs/ember-simple-auth#testing) and [Ember Lifeline](https://ember-lifeline.github.io/ember-lifeline/docs/testing) for example.

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,0 +1,2 @@
+import setupMirage from './setup-mirage';
+export { setupMirage };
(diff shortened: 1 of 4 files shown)


samselikoff:
Agree. Thank you @jelhan!

PR #1613, opened 2019-05-10
A model with two hasMany relationships to the same type doesn’t properly store both attributes

backspace:
Hey, I’m running into an obscure seemingly-bug in [this open-source application](https://github.com/backspace/prison-rideshare-ui) I maintain. This test demonstrates the problem in a more focused way, but here’s the context from the production application in case it’s helpful:

The relevant portion of the data model is that there are (car) `ride`s that can be assigned to `driver`s and `carOwner`s, which are both of type `person`. Here’s the awkward nomenclature:

```
ride:
  driver: belongsTo('person', { inverse: 'drivings' }),
  carOwner: belongsTo('person', { inverse: 'carOwnings' }),

person:
  drivings: hasMany('ride', { inverse: 'driver' }),
  carOwnings: hasMany('ride', { inverse: 'carOwner' }),
```

I ran into this because in the tests, I have a `PATCH` to update a `ride` by assigning a `driver` and `car-owner`, it looks like this abridged JSON:

```
"data": {
    "id": "1",
    "attributes": {
      …
    },
    "relationships": {
        "driver": {
            "data": {
                "type": "people",
                "id": "1"
            }
        },
        "car-owner": {
            "data": {
                "type": "people",
                "id": "1"
            }
        }
    },
    "type": "rides"
}
```

Here’s what comes back from Mirage (again abridged):

```
        "data": {
            "type": "rides",
            "id": "1",
            "attributes": {
                …
            },
            "relationships": {
                "driver": {
                    "data": {
                        "type": "people",
                        "id": "1"
                    }
                },
                "car-owner": {
                    "data": {
                        "type": "people",
                        "id": "1"
                    }
                }
            }
        },
        "included": [
            {
                "type": "people",
                "id": "1",
                "attributes": {
                    …
                },
                "relationships": {
                    "drivings": {
                        "data": [] // <- shouldn’t be empty!
                    },
                    "car-ownings": {
                        "data": [
                            {
                                "type": "rides",
                                "id": "1"
                            }
                        ]
                    }
                }
            }
        ]
```

The key thing to notice is that the `drivings` relationship is incorrectly empty.

I came across this because it seems like Ember Data 3.1 was ignoring the empty `drivings` relationship on the included `person` record, but Ember Data 3.2 interprets it properly, resulting in the relationship being unassigned after the server’s response is processed.

This failing test doesn’t faithfully reproduce the context of the failure I’m experiencing because I couldn’t figure out how to get the `included` attribute to show up in the `PATCH` response. I added a `GET` instead, which reveals the problem too, but less concisely. It’s strange and inexplicable to me that in my application, the `relationships.driver` is correct, but not the `included`.

It’s possible this test isn’t really at the correct level, as it seems like maybe it’s more of an ORM problem than a server one, but I wasn’t sure how to replicate it otherwise.

I know you’re always busy with your various projects; this is something I could look into fixing myself, I thought maybe you’d have better insight into where to start, but I’d be happy to dig in if you don’t have time.

Let me know if you need any more information or if you have a hint about where I could start on fixing this.

As always, thanks for your work on Mirage. I tried out the latest beta today and everything went smoothly! 💞

### DIFF
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -885,7 +885,10 @@ class Model {
 
         // Non-self-referential
         } else if (!tempAssociate.isSaving) {
+
+          // Save the tempAssociate and update the local reference
           tempAssociate.save();
+          this._syncTempAssociations(tempAssociate);
 
           let fkValue;
           if (association.isPolymorphic) {
@@ -976,6 +979,24 @@ class Model {
   // which can cause cycles with associations.
   _updateInDb(attrs) {
     this.attrs = this._schema.db[toInternalCollectionName(this.modelName)].update(this.attrs.id, attrs);
+  }
+
+  /*
+  Super gnarly: after we save this tempAssociate, we we need to through
+  all other tempAssociates for a reference to this same model, and
+  update it. Otherwise those other references are stale, which could
+  cause a bug when they are subsequently saved.
+
+  This only works for belongsTo right now, should add hasMany logic to it.
+
+  See issue #1613: https://github.com/samselikoff/ember-cli-mirage/pull/1613
+  */
+  _syncTempAssociations(tempAssociate) {
+    Object.keys(this._tempAssociations).forEach(key => {
+      if (this._tempAssociations[key].toString() === tempAssociate.toString()) {
+        this._tempAssociations[key] = tempAssociate;
+      }
+    });
   }
 
   /**
--- a/addon/orm/model.js
+++ b/addon/orm/model.js
@@ -993,7 +993,7 @@ class Model {
   */
   _syncTempAssociations(tempAssociate) {
     Object.keys(this._tempAssociations).forEach(key => {
-      if (this._tempAssociations[key].toString() === tempAssociate.toString()) {
+      if (this._tempAssociations[key] && this._tempAssociations[key].toString() === tempAssociate.toString()) {
         this._tempAssociations[key] = tempAssociate;
       }
     });
--- a/tests/integration/orm/mixed/regressions/1613-two-bidirectional-many-to-many-with-same-target-model-update-bug-test.js
+++ b/tests/integration/orm/mixed/regressions/1613-two-bidirectional-many-to-many-with-same-target-model-update-bug-test.js
@@ -1,0 +1,34 @@
+import { module, test } from 'qunit';
+import { Model, hasMany, belongsTo } from 'ember-cli-mirage';
+import Schema from 'ember-cli-mirage/orm/schema';
+import Db from 'ember-cli-mirage/db';
+
+module('Integration | ORM | Mixed | Regressions | 1613 Two bidirectional one-to-many relationships with same target model update ids bug', function(hooks) {
+  hooks.beforeEach(function() {
+    this.db = new Db();
+
+    this.schema = new Schema(this.db, {
+      user: Model.extend({
+        authoredPosts: hasMany('post', { inverse: 'author' }),
+        editedPosts: hasMany('post', { inverse: 'editor' })
+      }),
+      post: Model.extend({
+        author: belongsTo('user', { inverse: 'authoredPosts' }),
+        editor: belongsTo('user', { inverse: 'editedPosts' })
+      })
+    });
+  });
+
+  test(`it works, and all inverses are correctly updated`, function(assert) {
+    let user = this.schema.users.create();
+    let post = this.schema.posts.create();
+
+    post.update({
+      authorId: user.id,
+      editorId: user.id
+    });
+
+    assert.deepEqual(this.db.posts.find(1), { id: '1', authorId: '1', editorId: '1' });
+    assert.deepEqual(this.db.users.find(1), { id: '1', authoredPostIds: [ '1' ], editedPostIds: [ '1' ] });
+  });
+});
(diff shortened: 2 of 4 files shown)


backspace:
hmm, unfortunately the [failure](https://travis-ci.org/samselikoff/ember-cli-mirage/jobs/530583365#L2721-L2736) only shows the classique Javascript `[object Object]`, but this is what’s produced for me locally:

```
{
  "attributes": {},
  "id": "1",
  "relationships": {
    "authorings": {
      "data": []
    },
    "editings": {
      "data": [
        {
          "id": "1",
          "type": "posts"
        }
      ]
    }
  },
  "type": "users"
}
```

backspace:
I did some step-through debugging to try to figure this out and I don’t have any grand revelations, but I did find that maybe [this call](https://github.com/samselikoff/ember-cli-mirage/blob/master/addon/orm/model.js#L888) to save the post’s editor is where the author gets erased.

| | before `tempAssociate.save()` | after |
| --- | --- | --- |
| `post.author.attrs.authoringIds` | `["1"]` | `[]` |
| `post.author.attrs.editingIds` | `null` | `["1"]` |

I have to put this aside for now but hopefully I can come back to it soon.

samselikoff:
Been working on this today, thanks for the detailed writeup!

This is a pretty gnarly bug. Still working on the fix.

The problem is in the ORM layer. Turns out if you do

```js
post.update({
  author: user
});
post.update({
  editor: user
});
```

everything works, but updating them together

```js
post.update({
  author: user,
  editor: user
});
```

fails. The reason does indeed have to do with `_tempAssociations`.

When Mirage calls

```js
post.update({
  editor: user
});
```

it stores a reference to that `user` on the post's `_tempAssociations`. That's because that `user` could be a new user with no `id` yet, and it could have associations of its own. So, we store a reference to it, and as part of the saving process we call `.save()` on that temp associate as well.

So the bug with

```js
post.update({
  author: user,
  editor: user
});
```

is happening because Mirage is keeping a reference to the `user` in two places: `_tempAssociations.author` and `_tempAssociations.editor`.

When it goes through and saves the `author` relationship, the inverse is updated so that `user.authoringIds` is updated. The problem is, when it moves on to save the `editor` relationship, its working with a stale `tempAssociate`. That version of the `user` doesn't have the new `authoringIds`, so when it gets saved and updated, it's using the old version of the model which has no `authoringIds`.

samselikoff:
This raises an interesting question. What happens if someone does

```js
post.update({
  author: tempVersionUser1,
  editor: anotherTempVersionUser1,
});
```

and those two versions have inconsistent data of some sort.

This makes me think it'd be nice if all New records had some sort of `guid`, and saving/updating a graph of data could just work by using `guid`s to reference objects in that single graph.

samselikoff:
@backspace tests pass – want to give this a shot in your app and make sure it works there before I merge?

backspace:
Success! The test that was failing [here](https://travis-ci.org/backspace/prison-rideshare-ui/builds/532038026#L712) passes [here](https://travis-ci.org/backspace/prison-rideshare-ui/builds/532038302#L670) after I switched to this branch. The other failures are unrelated, probably from having jumped from Ember Data 3.1 to 3.10. Thanks for figuring it out!

[That](https://github.com/samselikoff/ember-cli-mirage/pull/1613#issuecomment-491949393) is a bit of a puzzle… in my case it wouldn’t be possible because the changes were happening through JSON:API, but it does seem confusing when changing the Mirage database directly 🤔

backspace:
hmm but maybe the fix has caused problems elsewhere. In [this test](https://github.com/backspace/prison-rideshare-ui/blob/2c236b4734502af7e3804c7ac20dbd6488945b44/tests/acceptance/reports-test.js#L61), the Mirage route handler now [throws a 500](https://travis-ci.org/backspace/prison-rideshare-ui/builds/532038302#L678-L683) when it’s meant to succeed.

I’ve tracked the exception to [this line](https://github.com/samselikoff/ember-cli-mirage/pull/1613/files#diff-9eee398c74e24a8173ecd103ce7691b2R996), where `this._tempAssociations[key]` is `null`, therefore `toString()` fails. When I’m debugging, `key` is `driver`, but if I look at `this._tempAssociations`, this is the whole object, which doesn’t bode well for comparisons: `{driver: null, carOwner: null}`

backspace:
This naïve change to that line restored the failing test to passing, so I added it in 2a26f4b.

```
      if (this._tempAssociations[key] && this._tempAssociations[key].toString() === tempAssociate.toString()) {
```

Could it be that easy? Maybe so, as with it, the other failures also disappeared and my whole test suite [passes](https://travis-ci.org/backspace/prison-rideshare-ui/jobs/532082442)! 🎉

That line is getting pretty long; I didn’t see any line-length restrictions, but could make it multiline or store the left portion of the comparison for checking.

samselikoff:
It's ok by me, it's a private detail, pretty edge-casey and likely to change at some point.

I think this looks good! Surprised the test suite didn't catch that as I know we test setting `null` relationships many places. I'll keep this all in mind if a related issue comes up in the future.

Thanks for following through!

backspace:
No problem, thanks for fix, it would definitely have taken me a long time to figure out something like it.

PR #1615, opened 2019-05-10
Fix where the  docs reference wrong method

barryofguilder:
It seems that the documentation for the `findOrCreateBy` method on the schema was referencing `findBy` instead. This updates it so it's using the correct method.


samselikoff:
Thanks Jason!

PR #1631, opened 2019-05-23
JSONAPISerializer Enhancements (Paging & Sorting)

Oreoz:
Was about to copy-paste my route handler paging and sorting functions for the nth time, figured I'd implement this. Congrats on 1.0 Sam! 🎉 :)

#1418 #1420

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ jekyll-tmp
 # IDE
 /.idea/
 /*.iml
+
+package-lock\.json
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -5,6 +5,7 @@ import _flatten from 'lodash/flatten';
 import _compact from 'lodash/compact';
 import _uniqBy from 'lodash/uniqBy';
 import _isEmpty from 'lodash/isEmpty';
+import _orderBy from 'lodash/orderBy';
 import assert from 'ember-cli-mirage/assert';
 
 /**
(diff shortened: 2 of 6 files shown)


Oreoz:
I've just added basic filtering support (id and first level attributes). Not quite sure how I'd go about providing the user means to attach filtering functions to specific filter queryParams like it was requested in #1419.

samselikoff:
@Oreoz Thanks so much – and thanks for the PR! 😄

I'm going to mark this as a feature request and hold off the moment... This expands the problem space that JSONAPISerializer covers quite a bit, and because this is a less rigidly defined part of JSON:API I know people do all sorts of things when it comes to filtering/paginating/sorting. I think we'll ultimately need a bit of a flexible solution here and I want to think of lower-level things we can provide so we don't end up in the same boat as we have with other primitives inside of Mirage.

Right now we're working on extracting core, so I think ultimately this feature will get merged over there, but I'll leave this here & open for now.

Oreoz:
@samselikoff Didn't know you guys already started extracting the core to make it framework-agnostic! That's great new.

I'll be happy to re-work it once you get a better idea of what you're looking for as far as the final API goes. Just ping me here ;)

Also, not quite sure why all the versionned tests exploded on my last commit seeing as everything was good on my end 🤔

samselikoff:
It’s not your code there’s a problem in master

devotox:
What we thinking about this PR?

samselikoff:
Prob need to think more about it before merging as it’s a pretty big change

devotox:
I understand. I have made some changes personally as well to make this a bit cleaner. And also taken it out of the core. Wondering how best to distribute?

https://gist.github.com/devotox/c9d641c2233781b6c6abaedd86be5985

samselikoff:
Don't have an answer right now, I just need to focus on some other things to get miragejs.com officially launched. I would say continue to use + play with it, see if it works out well, and think about what a general API for Mirage would look like. i.e. what API extensions would you need in Mirage to facilitate this feature.

samselikoff:
Going to close this for now because it needs some more thought. I'll keep a reference to this PR in the corresponding issue over in Discuss: https://github.com/miragejs/discuss/issues/28

Thanks for your work on this!

PR #1633, opened 2019-05-23
Depend on @miragejs/server

asantos00:
TLDR: We want to extract the *core* of mirage to a separate package (https://github.com/miragejs/server)

This PR starts to depend on that package for the core parts of it.

**Next steps**
- Manage to generate documentation from `@miragejs/server` and show it on documentation website


samselikoff:
We can’t import from ember-cli-Mirage... the test needs to verify then WHEN an ember developer configures ‘ember-inflector’, THEN Mirage’s server respects those new rules. The test code shouldn’t change here.

asantos00:
Oooh, so you mean we'll have to check if user changed ember-inflector and mimic that on server side?

samselikoff:
Yep exactly

samselikoff:
Don't we need [this logic](https://github.com/miragejs/server/blob/6b9b4415179e8359b20f872f8988a8bc93f1bdfe/lib/ember-data.js) somewhere in here?

asantos00:
That logic was already there, we haven't migrated that file to server

asantos00:
Managed to get it down to 3 failing tests. At the moment the problem is that when Server is instatiated directly on tests, we have no way of discovering the models (as we do on startMirage).

Still have to figure out a way of doing this

PR #1651, opened 2019-06-11
Fix issue with segaults during Travis builds

samselikoff:
Full description of issue: https://github.com/ember-cli/ember-try/issues/360#issuecomment-500969714


scalvert:
👍

PR #1689, opened 2019-07-13
Fix bug - Old testing link on Factories breaks documentation below

ianknauer:
Everything after/below this link is missing on the live page. This should fix the issue.


samselikoff:
Thank you!

PR #1794, opened 2019-10-20
Fix a mistake in Factories docs

dmzza:
I assume you meant to write `server.create('user')` unless there is actually a shorthand version.


samselikoff:
Yep, good catch! Thanks :)

Issue #1427, opened 2018-11-21
Support for loading other scenarios in ember --serve


samselikoff:
![image](https://user-images.githubusercontent.com/2922250/48871536-245fff00-edb3-11e8-9f1c-e0ab6a3d0a82.png)

---

2 upvotes from Canny

teejayhh:
I am building an addon inhouse atm which does exactly that. Its fairly straightforward to implement it yourself. Without much hacking you can modify the default scenario to load other scenarios. You can create a scenario manager  for instance, which either uses a url parameter or local storage to retrieve a scenario name. In the default scenario you can use the `this` or `server._config.scenarios` var to get a list of all available scenarios inside of your scenario folder. Now what's left is to match up the scenario from the list with your url param value or whatever. This approach has been working out pretty well for us. Give it a try.

allthesignals:
The use case for us is that we need to user test our apps under specific user scenarios. For example, user has xyz assigned projects and must complete abc tasks. I'll give @teejayhh's solution a try!

samselikoff:
Yeah I've come to believe this feature isn't really something Mirage needs to have a first-class API to support. It's basically a subset of "run different code under certain conditions" and those conditions could be build-time, run-time or something else.

I think a future API that might make sense for Mirage is a sort of "reset". I mean today you can do this with `server.db.emptyData()`, and then you can run a scenario through it. So yeah I think I'll be holding off on an official solution/recommendation here for a bit.

samselikoff:
Some examples on the new site on how people are solving this would be great though!

allthesignals:
@samselikoff hmmm that surprises me a little bit because the "scenario" concept in Mirage seems first-class through its usage in the folder hierarchy: a file entry in "mirage/scenarios/" implies the ability to add more "scenarios", as you would factories or fixtures or whatever else. On instinct I went around looking for a way to switch out the scenarios environmentally or through some other means.

I do think updating the docs with some examples will help clear up any confusion—I'm happy to look into that!—but I'm not sure it makes sense to dismiss this just yet.

Thanks!

samselikoff:
The other thing is, in `miragejs` we're calling them `seeds()` instead of `scenarios: { default() }` for precisely this reason. (No other scenario support was ever implemented.)

teejayhh:
I feel like it would not take too much to expose a simple api. For the project I was working I needed 4 key peaces. 
1. Make scenarios useful, other then a function which creates seed data
[...]

PR #1802, opened 2019-10-28
Add documention clarifying use of scenarios (#1427)

allthesignals:
This PR adds some documentation about scenarios and ways to use them more effectively.

### DIFF
--- a/tests/dummy/app/pods/docs/advanced/switching-between-scenarios/template.md
+++ b/tests/dummy/app/pods/docs/advanced/switching-between-scenarios/template.md
@@ -1,0 +1,77 @@
+# Switching between scenarios
+
+Mirage adds a `/scenarios` directory to your project, where you'll find a `default.js` file.
+The function in this file will run in development mode and act as your "development seeds".
+In testing it's ignored, so that your Mirage server starts out with a clean database. You
+can then use each test to seed Mirage as needed for that test. In this way, each test acts
+as its own Mirage scenario.
+
+Originally Mirage was going to add first-class support for development-time scenario switching,
+but currently only the `scenarios/default.js` file is used by Mirage. However, you can use various
+strategies to configure the behavior of your default scenario, to support multiple development
+server states or to facilitate things like running [user acceptance testing](https://en.wikipedia.org/wiki/Acceptance_testing#User_acceptance_testing).
+For example, you might want to test whether a user can complete a set of tasks specific to unique circumstances defined in a scenario.
+
+One strategy you could take is to configure your default scenario file to honor options passed
+to the runtime environment. For example, say you wanted to be able to run a scenario file like this:
+
+```bash
+MIRAGE_SCENARIO=some-scenario ember s
+```
+
+To affect the behavior of `scenarios/default.js`, add the `MIRAGE_SCENARIO` ENV variable to your `environment/config.js`:
+
+```js
+const { MIRAGE_SCENARIO } = process.env;
+
+module.exports = function(environment) {
+  const ENV = {
+    // ...other stuff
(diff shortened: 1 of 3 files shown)


tests/dummy/app/pods/docs/advanced/switching-between-scenarios/template.md
```diff
@@ -0,0 +1,66 @@
...
+of your application using Mirage data. For example, you might test whether a user 
+can complete a set of tasks specific to unique circumstances defined in a scenario.
+
+Currently, scenarios are not a first-class API: Mirage only uses the default scenario
+in your development environment.
+
+That said, you can still configure your default scenario file to honor options passed
+to the runtime environment. For example, if you created a scenario file called `some-scenario.js`
+and wanted to load that scenario based on an environment variable called `MIRAGE_SCENARIO`, you
+could add the following to `scenarios/default.js`:
```
samselikoff: Requested changes for first three paragraphs: generalize the motivation a bit (not just user acceptance testing), and open with the bit about not being a first-class API. Something like:

> Mirage adds a `/scenarios` directory to your project, where you'll find a `default.js` file. The function in this file will run in development mode and act as your "development seeds". In testing it's ignored, so that your Mirage server starts out with a clean database. You can then use each test to seed Mirage as needed for that test. In this way, each test acts as its own Mirage scenario.

> Originally Mirage was going to add first-class support for development-time scenario switching, but right only the the `scenarios/default.js` file is used by Mirage. However, you can use various strategies to configure the behavior of your default scenario, to support multiple development server states or to facilitate things like running [user acceptance testing](https://en.wikipedia.org/wiki/Acceptance_testing#User_acceptance_testing). For example, you might want to test whether a user can complete a set of tasks specific to unique circumstances defined in a scenario.

> One strategy you could take is to configure your default scenario file to honor options passed to the runtime environment. For example...

tests/dummy/app/pods/docs/advanced/switching-between-scenarios/template.md
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  return ENV;
+}
+```
+
+Finally, you could switch between scenario files on the command line:
+
+```bash
+MIRAGE_SCENARIO=some-scenario ember s
```
samselikoff: Let's move this part first, so as we're explaining the reader knows the starting point. Let's also avoid the private `server._config`.

---

One strategy would be to use ENV vars to switch scenarios. For example, say you wanted to be able to run a scenario file like this:

```
MIRAGE_SCENARIO=administrator ember s
```

To use this `MIRAGE_SCENARIO` ENV variable to alter the behavior of `scenarios/default.js`, add it to your `environment/config.js`:

```js
const { MIRAGE_SCENARIO } = process.env;
module.exports = function(environment) {
  const ENV = {
    // ...other stuff

    MIRAGE_SCENARIO,

    // ...other stuff
  };

  return ENV;
}
```

then access this new config variable in your `scenarios/default.js` file:

```js
// scenarios/default.js
import ENV from '../../config/environment';
import visitor from './visitor';
import subscriber from './subscriber';
import administrator from './administrator';

const scenarios = {
  visitor,
  subscriber,
  administrator
}

const activeScenario = ENV.MIRAGE_SCENARIO || 'visitor';

  const scenario = scenarios[activeScenario];

  if (scenario) {
    scenario(server);
  }

  // plus whatever default scenario code you want
}
```

allthesignals:
@samselikoff pushed more changes. Please let me know if I can improve anything else.

samselikoff:
Looks great – thank you Matt!

PR #1818, opened 2019-11-16
Add "Mirage logging" checkbox to QUnit user interface

### LINKED ISSUE #1783: Add "Mirage logging" checkbox to QUnit user interface
Bildschirmfoto 2019-10-09 um 14 43 31

QUnit has an API for adding checkboxes to it:

```js
QUnit.config.urlConfig.push({
  id: 'mirageLogging',
  label: 'Mirage logging',
});
```

The value of this checkbox is later available as:

```js
server.logging = QUnit.config.mirageLogging;
```

It would be great to have this be added by default when using `ember-cli-mirage` in combination with QUnit.

Happy to work on this if it is something that is considered useful :)

Turbo87:
Resolves #1783 

Unfortunately, I had to add `after: ember-qunit` to the `package.json` to ensure that `window.QUnit` is actually defined already by the time that we try to register the QUnit option. Registering it later when the server is created is unfortunately not possible, since the options have to be defined already by the time that QUnit is running the tests.

/cc @samselikoff

### DIFF
--- a/addon/server.js
+++ b/addon/server.js
@@ -16,6 +16,11 @@ export default class EmberServer extends Server {
     }
 
     super(options);
+
+
+    if (typeof location !== 'undefined' && location.search.indexOf('mirageLogging') !== -1) {
+      this.logging = true;
+    }
   }
 
 }
--- a/vendor/add-qunit-option.js
+++ b/vendor/add-qunit-option.js
@@ -1,0 +1,8 @@
+(function() {
+  if (typeof QUnit !== 'undefined') {
+    QUnit.config.urlConfig.push({
+      id: 'mirageLogging',
+      label: 'Mirage logging',
+    });
+  }
+})();
--- a/index.js
+++ b/index.js
@@ -48,22 +48,20 @@ module.exports = {
   },
 
   treeFor(name) {
-    let tree;
     let shouldIncludeFiles = this._shouldIncludeFiles();
+    if (shouldIncludeFiles) {
+      return this._super.treeFor.apply(this, arguments);
+    }
 
-    if (!shouldIncludeFiles && name === 'app') {
+    if (name === 'app') {
       // Include a noop initializer, even if Mirage is excluded from the build
-      tree = writeFile('initializers/ember-cli-mirage.js', `
+      return writeFile('initializers/ember-cli-mirage.js', `
         export default {
           name: 'ember-cli-mirage',
           initialize() {}
         };
       `);
-    } else if (shouldIncludeFiles) {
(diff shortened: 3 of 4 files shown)


addon/server.js
```diff
@@ -16,6 +16,11 @@ export default class EmberServer extends Server {
     }
 
     super(options);
+
+
+    if (window.QUnit && window.QUnit.config.mirageLogging) {
```
rwjblue: No need to require `window.QUnit` here, something like this would do:

```suggestion
    if (typeof QUnit !== 'undefined' && QUnit.config.mirageLogging) {
```

rwjblue: Also, FWIW, if you searched for a query param instead of QUnit.config.mirageLogging this would also work for ember-mocha users.

Something like:

```suggestion
    if (typeof location !== 'undefined' && location.search.indexOf('mirageLogging') > -1) {
```

vendor/add-qunit-option.js
```diff
@@ -0,0 +1,8 @@
+(function() {
+  if (window.QUnit) {
```
rwjblue:
```suggestion
  if (typeof QUnit !== 'undefined') {
```

addon/server.js
Turbo87: good idea, thanks 👍

rwjblue:
Looks like CI is failing due to some sort of sourcemap concat issue:

```
Build Error (SourceMapConcat)
	ENOENT: no such file or directory, open '/home/travis/build/miragejs/ember-cli-mirage/test-projects/02-app-that-excludes-mirage/tmp/source_map_concat-input_base_path-WHbgHBql.tmp/vendor/add-qunit-option.js'
```

rwjblue:
My guess is that this is because the `treeFor` override results in no files in the `vendor` tree if `_shouldIncludeFiles` returns false, assuming that is the behavior we want I think that you'd need to wrap the `app.import` in a conditional.

Turbo87:
@rwjblue yep, that was it. I've instead changed the `treeFor()` implementation to pass on the original `vendor` tree, since it wasn't used for anything else and needs to be explicitly imported from anyway.

rwjblue:
Gotcha, makes sense!

samselikoff:
This is awesome! Nice work <3

Alonski:
@samselikoff @Turbo87 Can we bet a new release cut so that this can be used please? :)

samselikoff:
Here ya go! https://github.com/miragejs/ember-cli-mirage/releases/tag/v1.1.6

Alonski:
Thanks @samselikoff !

PR #1833, opened 2019-11-24
Fix broken intra-documentation links

backspace:
I was able to access the desired pages via the sidebar, but might as well remove that moment of confusion. 💞


samselikoff:
🧡 TYVM!

PR #1912, opened 2020-03-26
Update Customizing the inflector docs

### LINKED ISSUE #1911: Inflector initializer file name matters
# Bug or question

`ember-cli-mirage` uses initializer to do some setup, but if custom inflector rules initializer is starting from any letter after `E`, it won't be counted. so I think docs should ask users to put `  before: 'ember-cli-mirage',` in their inflector initializer. In this case file name won't matter.

What do you think?

yratanov:
fixes #1911


samselikoff:
Thanks! And woops - actually didn't mean to merge this, wanted to ask why you made it into `diff`?

samselikoff:
I think we should keep it `js` and just add a sentence below that says something like "Make sure to use `before: "ember-cli-mirage"` so this runs before Mirage initializes itself."

yratanov:
@samselikoff I'll update

PR #1917, opened 2020-03-28
Update Customizing the inflector docs [2]

yratanov:
Apply feedback from #1912


samselikoff:
Awesome, thanks again!

PR #1948, opened 2020-04-25
Use native setters in documentation examples

allthesignals:
I think we should use native setter syntax in documentation examples. Is there a reason we should use .set? Backwards compatibility?


Turbo87:
@allthesignals I don't think that this actually works properly. It works fine if you set these values before rendering a component, but setting it again afterwards won't invalidate the render, so AFAIK for now we still need the `set()` calls on the test context object.

PR #1954, opened 2020-05-04
test-support to addon-test-support

ChangJoo-Park:
Fix typo from acceptance-test


ChangJoo-Park:
It is wrong PR from vscode guide :(

PR #1971, opened 2020-05-29
Replace getWithDefault (Fixes #1970)

### LINKED ISSUE #1970: getWithDefault deprecated on Ember canary
Our Ember canary tests are failing with:
```
Error: Using getWithDefault has been deprecated. Instead, consider using Ember get and explicitly checking for undefined.
```

See https://github.com/emberjs/rfcs/pull/554

Looks like `getWithDefault` is used here in `startMirage`:
https://github.com/miragejs/ember-cli-mirage/blob/dcef77adced352d43d314ab484e47075018addb7/addon/start-mirage.js#L33

I can open a PR with a proposed fix.

gitKrystan:
`getWithDefault` is deprecated by https://github.com/emberjs/rfcs/pull/554, causing failures with Ember Canary. See #1970

### DIFF
--- a/addon/start-mirage.js
+++ b/addon/start-mirage.js
@@ -1,4 +1,3 @@
-import { getWithDefault } from '@ember/object';
 import readModules from './utils/read-modules';
 import Server from './server';
 import { singularize, pluralize } from 'ember-inflector';
@@ -30,10 +29,12 @@ export default function startMirage(owner, { env, baseConfig, testConfig } = {})
   }
 
   let environment = env.environment;
-  let discoverEmberDataModels = getWithDefault(env['ember-cli-mirage'] || {}, 'discoverEmberDataModels', true);
+  let mirageEnvironment = env['ember-cli-mirage'] || {};
+  let discoverEmberDataModels = mirageEnvironment.discoverEmberDataModels;
+  if (discoverEmberDataModels === undefined) { discoverEmberDataModels = true; }
   let modules = readModules(env.modulePrefix);
   let options = Object.assign(modules, {environment, baseConfig, testConfig, discoverEmberDataModels});
-  options.trackRequests = env['ember-cli-mirage'].trackRequests;
+  options.trackRequests = mirageEnvironment.trackRequests;
   options.inflector = { singularize, pluralize };
 
   return new Server(options);


samselikoff:
Amazing – thank you!

gitKrystan:
@samselikoff Regarding the test failure, I believe it's unrelated. There is a version of moment with two copies of the `en-sg.js` file: https://github.com/moment/moment/commit/36a817d94eab0f070bbcc7d8431631ee71400b89#diff-fb216d9e8791e63c8d12bdc420956839
We fixed this issue in our app by resolving to a newer (fixed) version of moment. If you know how you want to handle it, I'm happy to open another PR to fix.

samselikoff:
Yes – sorry I haven't gotten to this yet. It looks like a problem with the floating dependencies test job. Usually this means there's a change to a transitive dependency somewhere that's broken something.

In the past I've fixed this by manually doing a binomial search on the `yarn.lock` file... it's kind of ridiculous. Basically, you should be able to reproduce the failure if you run the suite locally with `yarn install --no-lockfile`, and if you do, the problem is somewhere in there. The next step is to literally delete half of the lock file, run `yarn install`, and see if you have the error. If you do, the problem is in one of the remaining lines; if not, it was in one of the lines you deleted.

Continue trimming down by half each time until you identify the problematic line of the lockfile. At that point you should have identified the problematic dependency and we can decide whether to just lock it down using Yarn resolutions, or try to help fix the underlying bug.

LMK if that makes sense, and if you get around to it that would be extremely helpful! If not I'll probably get to it this weekend. After we solve it, we can open a PR (separately) with an updated `yarn.lockfile`, including whatever fix we applied, at which point we can re-run this PR and it should pass.

chancancode:
@samselikoff (or anyone that has time) TL;DR – https://github.com/jasonmit/ember-cli-moment-shim/issues/183

gitKrystan:
These tests should pass with https://github.com/miragejs/ember-cli-mirage/pull/1977

chancancode:
Rebased against mater to kick off another CI run

chancancode:
@samselikoff mind cutting a release when you get a chance? not sure what your release process is here

samselikoff:
Absolutely - thanks so much for taking care of these over the weekend!

chancancode:
No problem! And thanks for the release. I left most of the dep upgrades because they are either more significant or because they require dropping the EOL node versions here, which requires bumping major.

samselikoff:
@chancancode gotcha... do all dep upgrades that EOL node require e-c-mirage to bump major? I think this one might have already gotten in: https://github.com/miragejs/ember-cli-mirage/pull/1897

But if e-c-mirage is not using broccoli-funnel in a way that breaks on older nodes, it's technically not breaking for us, is it?

chancancode:
I think if you are running node 8, `npm install` (or `yarn`) will fail, because it will fail to find a version of `broccoli-funnel` that matches all the constraints – within the version range e-c-mirage asked for and supports node 8 according to their `package.json`, so I think it's generally best practice to consider dropping node versions a breaking change. If this breaks people code, we can revert the bump and re-release 1.1.8. cc @rwjblue

samselikoff:
I see. Happy to do that, def. don't mind reverting + bumping versions in the future.

PR #1993, opened 2020-06-17
Implement a push mirage into store utility

cah-briangantzler:
/**
 * This utility pushes the mirage store into the ember store
 *
 * If you call this function passing no parameters, it will push the entire mirage store into ember store
 *
 * You may also call this utility with an array of resource names. ['accounts', 'accountTypes']. These must
 * be the resource names camelized.
 *
 * You may also call it with a hash. This hash will have the resource names as the keys and the values will either
 * be true (push all models for that resource) or a where hash that will be applied to the that resource.
 * {
 *   accounts: true,
 *   accountTypes: { state: 'OH' }
 * }
 *
 * @param config   undefined means push all resources
 *                 An array will push only the resources in the array
 *                 hash in the form of resource=value    Value may be true, all models, or a where hash
 */

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,2 +1,3 @@
 export { default } from 'miragejs';
 export * from 'miragejs';
+export { pushMirageIntoStore, pushMirageRecordsIntoStore } from 'ember-cli-mirage/utils/push-mirage-into-store';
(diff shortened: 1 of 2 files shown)


cah-briangantzler:
Appears as if @ember/test-helpers is not available. 

I put the pushToStore in the utils, but it should have been in addon-test-support. 

Going to close this and rework

PR #2001, opened 2020-06-28
Provided a pushMirageIntoStore import for tests and one for outside tests

cah-briangantzler:
Exposes a method for the consumer to push mirage's database in the ember data store

If you call this method with no parameters, it will push the entire mirage store into ember store

You may also call this method with an array of resource names. ['accounts', 'accountTypes']. These must
be the resource names camelized.

You may also call it with a hash. This hash will have the resource names as the keys and the values will either
be true (push all records for that resource) or a where hash or function that will be applied to the that resource.
{
  accounts: true,
  accountTypes: { state: 'OH' }
  users: item => item.name.includes("Joe")
}

@param server   The mirage server
@param store    the store you wish mirage to push to
@param config   undefined means push all resources
                An array will push only the resources in the array
                hash in the form of resource=value    Value may be true, all records, or a where hash or function

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,2 +1,4 @@
 import setupMirage from './setup-mirage';
 export { setupMirage };
+
+export { pushMirageIntoStore } from "./push-mirage-into-store";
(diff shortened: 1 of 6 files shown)


cah-briangantzler:
Rebased and re-pushed this.

cah-briangantzler:
The linter fails on 

 Native JS classes should be used instead of classic classes  ember/no-classic-classes

Its in my tests. But its the way that MirageJS Documents doing it. Pretty sure you have to use extend because of the way the base class works.

Should we lighten this lint or should I look into turning it off for this one file?

PR #2073, opened 2020-11-03
Fix invalid use of ES modules

ef4:
The ES module spec doesn't support "optional" exports. If you try to import a name from a module and that name is not exported, it's a SyntaxError.

Mirage tries to import `default`, `testConfig`, and `makeServer` from the user's config, but they are treated as optional.

The solution is to import the module's whole namespace (`import * as...`) and then inspect the namespace object.


ef4:
(I'm adding an audit tool to Embroider, and any app that uses ember-cli-mirage will see a problem in their audit results due to this issue.)

samselikoff:
Makes sense - thanks for the PR Ed!

PR #2093, opened 2020-12-08
Update mirage logging option to work with qunit5

cah-briangantzler:
Fixes https://github.com/miragejs/ember-cli-mirage/issues/2091

There were no tests for the original adding of the option, I dont see how to add tests as it not a test of the addon but how the addon works with Qunit.

- Moved the adding of the option from vendor/add-qunit-option to test-support/qunit-configuration same as qunit5 does
- Moved import of code from index.js to test-support/index.js.  This means if you never import setupMirage you will never see the option added, which kinda makes sense. This also means you wont see it if you run the tests in this repo, since none of the tests import setupMirage

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,2 +1,9 @@
 import setupMirage from './setup-mirage';
 export { setupMirage };
+
+import { dependencySatisfies } from '@embroider/macros';
+
+if (dependencySatisfies('ember-qunit', '*')) {
+  window.QUnit.config.urlConfig.push({ id: 'mirageLogging', label: 'Mirage logging', });
+}
+
--- a/index.js
+++ b/index.js
@@ -43,8 +43,6 @@ module.exports = {
     } else {
       this.mirageDirectory = path.join(this.app.project.root, '/mirage');
     }
-
-    this.import('vendor/add-qunit-option.js', { type: 'test' });
   },
 
   blueprintsPath() {
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "prepare": "./scripts/link.sh"
   },
   "dependencies": {
+    "@embroider/macros": "^0.33.0",
     "broccoli-file-creator": "^2.1.1",
     "broccoli-funnel": "^3.0.3",
     "broccoli-merge-trees": "^4.2.0",
@@ -67,7 +68,6 @@
     "ember-cli-moment-shim": "^3.8.0",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-uglify": "^3.0.0",
-    "ember-compatibility-helpers": "^1.2.1",
     "ember-composable-helpers": "^4.1.2",
     "ember-data": "~3.14.0",
     "ember-disable-prototype-extensions": "^1.1.3",
--- a/tests/acceptance/home-test.js
+++ b/tests/acceptance/home-test.js
@@ -1,16 +1,13 @@
 import { module, test } from "qunit";
 import { visit, currentURL } from "@ember/test-helpers";
 import { setupApplicationTest } from "ember-qunit";
-import {
-  // General functions for checking against Ember version
-  gte
-} from 'ember-compatibility-helpers';
+import { dependencySatisfies } from '@embroider/macros';
 
 module("Acceptance | home", function(hooks) {
   setupApplicationTest(hooks);
 
   //  Ember Addon docs will fail for sources less then 3.16.0
-  if (gte('3.16.0')) {
+  if (dependencySatisfies('ember-source', '^3.16.0')) {
     test("the homepage renders without error", async function(assert) {
       await visit("/");
 
--- a/tests/index.html
+++ b/tests/index.html
@@ -21,6 +21,13 @@
     {{content-for "body"}}
     {{content-for "test-body"}}
 
+    <div id="qunit"></div>
+    <div id="qunit-fixture">
+      <div id="ember-testing-container">
+        <div id="ember-testing"></div>
+      </div>
+    </div>
+
     <script src="/testem.js" integrity=""></script>
     <script src="{{rootURL}}assets/vendor.js"></script>
     <script src="{{rootURL}}assets/test-support.js"></script>
--- a/vendor/add-qunit-option.js
+++ b/vendor/add-qunit-option.js
@@ -1,8 +1,0 @@
-(function() {
-  if (typeof QUnit !== 'undefined') {
-    QUnit.config.urlConfig.push({
-      id: 'mirageLogging',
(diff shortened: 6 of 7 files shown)


addon-test-support/qunit-configuration.js
```diff
@@ -0,0 +1,3 @@
+import * as QUnit from 'qunit';
```
Turbo87, line 1: what happens here if the app is using Mocha instead of QUnit?

cah-briangantzler: I was hoping it would be undefined, just tried it, it fails to link.

So I either have to find a way to check if qunit it installed and then conditionally import.

While it would be nice to do it automatically, Qunit5 is already making people edit the index.html so we could document adding the following lines to test-helper.js for qunit5

```javascript
import QUnit from 'qunit';

QUnit.config.urlConfig.push({ id: 'mirageLogging', label: 'Mirage logging', });
```

Does mocha in some way provide for adding this option to its UI? Mirage has code to add it for qunit, but I dont think there is code for mocha. Should the add-qunit-option.js code be removed from mirage to make it test framework independant?

Turbo87: > Does mocha in some way provide for adding this option to its UI? 

nope

> Should the add-qunit-option.js code be removed from mirage to make it test framework independant?

I don't think we should remove it. It brings quite a bit of value to QUnit users, and just because Mocha does not support it shouldn't mean that QUnit users also can't enjoy it.

> So I have to find a way to check if qunit it installed and then conditionally import.

yeah, I think that would be the way to go.

I think something like https://github.com/simplabs/qunit-dom/blob/a0aa749cf9d561ea5cf331fb8c50271d666b68e7/tests/test-helper.js#L7-L13 might be able to solve this issue

cah-briangantzler: The problem is the import statement not the setup code. In your example the import for quint is outside the IF. If qunit is not defined the import fails and will never get to the if.  

That being the case it would be easier it ember-qunit used the dependencySatisfies on mirage and did the add, since we know qunit is there.

The only solution I see is the user opting into this by documenting adding a couple lines. OR.... we could  put the new lines in the mirage/config.js as part of the template, and if you are using mocha document that you have to remove them.

What we need is a conditional import, Im just not good enough at modules to know how to do that. Can google some.

Turbo87: sorry, I wasn't explaining properly. I was wondering if we could use something roughly like:

```js
if (dependencySatisfies('ember-qunit', '*')) {
  const QUnit = importSync('ember-qunit');
  doSomething(QUnit);
}
```

cah-briangantzler: Originally the code was in ember-cli-build.js, when I looked there QUNIT was no longer a global. Thats where the need to import came from. To fix the problem I moved the code from ember-cli-build.js to test-support/index.js.   After more testing it turns out it is a global again there (I would assume making qunit a peer-dependancy caused a load order problem and it was no longer a global in ember-cli-build.js but was when test-support/index.js got around to executing). So I could have just gone with the original code 
```javascript
if (typeof QUnit !== 'undefined') {
    QUnit.config.urlConfig.push({
      id: 'mirageLogging',
      label: 'Mirage logging',
    });
  }
```

Instead went with the macros from embroider for a more specific solution 
```javascript
import { dependencySatisfies } from '@embroider/macros';

if (dependencySatisfies('ember-qunit', '*')) {
  window.QUnit.config.urlConfig.push({ id: 'mirageLogging', label: 'Mirage logging', });
}
```

Used window because this is now in linted code. 


Good to know about importSync for future, but ended up not needing it here. 

Was also using ember-compatibility-helpers but embroider macros did the same thing, so changed that over as well.

addon-test-support/index.js
```diff
@@ -1,2 +1,9 @@
 import setupMirage from './setup-mirage';
 export { setupMirage };
+
+import { dependencySatisfies } from '@embroider/macros';
+
+if (dependencySatisfies('ember-qunit', '*')) {
+  window.QUnit.config.urlConfig.push({ id: 'mirageLogging', label: 'Mirage logging', });
```
Turbo87, line 7: are you sure that this works with ember-qunit v5? I thought we now had to explicitly import it and it was no longer a global on `window`.

/cc @rwjblue

cah-briangantzler: Originally the code was in ember-cli-build.js, when I looked there QUNIT was no longer a global. Thats where the need to import came from. To fix the problem I moved the code from ember-cli-build.js to test-support/index.js. After more testing it turns out it is a global again there (I would assume making qunit a peer-dependancy caused a load order problem and it was no longer a global in ember-cli-build.js but was when test-support/index.js got around to executing). So I could have just gone with the original code

cah-briangantzler: I have tested with by temporarily upgrading the tests to ember-qunit v5 and it does work.  I did not leave it that way because I would also need to upgrade the test projects because of the way they share the node-modules directory.

cah-briangantzler:
What next to get this merged in?

Turbo87 (approved):
sorry, forgot about it 😱

Turbo87:
@cah-briangantzler I just tried this out on an app that is still using ember-qunit v4 and the checkbox is unfortunately not visible anymore. can you check if we can make this work again too?

cah-briangantzler:
I updated my application to use ember-cli-mirage 2.0.0, it is still using ember-qunit ^4.6.0 and the checkbox is visible for me.

I am unable to update my application to qunit v5 as I use ember-cli-page-object and that is NOT ember-qunit 5 compatible at this time. 

I did temporarily update (but did not commit) the ember-cli-mirage tests to ember-qunit 5 and the checkbox was visible and working.

cah-briangantzler:
@Turbo87 Were you able to get this to work? There is a nuance I mentioned earlier, previously just including the addon added the check box to the qunit testing, even if you never imported setupMirage (and presumably created a mirage server). Where I moved it to requires you to at least import something (mostly likely setupMirage) from 'ember-cli-mirage'. That was not on purpose, it is a by product of where I moved it, but kinda makes sense, if you never import and create a server, having the checkbox does not really do anything. 

Are you creating the mirage server on your own without importing setupMirage?

Turbo87:
I think I found the issue. We were importing `setupMirage()` like this:

```js
import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
```

but since the `urlConfig` change happens in the `index.js` file it was never triggered. I've now changed our imports to this:

```js
import { setupMirage } from 'ember-cli-mirage/test-support';
```

and now it works... 😅

cah-briangantzler:
Great to hear

PR #2101, opened 2020-12-26
Widen `ember-get-config` dependency requirement

Turbo87:
This addon works fine with 0.2.x and requiring 0.3.0 can cause dependency version conflicts, so let's widen the requirement so that users can choose a version that does not cause them any conflicts.

/cc @samselikoff


samselikoff:
Looks good!

PR #2159, opened 2021-04-02
Allow for requirejs.entries to be passed from the host application

bgantzler:
Updated the new config with make server to allow the optional passing of requirejs.entries for embroider compatibility.    If this solution becomes the permanent solution, then when embroider becomes common place this would no longer be an optional param 

Fixes issue https://github.com/miragejs/ember-cli-mirage/issues/2147


cah-briangantzler:
Closing in favor of a future PR. This was more than needed

PR #2160, opened 2021-04-05
Updated hasEmberData regex. refactored the way requirejs.entries to derived

bgantzler:
Updated the hasEmberData method to check against 2 regex. One is for backward compatibility, one is for the embroider compatibility

requirejs.entries was fetched in multiple places. Tried to limit where it was fetched and passed it around instead of re-fetching.  While not needed for embroider compatibility at this time, thought it was better to keep in so if in the future we can alter what is passed around and that might make it more flexible for changes in the way those things are tracked.

Fixes https://github.com/miragejs/ember-cli-mirage/issues/2147


addon/ember-data.js
```diff
@@ -21,26 +21,27 @@ let DsSerializers, Serializers
...
 
   DsModels = {};
 
-  if (!hasEmberData) {
+  if (!hasEmberData(options)) {
     return DsModels;
   }
 
-  Object.keys(moduleMap)
+  Object.keys(options.moduleMap)
```
Turbo87, line 44: what happens if `options` is `undefined`? I guess requiring it might be a breaking change since this is an exported function?

Turbo87:
@bgantzler it looks like this PR does two separate things:

- adjust the ED regex
- pass the requirejs map around

they are certainly related, but they don't seem directly dependent on each other. since a review comment on one could potentially block the other I'm wondering if it would be best to split this into two PRs :)

addon/ember-data.js
cah-briangantzler: I guess I didnt notice it was exported. Its only ever called within the same file. I guess it was exported in case someone outside wanted to call it, but I can imagine doing that. Probably shouldnt have been exported

cah-briangantzler:
The initial research said the ED regex wouldnt work, so I did all the rest to try and isolate requirejs.entries.    Then it turned out the regex did work.  So thought I would leave all the work done since it might be valuable in the future it a change was needed.


So... Yea will just make a PR for the regex change and leave it at that.

cah-briangantzler:
https://github.com/miragejs/ember-cli-mirage/pull/2177 implements the regex update.      Closing as rest of code is not needed at this time

PR #2167, opened 2021-04-18
Public API to query if files are included in build

### LINKED ISSUE #2166: Public API for addons to determine if mirage files are included in the buld
Some addons export code from `addon/` that imports code from mirage, e.g. a route handler function that returns a `Response`, or mirage model definitions that need to import `Model` to subclass it, so applications can use them in their mirage setup. A couple of examples are [ember-cli-graphql](https://github.com/kloeckner-i/ember-cli-mirage-graphql) and [ember-file-upload](https://github.com/adopted-ember-addons/ember-file-upload/blob/556075f97c2ceada0ac11c3bcfbd734b84bd6245/addon/mirage/index.js). Currently there is not a very safe way to do this because the addon has no good way of knowing if mirage's files are included in the build according to the mirage config or not.

Before embroider, this has kind of flown below the radar because in cases where the mirage code wasn't included in the build, the application obviously wouldn't be starting the mirage server, so none of the routes or models or anything would be required at runtime (even if they are included in the bundle), meaning the code supplied by the addon that would have imports pointing to not-present mirage code would just be inert and never executed, and not cause any problems other than inflating the bundle size. But embroider does more static analysis, and fails if an import can't be resolved, so this now causes a problem, and any code that imports mirage code must be excluded from the build whenever mirage is excluded from the build, whether it will ultimately be executed at runtime or not.

`ember-cli-graphql` solves this by [replicating mirage's configuration logic](https://github.com/kloeckner-i/ember-cli-mirage-graphql/blob/9e2b7c0240cf753e2cfe3a42c49bfae6653c0c4e/index.js#L37-L49), but this doesn't seem like a great solution to apply generally. So, my proposal is to provide a Node.js API that addons can use at build-time to ask `ember-cli-mirage` if its files will be included in the build.

I can think of several ways to do this:
1. Make [_shouldIncludeFiles()](https://github.com/miragejs/ember-cli-mirage/blob/a9b38e7740f9a496678b68c1d849df224e6136fa/index.js#L108-L122) public so addons can just `require('ember-cli-mirage').shouldIncludeFiles()`. However, then addons have to either make `ember-cli-mirage` a runtime dependency which seems less than ideal, or handle the case where `ember-cli-mirage` isn't installed in the consuming application, which isn't difficult practically, but just adds extra complexity.
2. Have `ember-cli-mirage` check the config sometime early in the build process and export an environment variable when its files are included in the build that other addons can check. That way if `ember-cli-mirage` is either not installed at all, or installed but not including its files, the environment variable would not be set and addons would know to exclude their mirage-dependent files. This would probably work, but using the environment this way rather than some explicit `ember-cli` mechanism seems kinda kludgy.
3. Have `ember-cli-mirage` call `_shouldIncludeFiles()` in its `included()` hook and write the result back to the `ember-cli-mirage` object in the config (the same place it reads its `enabled` and other values). That way addons could use the `after` key in the `ember-addon` section of `package.json` to ensure their hooks are called after `ember-cli-mirage`'s, and then they can look in the configuration object, just like mirage does, to determine whether to add mirage-dependent files, except there will be a flag explicitly indicating whether to include mirage files or not.

(3) seems like the right approach to me, so I'll be opening a PR proposing that change shortly.

bendemboski:
Implement a public API (using the environment config) for addons to determine if mirage's files are included in the build, so they can decide whether to include any mirage-dependent files.

Fixes #2166

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,10 +28,20 @@ module.exports = {
     }
 
     this.app = app;
-    this.addonConfig = this.app.project.config(app.env)['ember-cli-mirage'] || {};
+
+    // Make sure the mirage config exists, and set the `includedInBuild` key so
+    // addons that provide code that relies on mirage can decide whether to
+    // include their code or not based on whether or not mirage is included in
+    // the build.
+    let config = this.app.project.config(app.env);
+    config['ember-cli-mirage'] = config['ember-cli-mirage'] || {};
+    this.addonConfig = config['ember-cli-mirage'];
+    this.addonConfig.includedInBuild = this._shouldIncludeFiles();
+
     this.addonBuildConfig = this.app.options['ember-cli-mirage'] || {};
 
-    // Call super after initializing config so we can use _shouldIncludeFiles for the node assets
+    // Call super after initializing config so we can check the `includeInBuild`
+    // flag for the node assets
     this._super.included.apply(this, arguments);
 
     if (this.addonBuildConfig.directory) {
@@ -50,8 +60,7 @@ module.exports = {
   },
 
   treeFor(name) {
-    let shouldIncludeFiles = this._shouldIncludeFiles();
-    if (shouldIncludeFiles || name === 'vendor') {
+    if (this.addonConfig.includedInBuild || name === 'vendor') {
       return this._super.treeFor.apply(this, arguments);
     }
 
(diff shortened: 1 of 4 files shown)


bendemboski:
Outstanding issues:
* I have not tested this manually yet -- I will soon, but wanted to get this open for discussion/feedback ASAP
* There are no automated tests for this functionality, although it looks like there are no automated tests for the code inclusion/exclusion at all so maybe that's okay

PR #2171, opened 2021-04-27
build: only include used composable-helpers

MrChocolatine:
This PR configures the add-on [`ember-composable-helpers`](https://github.com/DockYard/ember-composable-helpers) to only include a specific set of helpers when the project is built.

According to your HBS files, you only use the helper `sort-by`:
https://github.com/miragejs/ember-cli-mirage/search?l=handlebars

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -13,6 +13,9 @@ module.exports = function (defaults) {
         { name: "miragejs", sourceDirectory: "lib" },
       ],
     },
+    "ember-composable-helpers": {
+      only: ["sort-by"]
+    },
     addons: {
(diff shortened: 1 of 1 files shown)


cah-briangantzler:
Is this really needed? composable helpers is a dev dependency meaning it really only applies to the  docs. This change would strip the unused helpers from the docs site, but really has no bearing on any app that includes mirage.

Turbo87:
@cah-briangantzler I guess it's not strictly needed, but since it also only affects the docs site, and make the payload a little smaller I'm still going to merge it :)

Turbo87 (approved):

PR #2176, opened 2021-04-29
docs: factory - fix traits example

canrozanes:
The code snippet changed in this PR is described with the following text `Here we define a withComments trait that creates 3 comments for a newly created post:`

Creating a post requires having a factory on the path `mirage/factories/post.js` as opposed to `mirage/factories/user.js`

Also to keep it consistent with the rest of the examples where posts have a property called `title`, as opposed to `name`, I went ahead and updated the `name` property on the factory to be `title`.


Turbo87 (approved):
👍

PR #2177, opened 2021-04-29
Fix regex that checks if ember data is present to allow for difference in namespace

### LINKED ISSUE #2147: Embroider: DBs are not populated when `staticAddonTrees: true`
When I enable `staticAddonTrees: true` in the build it seems to break the mirage database.

Specifically the detection of ember-data models.

With the above option enabled the mirage `server.db` is missing all the regular ember-data models in my app and only has entries for models I defined a mirage factory for.

Asked in `#ec-mirage` on the Ember discord and another user mentioned they had the same issue.

I made a [dummy repo](https://github.com/gilest/embroider-mirage) with a failing test demonstrating the issue. Disabling `staticAddonTrees`  will cause this test to pass.

cah-briangantzler:
Fixes #2147

### DIFF
--- a/addon/utils/ember-data.js
+++ b/addon/utils/ember-data.js
@@ -1,8 +1,10 @@
 /* global requirejs */
 
 function _hasEmberData() {
-  let matchRegex = /^ember-data/i;
-  return !!(Object.keys(requirejs.entries).find(e => !!e.match(matchRegex)));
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
sweet, thanks :)

gilest:
Thanks again @cah-briangantzler 🙇

PR #2184, opened 2021-05-08
Added embroider-safe and embroider-optimized to the test scenarios

cah-briangantzler:


thoov:
@cah-briangantzler in the next release of Embroider we will handle the webpack dependency within the helpers so that dep does not need to be explicitly set. So it might be worth waiting for that

cah-briangantzler:
Good to know. Since the change was just made to require the app to do the webpack, I should have just added webpack as a devDependancy instead of what I did and figured that was the desired direction.

rahulk94:
Oh nice to see this is already raised @cah-briangantzler. If ya upgrade your dependency of the `@embroider/test-setup` to the latest now you should be able update remove the explicit webpack handling as @thoov mentioned earlier 🙂

PR #2215, opened 2021-07-23
Fix the way that the new server creation using mirageJS is determined

cah-briangantzler:
This does a better job at detecting if the default export from the config.js is a routes config or a make server config. This will help with the issues people have been seeing with embroider.

The current named makeServer function is deprecated with instructions to create a default export.

Server configuration docs have been updated to show creating a default export instead of a named makeServer function. 

This should address the needs from the PR https://github.com/miragejs/ember-cli-mirage/pull/2213. That PR should no longer be needed.


timiyay (approved):
Thanks so much for this work @cah-briangantzler

PR #2227, opened 2021-09-08
Bump ember-cli-babel

charlesfries:


cah-briangantzler:
If I understand the package.json correctly, this says that it works with any version 7.5+, if we change this, doesnt it then say that it only works with 7.26.6+.   Is that what we want to say?  Was this PR required for you somewhere to fix something?

charlesfries:
I was just seeing this package listed among many others in a v3.27 deprecation warning caused by `ember-cli-babel`, so I thought I would bump it:

```
Prior to v7.26.6, ember-cli-babel sometimes transpiled imports into the equivalent Ember Global API, potentially triggering this deprecation message indirectly, even when you did not observe these deprecated usages in your code.
```

Do you know if `ember-cli-babel` respects semver? I assumed a minor version bump from 5 to 26 would still be backwards compatible, but I'm not sure.

cah-briangantzler:
As I understand it, if your app is using 7.20, this says I work with 7.5+ so Ill use the 7.20, that would then trigger the deprecation. But if your app is using 7.26.6, this says hey I can use it, and so will use 7.26.6 and not trigger the warning. So really, the app. should be updated not the addon. 

If we change this to 7.26.6+ then if your app is using 7.20 then it will say nope, I cant use it, and bring in a copy of 7.26.6 and use it. Not triggering a warning. So in effect, yes this does fix it by bringing in another copy. But by changing it here we are actually saying we require 7.26.6+ to work like we are relying on a feature that isnt in 7.25- which is not actually true and we could cause another copy that might not have been needed. 

If this is all correct, I would then assume that addons should always specify the minimum requirement they need to work, not the most recent one. 

Again this is just my understanding of how things work and would appreciate anyone correcting me if any of this is wrong. 

All the above is for more generic dependancies,  this particular problem is unique though in that maybe this PR is the correct thing to do. Hoping someone smarter weighs in here.

rahulk94:
Hey team, so I've come across what I initially thought was an issue of `ember-cli-mirage` using an outdated babel version when upgrading our Ember app from 3.24 to 3.28. Logs below (removed the bits from other dependencies to keep things shorter).
```
WARNING: [DEPRECATION] [DEPRECATION] Usage of the Ember Global is deprecated. You should import the Ember module or the specific API instead.

See https://deprecations.emberjs.com/v3.x/#toc_ember-global for details.

Usages of the Ember Global may be caused by an outdated ember-cli-babel dependency. The following steps may help:

* Upgrade the following addons to the latest version:
  * ember-cli-mirage
  ...
### Important ###

In order to avoid repeatedly showing the same deprecation messages, no further deprecation messages will be shown for usages of the Ember Global until ember-cli-babel is upgraded to v7.26.6 or above.

To see all instances of this deprecation message, set the `EMBER_GLOBAL_DEPRECATIONS` environment variable to "all", e.g. `EMBER_GLOBAL_DEPRECATIONS=all ember test`.

### Details ###

Prior to v7.26.6, ember-cli-babel sometimes transpiled imports into the equivalent Ember Global API, potentially triggering this deprecation message indirectly, even when you did not observe these deprecated usages in your code.

The following outdated versions are found in your project:

  * ember-get-config@0.2.4
    * Depends on ember-cli-babel@^6.3.0
    * Added by ember-cli-mirage@2.2.0
    * Added by ember-cli-imgix@2.1.2

Note: Addons marked as "Dormant" does not appear to have any JavaScript files. Therefore, even if they are using an old version ember-cli-babel, they are unlikely to be the culprit of this deprecation and can likely be ignored.
```

So from what I can tell the issue in my case was not that `ember-cli-mirage` brings in the wrong `ember-cli-babel` version, rather its transitive dependency of `ember-get-config@0.2.4` is using `ember-cli-babel@6`.

Looking at [ember-cli-mirage's package.json](https://github.com/miragejs/ember-cli-mirage/blob/master/package.json#L44) we can see that it works with both `ember-get-config` `^0.2.4` and `^0.3.0` so if your application is using the latter there should be no deprecation warning as that one uses `ember-cli-babel@7`.

Generating a new Ember 3.28 app and doing `ember install ember-cli-mirage` does not reproduce the deprecation warnings for me which indicates that in my case I just have some dependency lock file management to do.

We use Yarn so I've used [Yarn resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) to explicitly tell my application to use `ember-get-config@^0.3.0`.
```
# package.json
  "resolutions": {
    "ember-get-config": "^0.3.0"
  }
```
For NPM users I'm not sure of a equivalent API sorry so probably so package-lock hacking for you.

This has removed the deprecation warning for me. I hope this helps someone. Cheers!

cah-briangantzler:
Thanks. Hoping that someone will be avail to merge and release

rahulk94:
Per my earlier comment, I've raised https://github.com/miragejs/ember-cli-mirage/pull/2242 to get ember-get-config's minimum version updated to 0.3.0

PR #2244, opened 2021-11-30
Deprecate miragejs re-exports

SergeAstapov:
@cah-briangantzler this is the dirty but seem to be working solution for re-exports deprecation.

### DIFF
--- a/addon/deprecate-imports.js
+++ b/addon/deprecate-imports.js
@@ -1,0 +1,32 @@
+import { deprecate } from '@ember/debug';
+import * as ecMirageExports from './index';
+import * as mirage from 'miragejs';
+
+const nonDeprecatedImports = ['default'];
+
+export function initDeprecatedImports () {
+  Object.entries(mirage).forEach(([name, value]) => {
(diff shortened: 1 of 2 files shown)


cah-briangantzler:
This works wonderfully. I googled for some time and was not able to find a way. Might want to write a blog post on this.

cah-briangantzler:
The docs say to import like this 
```
import { Factory } from 'ember-cli-mirage';
```

Does anyone import like this?
```
import Factory from 'ember-cli-mirage/factory';
```

because this file supports that. https://github.com/miragejs/ember-cli-mirage/blob/master/addon/factory.js

There are a few others. May have to deprecate that as well.

SergeAstapov:
@cah-briangantzler lot me take another look, will try to catch all the re-exports as they have to be deprecated

cah-briangantzler:
Ironically ember-cli-mirage still imported the re-exports in its tests instead of mirage proper.  Updated those if you want to take a look. 

https://github.com/miragejs/ember-cli-mirage/pull/2246

PR #2253, opened 2021-12-03
Add ember-auto-import and webpack to Ember v4 scenarios

SergeAstapov:
This does not fix CI but fixes following error:

```
To use these addons, your app needs ember-auto-import >= 2: ember-source
```

We'll need to update some other dependencies to make CI pass, but let's keep PRs focused and fixing one error at a time.


SergeAstapov:
Note that `ember-auto-import@2` upgrade is breaking change hence we'll do it later once we start working on v3 of this addon

PR #2255, opened 2021-12-03
Enable Ember Octane features to fix Ember v4 scenarios

SergeAstapov:
This removes deprecations and build failure caused by `ember.edition`  field in `package.json` and missing optional features


SergeAstapov:
CI still fails because of `liquid-fire` which is transitive dependency added by ember-cli-addon-docs.

`liquid-fire` was removed from `ember-cli-addon-docs` v4 but we are not ready for the upgrade just yet, hence merging as is (as it actually fixes one of the issues in v4 scenarios)

PR #2256, opened 2021-12-07
Deprecate miragejs re-exports from `ember-cli-mirage/*` modules

SergeAstapov:
This is continuation for #2244.

As discussed with @cah-briangantzler, this is necessary for the next major release where `ember-cli-mirage` should become thiner Ember.sj glue to bring `miragejs` to the Ember world.

Few more re-exports needs to be updated/deprecated but you should be able to grasp the direction this is heading.

### DIFF
--- a/addon/utils/uuid.js
+++ b/addon/utils/uuid.js
@@ -1,6 +1,10 @@
 import { _utilsUuid } from 'miragejs';
 import { deprecateNestedImport } from '../deprecate-imports';
 
+/**
+ @function uuid
+ @hide
+ */
 export default function uuid (...args) {
   const message = `Importing 'uuid' from 'ember-cli-mirage/utils/reference-sort' is deprecated. ` +
     `This wasn't intended to be a public API. If you absolute know what you are doing, ` +
--- a/addon/utils/uuid.js
+++ b/addon/utils/uuid.js
@@ -5,8 +5,9 @@ import { deprecateNestedImport } from '../deprecate-imports';
  @function uuid
  @hide
  */
-export default function uuid (...args) {
-  const message = `Importing 'uuid' from 'ember-cli-mirage/utils/reference-sort' is deprecated. ` +
+export default function uuid(...args) {
+  const message =
+    `Importing 'uuid' from 'ember-cli-mirage/utils/reference-sort' is deprecated. ` +
     `This wasn't intended to be a public API. If you absolute know what you are doing, ` +
     `install the \`miragejs\` package and use \`import { _utilsUuid } from 'miragejs';\` instead.`;
 
@@ -14,4 +15,3 @@ export default function uuid (...args) {
 
   return _utilsUuid(...args);
 }
-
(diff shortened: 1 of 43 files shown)


cah-briangantzler:
Yea, Most these files will be able to be deleted in the breaking release

cah-briangantzler:
I thought you had a PR merged that added the auto-import 2.x to the tests? Or is that still a work in progress

SergeAstapov:
> I thought you had a PR merged that added the auto-import 2.x to the tests? Or is that still a work in progress

I think you refer to https://github.com/miragejs/ember-cli-mirage/pull/2253/files where I updated `ember-try` scenarios as I believe ember-auto-import v2 is a breaking change and we'll do it once we start working on v3 of this addon

cah-briangantzler:
Yes thats what I was referring to. But since it was merged I wondered why this failed the ember 4.x tests. I hadnt looked, thought  it was auto import, now it looks like is `Named outlets were removed in Ember 4.0`. Not sure where that is being used in the tests. Prob the docs and updating addon docs will get past that. Perhaps in addon docs and when they are updated that might go away.

SergeAstapov:
@cah-briangantzler I've outlined the (IMO) safe plan for addon-docs upgrade in [this comment](https://github.com/miragejs/ember-cli-mirage/pull/2251#issuecomment-985227158), which is IMO orthogonal ti changes in this PR but still should bring us closer to the new [major] release and green CI.

cah-briangantzler:
Since you have extracted a lot of the code to some methods, you might want to consider (in the future) creating an addon that supplies this re-export deprecating. I googled for a long time and didnt find any code to do it. So an addon that would provide this for others when they need it would be nice.

cah-briangantzler:
I think this is fantastic work and sad that it will all go away when the deprecations are removed.  I think that some of the code could be put in an addon to be used other addons that also need to deprecate their own exports.

cah-briangantzler (approved):

SergeAstapov:
> I think this is fantastic work and sad that it will all go away when the deprecations are removed. I think that some of the code could be put in an addon to be used other addons that also need to deprecate their own exports.

Thanks! I was thinking about making it generic but couldn't came up with any nice interface.
Each import is so specific and combinations of all named/default imports + named/default re-exports would make that util function have inconvenient API.

I think some sort of searchable blog post would be helpful (this PR also should be indexed and searchable by Google and co).

cah-briangantzler (approved):

addon/utils/uuid.js
```diff
@@ -1 +1,17 @@
...
+import { _utilsUuid } from 'miragejs';
+import { deprecateNestedImport } from '../deprecate-imports';
+
+/**
+ @function uuid
+ @hide
+ */
+export default function uuid(...args) {
+  const message =
+    `Importing 'uuid' from 'ember-cli-mirage/utils/reference-sort' is deprecated. ` +
```
Turbo87, line 10: I assume this was meant to be `ember-cli-mirage/utils/uuid`?

SergeAstapov: @Turbo87 indeed! created #2306 to fix this

SergeAstapov: @Turbo87 Released in [v2.3.1](https://github.com/miragejs/ember-cli-mirage/releases/tag/v2.3.1)

PR #2265, opened 2021-12-12
Add 3.28 LTS scenario to test matrix

SergeAstapov:


SergeAstapov:
Tests fail with error

```
TypeError: Ember.computed.reads is not a function at http://localhost:7357/assets/vendor.js, line 112172
```

will keep looking and updating other reps

SergeAstapov:
updated some other deps so this now pass

PR #2285, opened 2021-12-13
Run ember-cli-update to v3.28.4

SergeAstapov:


cah-briangantzler:
There are a lot of files given the sub projects. Looked over it, didnt see anything to comment on.

PR #2297, opened 2022-01-05
remove some usage of runtime require

ef4:
`require` a pre-ES-module feature that continues to work in Ember but is fragile and unreliable under ES-module-first build tooling like Embroider.

This PR replaces two uses with embroider macros instead, which are designed to allow dynamism without sacrificing static analysis.

### DIFF
--- a/addon/get-rfc232-test-context.js
+++ b/addon/get-rfc232-test-context.js
@@ -1,4 +1,4 @@
-import require from 'require';
+import { importSync, dependencySatisfies, isTesting } from '@embroider/macros';
 
 /**
   Helper to get our rfc232/rfc268 test context object, or null if we're not in
@@ -10,8 +10,8 @@ export default function getRfc232TestContext() {
   // Support older versions of `ember-qunit` that don't have
   // `@ember/test-helpers` (and therefore cannot possibly be running an
   // rfc232/rfc268 test).
-  if (require.has('@ember/test-helpers')) {
-    let { getContext } = require('@ember/test-helpers');
+  if (dependencySatisfies('@ember/test-helpers', '*') && isTesting()) {
+    let { getContext } = importSync('@ember/test-helpers');
     return getContext();
   }
 }
--- a/addon/utils/ember-data.js
+++ b/addon/utils/ember-data.js
@@ -1,18 +1,9 @@
-/* global requirejs */
-
-function _hasEmberData() {
-  let matchRegex1 = /^ember-data/i;
-  let matchRegex2 = /^@ember-data/i;
-
-  return !!Object.keys(requirejs.entries).find(
-    (e) => !!e.match(matchRegex2) || !!e.match(matchRegex1)
-  );
-}
+import { dependencySatisfies } from '@embroider/macros';
 
 /**
   @hide
 */
-export const hasEmberData = _hasEmberData();
+export const hasEmberData = dependencySatisfies('ember-data', '*');
 
 /**
   @hide
--- a/package.json
+++ b/package.json
@@ -53,10 +53,14 @@
   },
   "peerDependencies": {
     "@ember/test-helpers": "*",
+    "ember-data": "*",
     "ember-qunit": "*"
   },
   "peerDependenciesMeta": {
     "@ember/test-helpers": {
+      "optional": true
+    },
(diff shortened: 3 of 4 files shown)


addon/utils/ember-data.js
```diff
@@ -1,18 +1,9 @@
...
-    (e) => !!e.match(matchRegex2) || !!e.match(matchRegex1)
-  );
-}
+import { dependencySatisfies } from '@embroider/macros';
 
 /**
   @hide
 */
-export const hasEmberData = _hasEmberData();
+export const hasEmberData = dependencySatisfies('ember-data', '*');
```
cah-briangantzler, line 6: Thanks, I never liked the way it checked before.

addon/get-rfc232-test-context.js
```diff
@@ -1,4 +1,4 @@
-import require from 'require';
+import { importSync, dependencySatisfies, isTesting } from '@embroider/macros';
```
cah-briangantzler, line 1: Thanks. This entire file should be done away with in the next version

package.json
```diff
@@ -51,6 +51,18 @@
...
+    "ember-qunit": "*"
+  },
+  "peerDependenciesMeta": {
+    "@ember/test-helpers": {
+      "optional": true
+    },
+    "ember-qunit": {
+      "optional": true
+    }
+  },
```
cah-briangantzler, line 65: Could you elaborate on why these are here. They are listed under devDependancies. Why are these two peer and optional and not the rest? Trying to understand if I should be doing something like this in other addons.

ef4: `dependencySatisfies` follows the real NPM package resolving rules. 

The rule is that a package looks in its own `node_modules` folder and then proceeds to look up the filesystem for higher node_modules folders.

An app will often *accidentally* allow ember-cli-mirage to resolve the app's copy of ember-qunit:

```
/the-app
└─ node_modules
    └── ember-qunit
    └── ember-cli-mirage
```

But this is not guaranteed. For example, if the app lives in a monorepo with many projects, some dependencies may be shared at the top level and others may not:


```
/the-monorepo
└─ node_modules
│  └── ember-cli-mirage
└── app1
│  └── node_modules
|    └── ember-qunit
```

In this case, app1 can see both addons, but ember-cli-mirage cannot resolve ember-qunit, so `dependencySatisfies` would return false.

NPM's solution to this problem is `peerDependencies`. They declare that you must see the *same copy* as your parent package. With a peerDependency declaration, NPM would ensure that ember-cli-mirage and ember-qunit are always located so that ember-cli-mirage can see the correct copy of ember-qunit.

Usually peerDependencies also appear in devDependencies. That's because they're simultaneously dependencies of the dummy app and peerDependencies of the addon.

I marked these as optional peer dependencies because the existing code also treats them as optional. By declaring these optional peer deps, we're telling NPM that if the app has these packages, we need to be able to see them too, but if it doesn't have them that's OK.

cah-briangantzler: I knew most of what you said, what I missed was the `dependencySatisfies` is why you did this. `@ember/test-helpers` and `ember-data` appears in `dependencySatisfies`.

I had wanted to make `ember-qunit` a `peerDependancy` much like `ember-qunit` made `qunit` a `peerDependancy`.  In your example of the mono-repo, should not event `devDependancy` be listed as a `peer` because we cant predict how the end consumer will structure their repo?

Also should `ember-data` be listed as a `peerDependancy` since it is directly referred to in a `dependencySatisfies`?

SergeAstapov (approved):

SergeAstapov:
@ef4 one more thing:
`ember-cli-mirage` still requires `"node": ">= 10.*"` whereas `@embroider/macros` has dropped Node.js 10 support in v0.42.0

@ef4 can we use `"@embroider/macros": "^0.41.0"` for the time being and bump to the latest version once we release `ember-cli-mirage@3.0.0`?

package.json
ef4: You're right, it would be correct to add ember-data to peerDeps too. Done.

ef4:
Yes, I downgraded back to the older embroider/macros for Node 10 support.

SergeAstapov:
@ef4 Thank you!

PR #2300, opened 2022-01-10
Pin colors transitive dependency to fix CI

SergeAstapov:


SergeAstapov:
For background one may follow https://github.com/Marak/colors.js/issues/285

PR #2301, opened 2022-01-11
Fix: avoid @embroider/macros issue

wozny1989:
After today's [release](https://github.com/mansona/ember-get-config/releases/tag/v1.0.0) `ember-get-config` app throw error:
```
Build Error (broccoli-persistent-filter:Babel > [Babel: @embroider/macros]) in @embroider/macros/runtime.js

/Users/adam/Documents/Projects/opensource/my-forks/ember-cli-mirage/@embroider/macros/runtime.js: @babel/template placeholder "ADDON_DOCS_DEPLOY_VERSION": Expected string substitution

Stack Trace and Error Report: /var/folders/q6/w2pl66k56ldbzpcmslws5vmh0000gn/T/error.dump.4552549e8357d2d49d82ff4e3a084240.log
```

Bump `embroider` packages resolve this problem. Reported similar issue: https://github.com/embroider-build/embroider/issues/883


SergeAstapov:
@wozny1989 we can't bump `@embroider/macros` package beyond v0.41 as v2 of this addon still supports Node.js 10 and `@embroider/macros` v0.42 and up require Node.js 12 minimum.

This addon still depends on `ember-get-config` versions `0.2.4 - 0.5.0`.

I'm sorry you have this issue but we are not yet ready for this dependency upgrade.
We are working towards v3 where we'll drop Node.js 10 support and upgrade dependencies to the latest

wozny1989:
@SergeAstapov Ahh, now I see latest commits about node support. I wonder what we can do, because I tried to create Ember app in v3.24 from scratch and when I add `ember-cli-mirage` package then build crash

cah-briangantzler:
They have been bumping those versions every few days https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L28

If those updates to the blueprints are getting out they are not really applicable to a 3.24 app. When you created your app was it 3.24? or actually higher and you downgraded it to 3.24?

Hopefully a release of ember-cli-mirage cant get out soon to close out the 2.x series and then a 3.x series will fix all this (as mentioned above)

wozny1989:
@cah-briangantzler Today I created an app in 3.24 but that's weird - now I noticed that when I install dependency via `yarn` problem still exists but if I use `npm` it works fine

cah-briangantzler:
Even weirder, we had the issue when the tests ran releasing 2.3.0 using npm and got past it switching to yarn

PR #2314, opened 2022-01-14
Remove deprecated `miragejs` re-exports

SergeAstapov:
In PRs #2256 and #2244 re-exports from `miragejs` package were deprecated and released in version v2.3.0

This PRs deletes deprecated re-exports making `ember-cli-mirage` thinner layer on top of `miragejs`


cah-briangantzler:
Looks like we have to cleanse the test apps from using the deprecated exports as well.. Example. https://github.com/miragejs/ember-cli-mirage/blob/master/test-projects/01-basic-app/tests/acceptance/ember-data-test.js#L3

cah-briangantzler (approved):

PR #2315, opened 2022-01-14
Run `npx ember-cli-update --to=4.1.0` to sync with the latest blueprint

SergeAstapov:


cah-briangantzler:
This repo doesnt really depend on ember-source (its in the devDependancies), The apps really supply the ember source.

In this case the test apps will effectively be updated to ember source 4.x, since code has been removed, you might find the tests will fail with they run the test-projects tests and you will have to update those apps to ember 4

PR #2323, opened 2022-01-18
Point users of deprecated inflector methods to ember-inflector

jrjohnson:
The [@ember/string](https://api.emberjs.com/ember/release/classes/String) library doesn't provide these methods, they are to be
found in [ember-inflector](https://github.com/emberjs/ember-inflector).


SergeAstapov:
@jrjohnson just after merge realized this should point `v2` branch so recreated PR in #2324

PR #2329, opened 2022-01-20
Deprecate the routes only version of the config

cah-briangantzler:


addon/start-mirage.js
```diff
@@ -51,6 +51,18 @@ export default function startMirage(
...
   if (baseConfig && baseConfig.length === 0) {
     routes = baseConfig;
+
+    deprecate(
+      'The routes only function has been deprecated. Please use the make server version your default export in the config.',
+      false,
+      {
+        id: 'ember-cli-mirage-config-routes-only-export',
+        for: 'ember-cli-mirage',
+        since: '2.3.3',
```
SergeAstapov, line 61: @cah-briangantzler I think we should do 2.4.0 here, as per SemVer spec

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated.

cah-briangantzler: just figured the next release was going to be 2.3.3.   adding deprecations should be ok for 2.3.3 but if you are going to make the next version 2.4.0 then yea Ill change it. 

Please confirm what the next version number is going to be

cah-briangantzler: Didnt read the quote fully.   Np.. Will fix, sec

SergeAstapov (approved):

PR #2333, opened 2022-01-21
v3.0.0-alpha.1

SergeAstapov:


.github/workflows/publish.yml
```diff
@@ -24,3 +24,4 @@ jobs:
       - uses: JS-DevTools/npm-publish@v1
         with:
           token: ${{ secrets.NPM_TOKEN }}
+          tag: next
```
SergeAstapov, line 27: a small safety net cc @cah-briangantzler

PR #2334, opened 2022-01-22
Export default in server docs

jrjohnson:
Exporting a named function is deprecated and this now matches the docs
directly above.


cah-briangantzler:
The docs need updated quite a bit yes. Weird that there seems to be two similar PRs.

In removing the old code the supports the previous way of doing this, I plan on updating the docs

cah-briangantzler:
I removed all the code and updated the docs. https://github.com/miragejs/ember-cli-mirage/pull/2337 Feel free to fix any of the docs I might have missed after that PR is merged.

Thanks for helping.

SergeAstapov (approved):

PR #2335, opened 2022-01-22
Export default in server docs

jrjohnson:
Exporting a named function is deprecated and this now matches the docs
directly above. Same as #2334, but against the v2 branch.


cah-briangantzler:
The docs need updated quite a bit yes. Weird that there seems to be two similar PRs.

In removing the old code the supports the previous way of doing this, I plan on updating the docs

cah-briangantzler:
Ahhh, I didnt read, this was against the V2 branch. I actually left it this way as a way of testing that the deprecation was working :)

SergeAstapov (approved):

PR #2355, opened 2022-02-10
Detect @ember-data/model instead of ember-data meta-package

jrjohnson:
While ember-data is usually installed it's also possible to use
@ember-data/model (and other packages directly). Since ember-data
provides @ember-data/model I don't think we need to detect both.

I'm not sure why this resolves to false in an app with either `ember-data` or `@ember-data/model` installed. I would expect true in both of those cases. I'm hoping @runspired or @ef4 might be able to shine some light on why this package may not work with the check in `dependencySatisfies`? I feel like maybe I'm missing something obvious, but I can't see it.

### DIFF
--- a/packages/ember-cli-mirage/addon/utils/ember-data.js
+++ b/packages/ember-cli-mirage/addon/utils/ember-data.js
@@ -3,7 +3,9 @@ import { dependencySatisfies } from '@embroider/macros';
 /**
   @hide
 */
-export const hasEmberData = dependencySatisfies('ember-data', '*');
+export const hasEmberData =
+  dependencySatisfies('@ember-data/model', '*') ||
+  dependencySatisfies('ember-data', '*');
 
 /**
   @hide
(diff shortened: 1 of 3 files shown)


SergeAstapov:
@jrjohnson we have only `ember-data` listed in `peerDependencies `, probably we should add or even replace with `@ember-data/model`

runspired:
What's the goal of this check? Probably safest to just check for the store package. The model package is optional for ember-data, the store is not.

jrjohnson:
I added (instead of replaced) it because I'm not sure about the DX of NPM/Yarn when they're missing. It might by confusing to be told to install the `@ember-data/model` when all the guides still talk about using `ember-data`. 

@runspired as far as I can tell this is used specifically to detect the presence of possible models [here](https://github.com/miragejs/ember-cli-mirage/blob/master/addon/ember-data.js#L25) so I went with the model package. I'm not sure if store would be a better choice though.

jrjohnson:
And that was the trick to making it work @SergeAstapov. I didn't even think to make it a peerDep here assuming if it was installed in the app then it would work. Which is the "something obvious" I was missing! This is ready for a serious look now, I have no opinion on what ember-data package we should depend on here. I could also add several of them?

PR #2366, opened 2022-02-21
Add TS declaration for `setupMirage`

dfreeman:
This PR ensures that `import { setupMirage } from 'ember-cli-mirage/test-support';` works as expected for TypeScript users. It also adds a small `types-tests` suite to verify that they work as expected.

@SergeAstapov if I were to open a second copy of this PR against the `v2` branch, is there a chance you'd be willing to include it in a 2.x release as well, since 3.0 is still in alpha? 🙂

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -54,5 +54,12 @@ module.exports = {
       files: ['tests/**/*-test.{js,ts}'],
       extends: ['plugin:qunit/recommended'],
     },
+    {
+      // types tests
+      files: ['types-tests/**/*.ts'],
+      rules: {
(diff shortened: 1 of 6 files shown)


SergeAstapov:
> @SergeAstapov if I were to open a second copy of this PR against the `v2` branch, is there a chance you'd be willing to include it in a 2.x release as well, since 3.0 is still in alpha? 🙂

Yes! I'm 100% up for this as v2 is still fully supported and I think we'll support it for some time until v2 surpasses v2

SergeAstapov (approved):

PR #2375, opened 2022-03-23
Remove deprecations for trackRequests, testConfig and baseConfig

cah-briangantzler:
Removed the `trackRequest` option. Updated docs to show how to use with MirageJS.
Removed the `testConfig` option for `startMirage`. Never documented.
Removed some references to `baseConfig` that should have been removed when `makeServer` was made the only way to create the server.


cah-briangantzler:
Prettier got me again. I dont think I changed that line, but I did change the file :( Will wait til all the tests are done in case there is something more and then will update

kevinansfield:
Does `testConfig` need to be removed? We use it quite a bit as it allows us to mock only the in-development API endpoints for new features whilst developing without having to opt-in to the whole of the system being mocked.

I thought that was documented at some point, or at least included in the default generated config file 🤔 Looking back to 2015 when we first added ember-cli-mirage the following was at the bottom of our config:

```
/*
 You can optionally export a config that is only loaded during tests
 export function testConfig() {

 }
 */
```

cah-briangantzler:
Kevin,

 Please see issue https://github.com/miragejs/ember-cli-mirage/issues/2377

I created it to answer your question. Figured more people would find the answer in issues rather than in PR comments.

SergeAstapov (approved):

PR #2387, opened 2022-04-13
Bump `@embroider/*` packages to latest

SergeAstapov:
Solves this issue https://github.com/ember-learn/ember-cli-addon-docs/pull/1137 which is not released yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@ember/jquery": "^2.0.0",
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.6.0",
-    "@embroider/test-setup": "^1.0.0",
+    "@embroider/test-setup": "^1.6.0",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "@types/qunit": "^2.11.3",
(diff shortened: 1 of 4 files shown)


cah-briangantzler:
Its sadly humorous that the resolutions section is growing due to embroider when you would want that section to be as small as possible. Just had to add the `@embroider/util` to my applications `package.json` recently.

SergeAstapov:
> Its sadly humorous that the resolutions section is growing due to embroider when you would want that section to be as small as possible. Just had to add the `@embroider/util` to my applications `package.json` recently.

@cah-briangantzler yeah, ecosystem updates slower than I wanted to.

The only real reason I need this - fix docs app https://www.ember-cli-mirage.com/versions/v3.0.0-alpha.3/

Almost all of those actually come from `ember-cli-addon-docs` which I'm thinking about extracting into separate app.
With that, we can have a monorepo setup with addon, test app (now dummy app), docs app, test app1 (now test-projects/01-basic-app) and test app2 (now test-projects/02-app-that-excludes-mirage) being packages within that monorepo.

As a result, CI will run faster and we could update docs app independently form the test app(s).
But that's indeed separate PR I hope will get to somewhere next week.

cah-briangantzler:
Separating into to its own App is definitely the way to go. I still advise on cutting down the docs to ONLY what this addon does and leave the docs part about how to use Mirage to the Mirage Repo.

Yes the trend with the V2 versions of addons puts the test app in a separate app that solves many of these problems. Not sure if this can be a V2 addon due to the way it scans the application for mirage models, factories etc, it makes the addon reach into the app package space which Im sure is allowed when embroider is more strict. 


I think we should maybe move towards something like below. While its a little boilerplate, the addon supplies the code, but the code runs within the app instead of the addon. This should give the code access to the app assets in a way that might make embroider happier
```js
import {
  discoverEmberDataModels,
  discoverMirageModels,
  discoverMirageFactories,
  discoverMirageSerializers,
} from 'ember-cli-mirage';
import { createServer } from 'miragejs';
import addonConfig from 'ember-rxid/mirage/routes/config';

// Application Factories, Models, Serializers etc would no longer be supplied on config
export default function (config) {
  let finalConfig = {
    ...config,
    factories: discoverMirageFactories(), 
    models: { ...discoverEmberDataModels(), ...discoverMirageModels() },
    serializers: discoverMirageSerializers()
    routes,
  };

  return createServer(finalConfig);
}

function routes() {
   // Your routes here
}

```

SergeAstapov:
> cutting down the docs to ONLY what this addon does and leave the docs part about how to use Mirage to the Mirage Repo.

Yes, I'm totally in favor of doing this.

PR #2393, opened 2022-04-13
Drop support for Node.js 12

SergeAstapov:
Node.js 14 is currently in Maintenance LTS mode and slated for End Of Life (EOL) on 2023-04-30 (end of this month).


cah-briangantzler:
The title is misleading. I thought you were dropping Node 14,  you meant dropping 12 in favor of 14 :) 
Might want to update the title to make it clearer

cah-briangantzler (approved):

cah-briangantzler:
What about https://github.com/miragejs/ember-cli-mirage/blob/master/.github/workflows/deploy.yml#L14

SergeAstapov:
> What about https://github.com/miragejs/ember-cli-mirage/blob/master/.github/workflows/deploy.yml#L14 While its in the Github actions only really, should prob be consistant

Good point! Just pushed an update to update all workflows

PR #2395, opened 2022-04-14
Restructure into monorepo

SergeAstapov:


cah-briangantzler:
What is the difference between `test-packages` and `test-projects`?

This appears to keep the actual test still in dummy in the addon. Future restructure to to extract that to a `test-app` as well?

SergeAstapov:
> What is the difference between `test-packages` and `test-projects`?

technically none.
Since we introduce `packages` folder (which is widely used for monorepos) makes sense just to rename "projects" in to "packages" for consistency

> This appears to keep the actual test still in dummy in the addon. Future restructure to to extract that to a `test-app` as well?
Correct! Kept PR focused as extracting dummy app into test app would take another ~100 files changes

cah-briangantzler:
I may have misread the PR, so many files changed. It looked like the end result would have both, didnt notice it was just a rename, wouldnt have ask. will look some more

cah-briangantzler:
The scripts like `link.sh` used to share the node_modules amoung all the projects. With the workspaces assuming the scripts are no longer needed. Nice

cah-briangantzler:
It seems this file https://github.com/miragejs/ember-cli-mirage/pull/2395/files#diff-f79b9ecbffd42f337bfa51417cd40e5a274ce039d5042b27cc64cebf1558a1a0 was a `.gitkeep` renamed to `ci.yml` but thinking this file and directory should not be here. Could be misreading PR again

cah-briangantzler (approved):
Is there suppose to be an empty file at `packages/ember-cli-mirage/.github/workflows/ci.yml?

Other than that, looked, lotta file changes obviously. But seems correct for what i could see

SergeAstapov:
> Is there suppose to be an empty file at `packages/ember-cli-mirage/.github/workflows/ci.yml?

it was already removed, thanks for noticing!

PR #2407, opened 2022-04-22
Extract docs into separate app

SergeAstapov:
Follow up for #2395


cah-briangantzler:
It looked weird, I thought you had failed to delete some files, but I think its because the test app hasnt been extracted from the package, just the docs.

cah-briangantzler (approved):

SergeAstapov:
> It looked weird, I thought you had failed to delete some files, but I think its because the test app hasnt been extracted from the package, just the docs.

This usually is done when addon converted to v2 format and we are not there yet.

PR #2411, opened 2022-05-03
Replace faker with @faker-js/faker

### LINKED ISSUE #2298: FYI: faker.js is no more
Just a heads up that faker.js is no more- the most recently published version is empty. There are forks available; also chance.js has been pegged as a potential replacement.

SergeAstapov:
Closes #2298.


cah-briangantzler (approved):
If we streamlined the docs down to only the things this repo is now responsible for, all this would go away and this update would not need to be done.

SergeAstapov:
> If we streamlined the docs down to only the things this repo is now responsible for, all this would go away and this update would not need to be done.

yes, we should strip down the docs.
this still needed as faker used in `test-packages/01-basic-app/mirage/factories/user.js`

cah-briangantzler:
This line https://github.com/miragejs/ember-cli-mirage/blob/master/packages/ember-cli-mirage/package.json#L58 should be able to be removed then correct? Since it exists in the docs app and the test app package.json, it shouldnt need to be in the addon package.json. (Sorry I didnt see this earlier)

SergeAstapov:
> This line https://github.com/miragejs/ember-cli-mirage/blob/master/packages/ember-cli-mirage/package.json#L58 should be able to be removed then correct? Since it exists in the docs app and the test app package.json, it shouldnt need to be in the addon package.json. (Sorry I didnt see this earlier)

good point! Look like you are right

PR #2412, opened 2022-05-03
Simplify release setup and incorporate docs deploy

SergeAstapov:
this improves release setup and moves .md files back to root to not break DX when you open one.

Additionally, docs deploy will run as part of the release process.


cah-briangantzler (approved):
I think in some of the V2 format conversions I have done, the directions on Ed's site has us moving these files into the addon dir where they prob should be in the root. May have to look into fixing thos and maybe getting Ed to update the migration instructions

SergeAstapov:
yeah, I made similar PR to the guide and there was a good discussion around what might be best option https://github.com/embroider-build/embroider/pull/1188

cah-briangantzler:
Awesome, you were already on it. Great, thanks

PR #2444, opened 2022-09-12
Resolve the deprecate-early-static warning from ember-data

kiwi-josh:
By discovering the ED models via the `modelFor` method on the `store`, rather than statically `require`'ing them (@runspired lead me down this path as a resolution).

Resolves: https://github.com/miragejs/ember-cli-mirage/issues/2441

I've confirmed that the deprecation no longer appears for our app when running tests:
1. via `localhost:4200/tests`
2. via `yarn run test`

_For the record I don't think this PR should be merged in its current state - its more a POC to demonstrate there is a way to resolve this issue. Theres probably a much nicer way to get a handle on an application instance that doesn't involve a global on `window`_

### DIFF
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -16,6 +16,8 @@ const { default: baseConfig, testConfig, makeServer } = config;
 export default {
   name: 'ember-cli-mirage',
   initialize(application) {
+    window.__app_for_mirage = application;
+
     if (baseConfig) {
       application.register('mirage:base-config', baseConfig, {
         instantiate: false,
--- a/app/initializers/ember-cli-mirage.js
+++ b/app/initializers/ember-cli-mirage.js
@@ -2,6 +2,7 @@ import ENV from '../config/environment';
 import getRfc232TestContext from 'ember-cli-mirage/get-rfc232-test-context';
 import startMirageImpl from 'ember-cli-mirage/start-mirage';
 import * as config from '../mirage/config';
+import { AppSymbolForEmberDataModels } from 'ember-cli-mirage/utils/ember-data';
 const { default: baseConfig, testConfig, makeServer } = config;
 
 //
@@ -16,7 +17,7 @@ const { default: baseConfig, testConfig, makeServer } = config;
 export default {
   name: 'ember-cli-mirage',
   initialize(application) {
-    window.__app_for_mirage = application;
+    window[AppSymbolForEmberDataModels] = application;
 
     if (baseConfig) {
       application.register('mirage:base-config', baseConfig, {
--- a/addon/ember-data.js
+++ b/addon/ember-data.js
@@ -14,6 +14,19 @@ const { modulePrefix, podModulePrefix } = config;
 let DsModels, Models;
 let DsSerializers, Serializers;
 
+function _getAppInstance() {
+  const application = window.__app_for_mirage;
+  let appInstance = application.__deprecatedInstance__;
+  // If an appInstance wasn't found (such as when running the tests)
+  // then instantiate one manually, so we can use it to discover the
+  // ember-data models in a way that doesn't trigger deprecations.
+  if (!appInstance) {
+    application._buildDeprecatedInstance();
+    appInstance = application.__deprecatedInstance__;
+  }
+  return appInstance;
+}
+
 /**
  * Get all ember data models under the app's namespaces
  *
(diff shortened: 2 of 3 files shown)


app/initializers/ember-cli-mirage.js
```diff
@@ -16,6 +16,8 @@ const { default: baseConfig, testConfig, makeServer } = config;
 export default {
   name: 'ember-cli-mirage',
   initialize(application) {
+    window.__app_for_mirage = application;
```
SergeAstapov, line 19: @kiwi-josh minor, I wonder if it might be safer to use a `Symbol` here, to ensure no one else uses `__app_for_mirage` even accidentally

SergeAstapov (approved):

SergeAstapov:
@kiwi-josh @cah-brian-gantzler this seem ok solution to me to avoid deprecation although I would like to consider some way to be able to explicitly access store service (don't have better suggestion wight now though)

runspired:
while this correctly resolves the deprecation  (for now) I'd note a few things here

- `modelFor` is also slated for deprecation. Though it likely will not be deprecated for 5.0, it will stop working if not using `@ember-data/model` which is slated to be removed from the default setup for 5.0
- what mirage has done here has always been wrong. models have always been able to be registered dynamically.
- what mirage is doing will be even more wrong in the near future: there will be no models and dynamic registration  of schema will be highly encouraged
- its often the case in apps you only want to register schemas for tests a single time, while in addons you often want to register schemas per-test

The best future-proof refactor here is for mirage to expose a hook and a format for supplying schemas. The easiest thing would be to align to the format the store's schema service currently provides today.

app/initializers/ember-cli-mirage.js
kiwi-josh: @SergeAstapov yeah good call out - I've updated the code to use a `Symbol` instead. Let me know if theres a better way you would like it done.

cah-brian-gantzler:
I wrote a lengthy response to this and lost it :( give me a bit to re-write it.

The TL;DR is, I do not believe this addon is required anymore

cah-brian-gantzler:
Wrote the reply as an issue. Thought that might be better found later. If this addon is kept, should be stripped down even more and some processes handled differently. 

Embroider has already raised issues about the way this addon discovers files, and all @runspired points are very valid. In order to prepare for the future, current patterns and process are going to have to be adapted

cah-brian-gantzler:
As @runspired pointed out about discovering the data models, heres my actual situation

My backend has many fields named differently than what I named them on the front end. I have ember serializers that rename the fields. Because discover models makes mirages models look like my front end models (mirage should look like your backend, thats what it is pretending to be),  the fields are named wrong when mirage puts them in the payload. So I then had to create mirage serilizers to to rename the mirage field names into the backend field names so the front end serializer can rename them to the front end field names LOL. This was the main driver to me write the discover serializers (which was really a hack fixing a hack). 

TL;DR 
having mirage discover your models makes mirage models look like your front end models not your backend models which is what mirage is meant to simulate. 

Auto discover was a convenience shortcut to avoid taking the time to do the right thing. When I first learned mirage I thought auto discover was awesome and for most people it works, but they do not actually understand that the tests are mis representing the back end.  The first time you rename a field in a serializer you figure that out, but strangely there were never really any issues opened about it. I assume they created a mirage serializer to add a wrong to a wrong (which works, but didnt make it a right).     (Three lefts do make a right though)

kiwi-josh:
Sounds good. I'm looking forward to being able to point our app back at a stable version of ember-cli-mirage (instead of my fork)

cah-brian-gantzler:
@SergeAstapov what do you think? would you be able to do this? Work has me busy at the moment

kiwi-josh:
Kindly requesting a bump on this one 🙏 🙇

sebastianhelbig:
It would be really nice to get this solved. :-)

cah-brian-gantzler:
Kinda lost track of this. Looks like the code was approved. We just need to merge it? The tests failed, but so long ago the logs are gone. Should we re-run the tests and see why? I think we have an issue with tests failing when they shouldnt and we need to revamp the CI pipeline.

kiwi-josh:
If it helps, we have been running this patch in production since september last year when it was opened with no issues

gilest:
Hi all – ember-data v5 has been released now.

Shall we rebase and see what's causing the CI failures?

esbanarango:
Hello 👋 .

Any project using `mirage` trying to bump to `ember-data: 5.0.0` is blocked by this PR 🙏 .

bartocc:
Hello. What could we do to help this PR get merged?

PR #2457, opened 2022-10-24
Updated blueprint and guide for missing import

cah-brian-gantzler:


SergeAstapov (approved):

cah-brian-gantzler:
I dont know whats up with the jobs running. If you look at the failing jobs, its a build error of some sort. But its transient I believe. If you run the jobs again they work. If I run them locally they work. 

Its frustrating

cah-brian-gantzler:
I noticed the change for fixing the guide was missing so I added that. Could you re-approve please. Curious to see if the tests are able to build (then run) this time

PR #2464, opened 2022-11-08
include ember-get-config v2 in supported version range

jelhan:


cah-brian-gantzler:
Going to have to invest some time into why the tests are failing, or change the supported version 😦

cah-brian-gantzler:
I dont have a problem with the change. 

Because of ember-data changes, we are going to have to make changes here that mostly will not even require ember-get-config anymore

SergeAstapov:
Thank you @jelhan! This has been already updated in master as part of some other changes

PR #2501, opened 2023-07-19
Widen `ember-get-config` range

bertdeblock:


SergeAstapov:
Thank you @bertdeblock! This has been already updated in `master` as part of some other changes

Issue #641, opened 2016-03-15
Using mirage in a pre-ember-cli app


snovity:
Hi. My goal is to port non-CLI Ember 1.1 app to Ember 2.x. Currently there is no working back-end for the apps, so I though I would use mirage to mock a back-end for both of them to speed up transitioning. Is there any way of using mirage in that old non-CLI Ember 1.1 app?

samselikoff:
There is no dist build of Mirage currently. If you wanted, you could look
at compiling down index.js (or some other way of including the core files
under /addon) and then looking at the Initializer to see how to instantiate
a Mirage server (basically new Server(options)).

snovity:
@samselikoff thanks for the prompt answer, will try to do that

n8n8baby:
@snovity Did you have any success with this? I am in the same boat...

snovity:
@n8n8baby It was not easy, but I was able to do it, I used https://github.com/ef4/ember-giftwrap plus backported three Ember functions `isNone`, `isEmpty`, `isBlank`. Also had to drop the nice Ember CLI imports and revert to old school global naming of factories and scenarios. And certainly there was some luck involved when matching Ember & Mirage versions that would be successfully produce a gitfwrapped Mirage...

snovity:
@n8n8baby If you need this stuff, I can create a repo with giftwrapped Mirage version and instructions of how to integrate it into the old Ember app

n8n8baby:
@snovity That would be great if you could do that. Thanks a million!

snovity:
@n8n8baby sorry for the delay, here is the repo https://github.com/snovity/mirage-giftwrapped

n8n8baby:
Finally getting around to checking it out. Thanks again!

n8n8baby:
Works like a charm. You are, indeed, the MAN! :)

n8n8baby:
Are you looking for a job?

snovity:
hey @n8n8baby, nope, not looking, sorry :)

n8n8baby:
Can't blame a man for trying :) Thanks again!

PR #2502, opened 2023-08-01
refactor(ember-5): access Ember Models via Store

Pixelik:
## Refactor 

- ember-data 5.x only allows accessing Models via the Store
- This refactoring removes the following Deprecation Warning:

Screenshot 2023-08-01 at 13 35 12
<br><br>

Instead of accessing a Model with: 
```
// ember-cli-mirage/ember-data.js

let modelName = matches[1];
let model = require(path, null, null, true).default
```
It will now be accessed with:
```
// ember-cli-mirage/ember-data.js

let modelName = matches[1];
let model = application.__container__.lookup('service:store').modelFor(modelName)
```

### DIFF
--- a/packages/ember-cli-mirage/addon/ember-data.js
+++ b/packages/ember-cli-mirage/addon/ember-data.js
@@ -15,6 +15,24 @@ let DsModels, Models;
 let DsSerializers, Serializers;
 
 /**
+ * Gets application from global context, set inside initializers/ember-cli-mirage
+ *
+ * @method getApplication
+ * @private
+ * @hide
+ * @return {Object} Application
+ */
+function getApplication() {
+  let theGlobal = window || global || self
+    
+  try {
+    return theGlobal.emberCliMirageApplication
+  } catch(err) {
+    throw new Error('ember-cli-mirage/ember-data | global "emberCliMirageApplication" could not be found: ', err)
+  }
+}
+
+/**
  * Get all ember data models under the app's namespaces
  *
  * @method getDsModels
@@ -26,6 +44,16 @@ export function getDsModels() {
   if (DsModels) {
     return DsModels;
   }
+
+  let application = getApplication()
+
+  /*
+    applicationInstance.lookup('service:store') should've been used instead
+    but ember-cli-mirage/ember-data.js runs before any instance-initializers
+    so there is no way to get hold of the applicationInstance before this code executes,
+    at least not with the current setup
+  */
+  let store = application.__container__.lookup('service:store')
 
   let moduleMap = requirejs.entries;
   let classicModelMatchRegex = new RegExp(`^${modulePrefix}/models/(.*)$`, 'i');
@@ -46,7 +74,7 @@ export function getDsModels() {
     if (matches && matches[1]) {
       let modelName = matches[1];
 
-      let model = require(path, null, null, true).default;
+      let model = store.modelFor(modelName);
       if (isDsModel(model)) {
         DsModels[modelName] = model;
       }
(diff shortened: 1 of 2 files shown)


SergeAstapov:
@Pixelik sorry for delay here, do you mind to refresh branch with latest `master` to get CI going and be green?

Pixelik:
Hello. It all looks green to me and up-to-date, ready to be merged 🤔

SergeAstapov:
> Hello. It all looks green to me and up-to-date, ready to be merged 🤔

branch of this PR is outdated and although the PR status is green, it has no jobs run in "Checks" tab.

Anyways, I've created a unit test which uses `discoverEmberDataModels` function in #641 and yes it fails for embe-date@5 scenarios.
However, the fix suggested here does not work, because `application.__container__` is `undefined`.

Hence, for `discoverEmberDataModels` function compatibility with Ember Data v5 we need a different solution.

SergeAstapov (changes requested):
this actually does not work, see CI run https://github.com/miragejs/ember-cli-mirage/actions/runs/6193523102

SergeAstapov:
@Pixelik there is no such test in master now and this PR does not add anything.

@Pixelik please see https://github.com/miragejs/ember-cli-mirage/pull/2528/files#diff-a16b283fd8b39b8decc1f1e8824d3a777dca6c861c511f6999ace605541c8807 where test added

Pixelik:
@SergeAstapov  
Oh I see you created a separate PR with tests about this
Your unit test is not going to work because the required initializers won't run during your unit test.

SergeAstapov:
@Pixelik well, yes. but how can we guarantee that it will be present by the time `discoverEmberDataModels()` is invoked in mirage config?

SergeAstapov:
@Pixelik sounds like this needs to be documented then?

Pixelik:
Perhaps. Is it possible that someone would try to `discoverEmberDataModels()` **_not_** from within an Ember app that has already been initialized ? 🤔 Sounds weird to me but.. 🤷‍♂️

Pixelik:
But yes, documenting this sounds necessary to me now as well.

cah-brian-gantzler:
Sorry for being late to the party but I would think that autoDiscoverDataModels should be passed the store. 

In setupMirage the this.owner is passed into startMirage (so if startMirage is called outside tests, this is a required param).  https://github.com/miragejs/ember-cli-mirage/blob/master/packages/ember-cli-mirage/addon-test-support/setup-mirage.js#L23, so it will be accessible

(Quick aside, makeServer can be passed as an option on SetupMirage and passed into StartMirage)
```
import makeServer from 'app/mirage/config';

setupMirage(hooks, { makeServer });
```
then in startMirage, if one is passed, it uses it, if not looks it up. https://github.com/miragejs/ember-cli-mirage/blob/master/packages/ember-cli-mirage/addon/start-mirage.js#L21 (This look up is what was prepped in one of the initializers. I would love to see a mandate of passing it in, then could remove one of the intializers). This also has the benefit of specifying different configs for different tests if needed, or a different config for dev vs test. hence why testConfig is no longer needed. 

You could lookup the store and add it here https://github.com/miragejs/ember-cli-mirage/blob/master/packages/ember-cli-mirage/addon/start-mirage.js#L29 or add owner here and require them to look it. Either way you now have a store that you could pass into autoDiscoverDataModels which might be more straight forward and intuitive. Supports multiple stores if someone did have them, plus you dont depend on the initializer for `application.__container__.` (do you even need the initializer?)

SergeAstapov:
Thank you @cah-brian-gantzler and @Pixelik for ideas and discussion! I've landed fix in #2542

PR #2505, opened 2023-08-18
switch from yarn to pnpm

SergeAstapov:
yarn@1 does not handle peer dependencies properly in the monorepo.
pnpm does proper job in this case. Also it's faster

EDIT: I anticipate most of issues we see in #2495 will be resolved


SergeAstapov:
Will work on fixing v5 scenarios separately

PR #2551, opened 2023-10-04
Create models/foo.js exporting a ShimModelClass

bartocc:
This PR is here to illustrate this issue of `discoverEmberDataModels` with ShimModelClass

Run `pnpm --filter="basic-app" start && open http://localhost:4200` to see the error in the console

You can also

[Open in StackBlitz](https://stackblitz.com/~/github.com/miragejs/ember-cli-mirage/pull/2551)

### DIFF
--- a/test-packages/01-basic-app/app/models/foo.js
+++ b/test-packages/01-basic-app/app/models/foo.js
@@ -1,0 +1,1 @@
+export default class Foo {}
--- a/.stackblitzrc
+++ b/.stackblitzrc
@@ -1,0 +1,3 @@
+{
+  "startCommand": "pnpm --filter='basic-app' start"
+}


test-packages/01-basic-app/app/models/foo.js
```diff
@@ -0,0 +1 @@
+export default class Foo {}
```
SergeAstapov, line 1: per https://github.com/miragejs/ember-cli-mirage/commit/3cd81ef7ee3f67e5edfaf725cf64747ba2313183#r129177371 agree with @cah-brian-gantzler that this is not supposed to work as it's not Ember Data model

cah-brian-gantzler:
This https://github.com/miragejs/ember-cli-mirage/issues/2556 is related and after looking into it, it appears a test is done too late. See the issue.

gilest:
Try rebasing now that #2569 is merged

PR #2569, opened 2023-11-29
Filter out shim model class when running `discoverEmberDataModels`

### LINKED ISSUE #2556: `discoverEmberDataModels` crashes if there is a non ember-data object in the app/models folder
We have a bunch of objects which are plain objects in the `app/models` folder.
We use ember-data@3.28.13.
A PR to update ember-cli-mirage from `3.0.0-alpha.5` to `3.0.0` failed.
After adding `config.store` as an argument to `discoverEmberDataModels` I've got to the following exception:

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at ShimModelClass.eachRelationship (http://localhost:4204/assets/vendor.js:111439:14)
    at http://localhost:4204/assets/vendor.js:122501:13
    at Array.forEach (<anonymous>)
    at discoverEmberDataModels (http://localhost:4204/assets/vendor.js:122498:34)
    ...
```

I'm sure we are not the only ones who have non-ember-data objects in `models`. Is it possible to just skip them in `discoverEmberDataModels` rather than be obliged to move them to another folder?

gilest:
Maybe a naive fix, but I was making a fork to use in a project anyway 🤷🏻 

Maybe fixes #2556

Ref:
- [ShimModelClass in ember-data](https://github.com/emberjs/data/blob/main/packages/store/src/-private/legacy-model-support/shim-model-class.ts)

### DIFF
--- a/packages/ember-cli-mirage/addon/utils/ember-data.js
+++ b/packages/ember-cli-mirage/addon/utils/ember-data.js
@@ -11,5 +11,5 @@ export const hasEmberData =
   @hide
 */
 export function isDsModel(m) {
-  return m && typeof m.eachRelationship === 'function';
+  return m && typeof m.eachRelationship === 'function' && m.isModel === true;
 }
--- a/test-packages/01-basic-app/app/models/shim.js
+++ b/test-packages/01-basic-app/app/models/shim.js
@@ -1,0 +1,10 @@
+// Mimics the static apis of ShimModelClass from ember-data
+export default class Shim {
+  fields;
+  attributes;
+  relationshipsByName;
+
+  eachAttribute() {}
+  eachRelationship() {}
+  eachTransformedAttribute() {}
+}


packages/ember-cli-mirage/addon/utils/ember-data.js
```diff
@@ -11,5 +11,5 @@ export const hasEmberData =
   @hide
 */
 export function isDsModel(m) {
-  return m && typeof m.eachRelationship === 'function';
+  return m && typeof m.eachRelationship === 'function' && m.isModel === true;
```
gilest, line 14: Not sure if `isModel` will be `true` for all classes we want to check

test-packages/01-basic-app/app/models/shim.js
```diff
@@ -0,0 +1,10 @@
+// Mimics the static apis of ShimModelClass from ember-data
+export default class Shim {
+  fields;
+  attributes;
+  relationshipsByName;
+
+  eachAttribute() {}
+  eachRelationship() {}
+  eachTransformedAttribute() {}
+}
```
gilest, line 10: When I add this file without the fix, this package's test suite fails
image

gilest:
@SergeAstapov what are your thoughts on this fix?

chancancode:
FWIW, I ran into the same issue and applied the patch locally, and at least in my case, didn't run into issues (e.g. non-ED model brought by addons that happens to have a `isModel` property)

SergeAstapov:
Thank you @gilest!

gilest:
Please cut a release when you get a chance 🥺

zeppelin:
@SergeAstapov This seems to fix the issue with model discoverability in out apps.
Could you please cut a release? 🙏

SergeAstapov:
sorry for delays, this is published as v3.0.3. Thank you @gilest!

PR #2574, opened 2024-02-17
Fix initialiser location

mansona:
Having the main initialiser in the app folder means that a consuming app can't actually override the initialiser. 

This means that you can't have an embroider safe implementation that works in Fastboot, because we can't have a fastboot-specific ember-cli-mirage initialiser that doesn't (eventually) import pretender. This means that even though pretender doesn't actually work in node there is no way to prevent it being imported in your app, until this PR is merged 👍


packages/ember-cli-mirage/index.js
```diff
@@ -106,8 +106,8 @@ module.exports = {
     });
   },
 
-  treeForApp(appTree) {
-    let trees = [appTree];
+  treeForAddon(addonTree) {
```
SergeAstapov, line 109: @mansona could you please explain a bit this change?

I got the point about initializer location in app folder and that totally makes sense. But this change seems separate?

PR #2575, opened 2024-02-17
use pnpm node version selection instead of volta

mansona:
Volta isn't needed any more since you're using pnpm and it has its own way to pick the node version https://pnpm.io/npmrc#use-node-version


SergeAstapov:
Thank you @mansona!

PR #2578, opened 2024-09-01
add v5.x to the list of ember-inflector versions

poulet42:
This adds `^v5.0.0` to the list of allowed versions for ember-inflector. The major is not breaking, it's just been converted to a v2 addon


SergeAstapov:
Thank you @poulet42 !

